PR #122, opened 2026-04-13
fix: persist memories to disk immediately after memory_save

JasonLandbridge:
## Problem

In `standalone.ts`, `kv.persist()` (which writes the in-memory store to `~/.agentmemory/standalone.json`) was only called on `SIGINT`/`SIGTERM` signal handlers:

```ts
process.on("SIGINT", () => { kv.persist(); process.exit(0); });
process.on("SIGTERM", () => { kv.persist(); process.exit(0); });
```

If the MCP server process is killed with `SIGKILL`, crashes unexpectedly, or the host terminates it forcefully (e.g. when an AI coding agent session ends), **all memories saved during that session are permanently lost** because `standalone.json` is never written.

## Fix

Call `kv.persist()` immediately after `kv.set()` in the `memory_save` handler, so data is written to disk on every save:

```ts
await kvInstance.set("mem:memories", id, { ... });
kvInstance.persist();  // ← ensure data survives unexpected termination
```

## Additional Changes

- **Exported `handleToolCall`** with an injectable `kvInstance` parameter (defaults to the module-level `kv`) to enable unit testing
- **Mocked `transport.js` and `config.js`** in `test/mcp-standalone.test.ts` to prevent side effects when importing the module
- **Added 4 new tests** in a `handleToolCall` describe block:
  - `memory_save persists to disk immediately after saving` — verifies `writeFileSync` is called
  - `memory_save without persist path does not call writeFileSync` — no unnecessary writes when no path configured
  - `memory_save throws when content is missing` — validates error handling
  - `memory_recall returns matching memories` — end-to-end recall test

All 625 existing tests continue to pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory save and retrieval so saved items are consistently persisted, timestamped, and retrievable across sessions; save now reliably triggers disk persistence when configured.
  * Validation tightened so attempts to save without required content produce clear errors.
  * Memory listing/recall now reads from the active storage instance for accurate results.

* **Tests**
  * Added isolated tests covering persistence behavior, save/recall flows, and error cases.

### DIFF
--- a/src/mcp/standalone.ts
+++ b/src/mcp/standalone.ts
@@ -5,6 +5,7 @@ import { createStdioTransport } from "./transport.js";
 import { getVisibleTools } from "./tools-registry.js";
 import { getStandalonePersistPath } from "../config.js";
 import { VERSION } from "../version.js";
+import { generateId } from "../state/schema.js";
 
 const IMPLEMENTED_TOOLS = new Set([
   "memory_save",
@@ -31,7 +32,8 @@ export async function handleToolCall(
     case "memory_save": {
       const content = args.content as string;
       if (!content?.trim()) throw new Error("content is required");
-      const id = `mem_${Date.now().toString(36)}_${Math.random().toString(36).slice(2, 8)}`;
+      const id = generateId("mem");
+      const isoNow = new Date().toISOString();
       await kvInstance.set("mem:memories", id, {
         id,
         type: (args.type as string) || "fact",
@@ -43,8 +45,8 @@ export async function handleToolCall(
         files: args.files
           ? (args.files as string).split(",").map((f) => f.trim())
           : [],
-        createdAt: new Date().toISOString(),
-        updatedAt: new Date().toISOString(),
+        createdAt: isoNow,
+        updatedAt: isoNow,
         strength: 7,
         version: 1,
         isLatest: true,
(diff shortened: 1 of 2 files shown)


rohitg00:
Thanks for this — SIGKILL data loss was a real hole and the injectable `kvInstance` refactor + test mocks make this easy to merge confidently. Merging now.

Nice touches worth calling out:
- Switching to the shared `generateId("mem")` helper instead of the ad-hoc concat
- Single `isoNow` shared by `createdAt`/`updatedAt` so they can't drift
- The rationale for *not* using `fingerprintId` (memory_save is append-only, content-addressable would silently overwrite) matched the project's existing behavior exactly

Landing as v0.8.4 in the next patch bump.

JasonLandbridge:
> Thanks for this — SIGKILL data loss was a real hole and the injectable `kvInstance` refactor + test mocks make this easy to merge confidently. Merging now.
> 
> Nice touches worth calling out:
> 
>     * Switching to the shared `generateId("mem")` helper instead of the ad-hoc concat
> 
>     * Single `isoNow` shared by `createdAt`/`updatedAt` so they can't drift
> 
>     * The rationale for _not_ using `fingerprintId` (memory_save is append-only, content-addressable would silently overwrite) matched the project's existing behavior exactly
> 
> 
> Landing as v0.8.4 in the next patch bump.

Thanks for the quick merge, and thank you for this awesome project!

PR #131, opened 2026-04-13
fix(mcp): never respond to JSON-RPC notifications (#129)

### LINKED ISSUE #129: MCP doesn't work on codex-cli
MCP client for `agentmemory` failed to start: MCP startup failed: Transport closed

Tested with multiple command (#120 ).

Codex config.toml:
```
[mcp_servers.agentmemory]
command = "npx"
args = ["agentmemory-mcp"]
enabled = true
```

(I use npx agentmemory-mcp because I've installed @agentmemory/agentmemory globally)

It "correctly" start when launched in a terminal:
`[agentmemory-mcp] Standalone MCP server v0.8.1 starting...`

Other remote/stdio MCP are working

Codex-cli version: v0.120.0

rohitg00:
Fixes #129 reported by @GuillaumeOuint.

## Symptom

\`\`\`
MCP client for agentmemory failed to start: MCP startup failed: Transport closed
\`\`\`

Codex CLI v0.120.0 fails to start the standalone MCP server. Other MCP clients (Claude Desktop, Cursor, OpenClaw) work fine with the same binary. Running \`npx agentmemory-mcp\` directly in a terminal also "works" — it prints the v0.8.1 banner and waits for input. Only Codex's stdio handshake breaks.

## Root cause

\`src/mcp/transport.ts\` unconditionally writes a JSON-RPC response for **every** incoming message, including notifications. Per [JSON-RPC 2.0 §4.1](https://www.jsonrpc.org/specification#notification):

> A Notification is a Request object without an "id" member. [...] **The Server MUST NOT reply to a Notification**, including those that are within a batch request.

Lenient clients (Claude Desktop, Cursor) silently ignore unsolicited responses. Strict clients (Codex CLI) treat them as protocol violations and close the transport. That's exactly what Guillaume sees.

The exact sequence on Codex CLI:

1. Codex sends `initialize` (request, has `id`) → server responds ✓
2. Codex sends `notifications/initialized` (**notification, no `id`**)
3. Our handler in \`standalone.ts:138-139\` returns \`{}\`
4. \`transport.ts:62-67\` writes \`{\"jsonrpc\":\"2.0\",\"result\":{}}\` to stdout (no \`id\` field because \`request.id\` is undefined)
5. Codex parses an unsolicited response → closes transport
6. Guillaume sees \`Transport closed\`

\`memory_save\` / \`memory_recall\` etc never had a chance to fire because the handshake itself was broken.

## Fix

- **\`src/mcp/transport.ts\`**: detect notifications via missing/null \`id\`, run the handler for side effects, then **drop the response**. Errors inside notification handlers are logged to stderr (not silently swallowed) so production debugging still works
- **\`src/mcp/transport.ts\`**: malformed messages with no \`id\` are also dropped silently (consistent with notification semantics) instead of being responded to with \`id: null\`
- **\`src/mcp/transport.ts\`**: extracted \`processLine()\` as an exported function so the line-handling logic is testable independently of \`process.stdin\` / \`process.stdout\`. The \`createStdioTransport\` factory still works the same way externally
- **\`test/mcp-transport.test.ts\`** (new): 9 tests covering all paths
  - Request responses (success + error)
  - Notification suppression (no \`id\` + \`id: null\` + handler-throws)
  - Malformed input (parse error, invalid request with \`id\`, invalid request without \`id\`, empty/whitespace lines)

The existing \`test/mcp-standalone.test.ts\` mocks the transport entirely so it never exercised the notification path. The new file imports the real \`processLine\` and verifies the actual behavior.

## Why the bug went undetected

- All existing MCP clients we test against (Claude Desktop, Cursor, OpenClaw, Hermes, the gateway plugin) tolerate unsolicited responses — they just drop them
- Codex CLI's MCP support is newer and stricter
- The standalone server passes a manual smoke test in a terminal because no client is doing the handshake — only the stderr banner is visible

## Test plan

- [x] \`npm test\` → 693 / 693 passing (684 baseline + 9 new)
- [x] \`npm run build\` → clean
- [ ] After merge + release: ask @GuillaumeOuint to upgrade to the new version and re-test with the same Codex CLI config
- [ ] Should also fix any other strict MCP clients that hit the same issue (probably more to come as MCP adoption broadens)

## Related

- Closes #129 (pending verification from reporter)
- Reporter installed \`@agentmemory/agentmemory\` globally so they're seeing v0.8.1 — this fix needs a release on top of main to actually reach them

## Summary by CodeRabbit

* **Bug Fixes**
  * More robust JSON-RPC handling: malformed JSON yields a Parse error (id null); invalid request shapes return Invalid Request when an id is present; non-primitive ids are rejected.
  * Notifications (no id or id null) no longer generate responses; notification handler failures are logged without replying.
  * Handler exceptions for regular requests produce an internal error response echoing the request id.

* **Tests**
  * Added comprehensive tests covering parsing, request/notification behavior, id validation, and error responses.

### DIFF
--- a/src/mcp/transport.ts
+++ b/src/mcp/transport.ts
@@ -28,6 +28,19 @@ function isNotification(req: JsonRpcRequest): boolean {
   return req.id === undefined || req.id === null;
 }
 
+// Per JSON-RPC 2.0 §4, a valid request id must be a String, Number, or Null
+// (Null is technically only allowed in responses; in requests, omitting id
+// is the convention for notifications, which we treat the same as null).
+// Any other runtime type (object, array, boolean) is an Invalid Request.
+function isValidId(id: unknown): id is string | number | null | undefined {
+  return (
+    id === undefined ||
+    id === null ||
+    typeof id === "string" ||
+    typeof id === "number"
+  );
+}
+
 // Exported for unit tests so the line-handling logic is exercised
 // independently of process.stdin / process.stdout.
 export async function processLine(
@@ -52,18 +65,38 @@ export async function processLine(
   }
 
   const request = parsed as JsonRpcRequest;
+  const rawId = (request as { id?: unknown } | null)?.id;
+
+  // Invalid request shape (missing/wrong jsonrpc, non-string method).
   if (
     !request ||
+    typeof request !== "object" ||
     request.jsonrpc !== "2.0" ||
     typeof request.method !== "string"
   ) {
-    if (request && (request as { id?: unknown }).id != null) {
+    // Echo the id back only if it's a valid string/number. Notifications
+    // (missing/null id) and malformed ids both drop silently — we don't
+    // want to respond to something that could be a notification, and we
+    // can't invent an id for a malformed one.
+    if (typeof rawId === "string" || typeof rawId === "number") {
       writeOut({
         jsonrpc: "2.0",
-        id: (request as JsonRpcRequest).id as string | number,
+        id: rawId,
         error: { code: -32600, message: "Invalid Request" },
       });
     }
+    return;
+  }
+
+  // Request shape is valid but id may still be of the wrong type
+  // (object, array, boolean). Per the spec, that's an Invalid Request.
+  // Respond with id: null because we can't safely echo a non-JSON-RPC id.
+  if (!isValidId(rawId)) {
+    writeOut({
+      jsonrpc: "2.0",
+      id: null,
+      error: { code: -32600, message: "Invalid Request: id must be string, number, or null" },
+    });
     return;
   }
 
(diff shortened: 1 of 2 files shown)


rohitg00:
Addressed the CodeRabbit id-validation finding in 461ef82.

Before: `processLine` only checked `id != null` in the invalid-shape path, so (a) malformed messages carrying `id: {}` could be echoed back with that object as the response id, and (b) valid-shape requests with `id: {}` / `[]` / `true` fell through to the handler and got a response using the bogus id. Both are JSON-RPC 2.0 §4 violations.

Now:
- New `isValidId()` helper: `string | number | null | undefined`.
- `rawId` extracted once as `unknown`.
- Invalid-shape path only echoes when `rawId` is string or number; malformed-with-bad-id drops silently (can't safely synthesize an id).
- After shape validation, bad-id requests get `-32600` with `id: null` *before* the handler is invoked.

5 new tests in `test/mcp-transport.test.ts` cover object/array/boolean ids, the malformed+non-primitive drop, and a positive string-id path. Full suite: 698 passing.

PR #132, opened 2026-04-13
fix: route semantic memory eviction to correct KV store using source …

Tanmay-008:
The mem::retention-evict loop was previously hardcoded to delete candidates 
only from KV.memories. This caused semantic memories to fail silently and 
remain alive in KV.semantic indefinitely, even with sub-threshold scores.

Changes implemented:
- Added `source: 'episodic' | 'semantic'` to the RetentionScore interface.
- Populated the `source` field during mem::retention-score calculation.
- Branched the deletion logic in mem::retention-evict to route deletes 
  to either KV.memories or KV.semantic based on the candidate's source.
- Added test cases to seed both memory types and verify successful eviction 
  from their respective KV stores when scores drop below the threshold.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced memory eviction logic to properly validate and identify correct storage locations before deletion. When the source is unclear, the system now probes available storage options to ensure data is deleted from the right location, preventing unintended data loss.

* **Updates**
  * Extended type definitions to improve retention score tracking and source identification.


rohitg00:
Is this related to some issue?

PR #185, opened 2026-04-22
providers: add OpenAI compatible provider

saem:
## Summary

Created OpenAI compatible model and embedding providers.

## Details

The new provider is compatible with any OpenAI API endpoint. This includes OpenRouter, which has been subsumed as part of this change.

Detection has been added for:

- OpenAI
- LM Studio
- ollama
- vllm

This should allow for easy local model detection and usage.

---

- [x] undo the consolidation of `OpenRouter`

Notes for Reviewers:

- I only have LM Studio setup so testing that locally
- full disclosure, this was developed via Gemini
- gemini went with two style changes that I'm not sure about:
  1. usage of mocks rather than process env vars + clean-up
  2. usage of constructor parameters for member declaration

## Summary by CodeRabbit

* **New Features**
  * Added support for OpenAI, LM Studio, Ollama, and vLLM as LLM and embedding providers; new local/default base URL and model handling.
  * Introduced a new provider implementation for OpenAI-compatible backends with configurable timeouts and headers.

* **Documentation**
  * Updated README and .env example to document new providers, embedding config, and a minor wording tweak.

* **Tests**
  * Expanded tests for provider detection, embedding behavior, and request handling; adjusted filesystem watcher timing.


saem:
FYI, I'm:

- [x] getting rid of the mocks and increasing test covering
- [x] making sure I have LM Studio working end to end for me

---

I would like to know if the general idea of the PR (an openai router + detection for various local runtimes) is something that the project would like to see developed? I haven't spent much time on it, so it's fine if you're not interested.

src/providers/embedding/openrouter.ts
saem, line 1: This should either be restored or the `baseUrl` + `api` prefix needs to be handled properly

saem:
@rohitg00 apologies for the ping, just wondering if you're ok with this PR in spirit? If so I'll keep going otherwise I'll close it down.

rohitg00:
I think, This is already addressed in today's release by someone.

saem:
> I think, This is already addressed in today's release by someone.

If you're thinking of #186 that was only the openai embedding provider

saem:
I've cleaned things up, I still think this is useful.

Please take a look and let me know.

src/providers/index.ts
```diff
@@ -59,6 +60,38 @@ export function createFallbackProvider(
...
+        config.baseURL || "http://localhost:11434",
+      );
+    case "vllm":
+      return new OpenAIProvider(
+        "vllm",
+        "no-key-required",
+        config.model,
+        config.maxTokens,
+        config.baseURL!,
+      );
```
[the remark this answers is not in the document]
saem: Since this is primarily meant for local use, the default baseURL should be `http://localhost:8000`, as `8000` is the default port for vllm.

Qodo-Free-For-OSS:
Hi, detectEmbeddingProvider() returns "lmstudio"/"vllm" when only one of base URL or model is set, but createEmbeddingProvider() throws if required variables are missing. This can crash worker startup for partially-configured environments (including when only a base URL is set).

**How to fix:** Align detect/create requirements

> ## Issue description
> `detectEmbeddingProvider()` can return `lmstudio`/`vllm` with partial env configuration, but `createEmbeddingProvider()` throws unless required variables are present. This can crash startup.
> 
> ## Issue Context
> - Detection currently checks `(BASE_URL || MODEL)` for lmstudio/vllm.
> - Creation requires:
>   - lmstudio: `LMSTUDIO_EMBEDDING_MODEL` (base has a default)
>   - vllm: both `VLLM_EMBEDDING_BASE_URL` and `VLLM_EMBEDDING_MODEL`
> - `src/index.ts` calls `createEmbeddingProvider()` during startup.
> 
> ## Fix Focus Areas
> - src/config.ts[201-219]
> - src/providers/embedding/index.ts[30-64]
> 
> ## Suggested direction
> Make detection require the same minimum inputs as creation (e.g., lmstudio: require MODEL; vllm: require both BASE_URL+MODEL), or make creation return `null` (or a clearer error/warning) when inputs are incomplete instead of throwing during startup.

We noticed a couple of other issues in this PR as well - happy to share if helpful.

---
*Found by [Qodo](https://qodo.ai) code review*

rohitg00 (changes requested):
Thanks for this — local-provider support is a real unlock and the generic `OpenAIProvider` is clean. Few things to tune before merge:

## Blockers

### 1. Provider detection order flips existing users

`detectProvider` now checks in this order:
```
minimax → lmstudio → ollama → vllm → openai → anthropic → gemini → openrouter
```

Any user with both `ANTHROPIC_API_KEY` and `OPENAI_API_KEY` set (common — lots of people keep both in `~/.env`) silently flips from Anthropic to OpenAI after upgrading. That's a behavior change we can't ship without either a priority-preserving order OR a loud warning.

**Fix:** Move the four new detections **after** `openrouter` so legacy keys still win. Or introduce an explicit `AGENTMEMORY_PROVIDER=openai` override for deliberate selection and keep new detections behind it.

### 2. Hardcoded `"gpt-4o"` default

`src/config.ts`:
```ts
model: env["OPENAI_MODEL"] || "gpt-4o",
```

This violates the unopinionated-routing rule the rest of the codebase follows — agentmemory doesn't ship a model catalog. Other providers require the user to pick a model explicitly (or pick from `*_MODEL` env). A hardcoded `"gpt-4o"` bakes a pricing/capability assumption into the library that will age badly.

**Fix:** Require `OPENAI_MODEL` (throw with a clear error if unset) or use a neutral placeholder like `"openai-default"` and document it.

## Needs cleanup (non-blocking but request before merge)

### 3. `"no-key-required"` sentinel string

`src/providers/openai.ts` + `src/providers/embedding/openai.ts` use the literal `"no-key-required"` to mean "skip Authorization header." That's fragile — any user who accidentally names their key `no-key-required` hits it, and the intent is hidden.

**Fix:** `constructor(apiKey: string | null, ...)` + `if (this.apiKey) { headers.Authorization = ... }`. Same intent, zero magic string.

### 4. Overlap with #186 already shipped in v0.9.2

`OPENAI_BASE_URL` + `OPENAI_EMBEDDING_MODEL` landed in v0.9.2 (#186). Your ctor-param additions extend that cleanly (`arg ?? env ?? default`) but please rebase on latest `main` to surface any unexpected drift, and add a `### Changed` note to CHANGELOG explaining "`OpenAIEmbeddingProvider` constructor now accepts optional `baseUrl` + `model` args; env vars remain fallback."

### 5. Priority of OpenRouter vs. generic OpenAIProvider

OpenRouter is OpenAI-compatible — could be subsumed under `OpenAIProvider` with `baseURL: "https://openrouter.ai/api/v1"`. You flagged this yourself and kept the separate `OpenRouterProvider`. For now keeping both is fine; raise an issue to consolidate in a later release so we don't have two code paths for the same protocol.

## Minor

- `feedback_original_naming` style: PR adds `VLLM_BASE_URL` / `LMSTUDIO_BASE_URL` — brands in env names. Fine, matches the existing `OPENAI_BASE_URL` / `COHERE_API_KEY` pattern.
- Tests using mocks vs env-var+cleanup (you flagged this) — either pattern is OK, keep what's there.
- Constructor-param member declaration — non-issue, modern TS idiom.

## Summary

**Blockers** (1) + (2) would silently change behavior for anyone already using agentmemory with Anthropic. Fix those and this ships. (3) is a 5-line cleanup. (4) is a rebase + one CHANGELOG line.

Happy to re-review as soon as those three are in.

saem:
I think I've addressed all the code feedback, including creating the issue (https://github.com/rohitg00/agentmemory/issues/199)

saem:
There is a minor issue with the openai provider having a default model approach, if openai switches to a reasoning model by default we won't be able to detect it, but I think that's fine (deal with it when it happens).

CHANGELOG.md
```diff
@@ -6,6 +6,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
 
+### Added
+
+- **Support for local OpenAI-compatible providers.** Full support for Ollama, vLLM, and LM Studio as first-class providers. Added new configuration variables:
+  - **Ollama**: `OLLAMA_BASE_URL` (default: `http://localhost:11434`) and `OLLAMA_MODEL`.
+  - **LM Studio**: `LMSTUDIO_BASE_URL` and `LMSTUDIO_MODEL`.
+  - **vLLM**: `VLLM_BASE_URL` and `VLLM_MODEL`.
+  Auto-detection now supports these local endpoints while ensuring existing legacy keys (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`) remain prioritized for stable upgrades.
+- **`AGENTMEMORY_PROVIDER` environment variable.** Allows explicit selection of the LLM provider (e.g., `AGENTMEMORY_PROVIDER=ollama`), bypassing auto-detection and supporting setups with multiple keys.
+- **Improved `OpenAIEmbeddingProvider` flexibility.** The constructor now accepts optional `baseUrl` and `model` overrides, enabling direct instantiation for custom OpenAI-compatible embedding proxies.
```
saem:
```suggestion
- **Improved `OpenAIEmbeddingProvider` flexibility.** The constructor now accepts optional `baseUrl` and `model` overrides, enabling direct instantiation for custom OpenAI-compatible embedding proxies. Environment variables remain as fallback.
```

rohitg00:
Hi @saem — thanks for putting this together, this was a real gap and the OpenAI / OpenRouter / Ollama / vLLM detection logic here is solid.

Status: this branch has gone stale (CHANGES_REQUESTED + CONFLICTING against current `main`), and a smaller, cleaner OpenAI-compatible provider PR has come in since — #240 — that keeps the OpenRouter detection separate (your TODO around "undo the consolidation of OpenRouter") and lands the same `OPENAI_API_KEY` + `OPENAI_BASE_URL` surface.

Plan: we'll land #240 as the v1 of the OpenAI-compatible provider, and roll any extras from this branch (LM Studio / Ollama / vLLM-specific detection, the embedding-side OpenRouter unification) as follow-up issues with attribution to your work here. Closing this in favor of that path — please open follow-up PRs against the `openai` provider that lands in #240 if you want any of those pieces preserved, and I'll prioritize review.

Really appreciate the time you put into the original investigation.

PR #184, opened 2026-04-21
fix(runtime): make v0.9.1 dist actually start

rohitg00:
## Summary

v0.9.1 dist crashed on first import. Local reproduction during end-to-end slot testing surfaced four independent bugs that each block startup.

## What was broken

1. **`getContext` re-imported in three files** — PR #111's conflict resolution re-added \`import { getContext } from \"iii-sdk\"\` to \`compress.ts\`, \`disk-size-manager.ts\`, \`image-quota-cleanup.ts\`. iii-sdk v0.11 dropped that export (see the shim comment in \`src/logger.ts\`). Tests mocked iii-sdk so \`npm test\` passed, but \`node dist/cli.mjs\` died with \`The requested module 'iii-sdk' does not provide an export named 'getContext'\`.
2. **Old \`registerFunction({ id, description }, handler)\` shape** — two of those files also used the pre-0.11 signature. 0.11 only accepts \`registerFunction(id, handler, options?)\`.
3. **\`KV.state\` missing** — same conflict resolution pass dropped \`state: \"mem:state\"\` from \`src/state/schema.ts\`. \`disk-size-manager.ts\` uses that key for persistence.
4. **tsdown inlined onnxruntime-node + @xenova/transformers** — the bundled \`binding.js\` rewrote the relative require for \`../bin/napi-v3/darwin/arm64/onnxruntime_binding.node\` so it no longer resolved from \`dist/\`. Even with \`AGENTMEMORY_IMAGE_EMBEDDINGS\` off, the module graph still evaluated the bundled binding on startup and threw.

## What shipped

- Dropped the three \`getContext\` imports, switched to module-level \`logger\`.
- Collapsed \`registerFunction({ id, description }, ...)\` to \`registerFunction(id, ...)\`.
- Restored \`state: \"mem:state\"\` in \`src/state/schema.ts\`.
- \`tsdown.config.ts\` marks these as \`external\`: \`@xenova/transformers\`, \`onnxruntime-node\`, \`onnxruntime-web\`, \`@anthropic-ai/claude-agent-sdk\`, \`@anthropic-ai/sdk\`. Bundle 6.1 MB → 1.9 MB. CLIP + local embedding providers lazy-load them from \`node_modules\` where relative paths work.
- Bumped \`iii-sdk\` 0.11.0 → 0.11.2 to match the shipping API (Logger / durable:subscriber / durable:publisher / TriggerAction.void).
[...]

PR #186, opened 2026-04-22
feat: support OPENAI_BASE_URL and OPENAI_EMBEDDING_MODEL for OpenAI-compatible endpoints

Edison-A-N:
## Summary

- Add `OPENAI_BASE_URL` and `OPENAI_EMBEDDING_MODEL` env vars to `OpenAIEmbeddingProvider`, enabling use with OpenAI-compatible proxies (Azure OpenAI, vLLM, LM Studio, domestic relays, etc.)
- Follows the same pattern as `MinimaxProvider` (`MINIMAX_BASE_URL`) — env var with sensible default, zero breaking changes

## Details

**Problem:** `OpenAIEmbeddingProvider` hardcodes `https://api.openai.com/v1/embeddings` and `text-embedding-3-small`. Users behind proxies or using compatible APIs cannot configure the endpoint.

**Fix:** Two optional env vars with backward-compatible defaults:

| Env Var | Default | Description |
|---|---|---|
| `OPENAI_BASE_URL` | `https://api.openai.com` | Base URL without path |
| `OPENAI_EMBEDDING_MODEL` | `text-embedding-3-small` | Embedding model name |

**Files changed:** 2 (provider + tests)
**Tests added:** 4 (default behavior, error case, base URL override, model override)

## Related

See also #185 which takes a broader approach (full OpenAI-compatible provider + local runtime detection). This PR is intentionally minimal and focused on the embedding provider only.

## How to verify

```bash
OPENAI_API_KEY=sk-test OPENAI_BASE_URL=https://your-proxy.com npm test -- test/embedding-provider.test.ts
```

## Release Notes

* **New Features**
  * OpenAI embedding provider now supports custom OpenAI-compatible API endpoints
  * Configurable embedding model selection via environment variables
  * Default settings preserved for backward compatibility

### DIFF
--- a/src/providers/embedding/openai.ts
+++ b/src/providers/embedding/openai.ts
@@ -1,16 +1,33 @@
 import type { EmbeddingProvider } from "../../types.js";
 import { getEnvVar } from "../../config.js";
 
-const API_URL = "https://api.openai.com/v1/embeddings";
+const DEFAULT_BASE_URL = "https://api.openai.com";
+const DEFAULT_MODEL = "text-embedding-3-small";
 
+/**
+ * OpenAI-compatible embedding provider.
+ *
+ * Required env vars:
+ *   OPENAI_API_KEY           — API key
+ *
+ * Optional:
+ *   OPENAI_BASE_URL          — base URL without path (default: https://api.openai.com)
+ *   OPENAI_EMBEDDING_MODEL   — model name (default: text-embedding-3-small)
+ */
 export class OpenAIEmbeddingProvider implements EmbeddingProvider {
   readonly name = "openai";
   readonly dimensions = 1536;
   private apiKey: string;
+  private baseUrl: string;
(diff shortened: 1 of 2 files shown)


rohitg00:
Thanks @Edison-A-N for the contribution — merged!

Mirrors the `MINIMAX_BASE_URL` pattern cleanly and unlocks Azure OpenAI / vLLM / LM Studio for embeddings with zero breakage. Rebased on latest main before merging so the two-step CI install + lockfile gitignore from #184/#187/#188 are picked up.

Opened a small follow-up at #189 to also derive `dimensions` from the configured model (CodeRabbit flagged that `text-embedding-3-large` would still report 1536). New `OPENAI_EMBEDDING_DIMENSIONS` env var + known-models lookup table handle it, with a fallback to 1536 so existing deployments stay put.

PR #193, opened 2026-04-23
feat: add --host flag and AGENTMEMORY_HOST for configurable bind address

Edison-A-N:
## Summary

Adds opt-in support for binding agentmemory services to non-localhost addresses (e.g. `0.0.0.0` for LAN access). The default remains `127.0.0.1` per security advisory GHSA #03.

## Motivation

Users deploying agentmemory on remote dev servers or shared machines need to access the Viewer (port 3113), REST API (3111), and WebSocket streams (3112) from other hosts on the network. Currently this is impossible without patching source or running a reverse proxy, since all three services hardcode `127.0.0.1`.

## Changes

- **CLI**: New `--host <addr>` flag (sets `AGENTMEMORY_HOST` env var)
- **Config**: New `host` field in `AgentMemoryConfig` (default `127.0.0.1`)
- **Viewer server**: `startViewerServer()` accepts `host` param, binds accordingly
- **iii-engine**: CLI patches `iii-config.yaml` at startup when host is overridden, creating a temp config with the new bind address and wildcard CORS
- **CORS**: Viewer dynamically allows origins on configured ports when bound to `0.0.0.0`
- **CSP**: `connect-src` relaxed to `*` when externally bound
- **Security**: Warns at startup if `--host` is non-localhost without `AGENTMEMORY_SECRET` set

## Usage

```bash
# CLI flag
npx @agentmemory/agentmemory --host 0.0.0.0

# Or via env in ~/.agentmemory/.env
AGENTMEMORY_HOST=0.0.0.0
```

## Security Considerations

- Default behavior is unchanged (`127.0.0.1`)
- Explicit opt-in required via `--host` flag or `AGENTMEMORY_HOST` env
- Startup warning when binding externally without `AGENTMEMORY_SECRET`
- CSP and CORS only relaxed when `AGENTMEMORY_HOST=0.0.0.0`


Edison-A-N:
Closing — branch had unintended file deletions from partial clone. Will re-submit with clean branch.

PR #194, opened 2026-04-23
feat: add --host flag and AGENTMEMORY_HOST for configurable bind address

Edison-A-N:
## Summary

Adds opt-in support for binding agentmemory services to non-localhost addresses (e.g. `0.0.0.0` for LAN access). The default remains `127.0.0.1` per security advisory GHSA #03.

## Motivation

Users deploying agentmemory on remote dev servers or shared machines need to access the Viewer (port 3113), REST API (3111), and WebSocket streams (3112) from other hosts on the network. Currently this is impossible without patching source or running a reverse proxy, since all three services hardcode `127.0.0.1`.

## Changes (6 files, ~85 lines added)

- **`src/cli.ts`**: New `--host <addr>` flag, security warning when binding without `AGENTMEMORY_SECRET`, runtime patching of `iii-config.yaml` for iii-engine host override
- **`src/config.ts`**: Read `AGENTMEMORY_HOST` env var into config
- **`src/types.ts`**: Add `host` field to `AgentMemoryConfig` interface
- **`src/index.ts`**: Pass `host` to viewer server, display actual bind address in logs
- **`src/viewer/server.ts`**: Accept `host` param in `startViewerServer()`, dynamic CORS origin checking when bound to `0.0.0.0`
- **`src/auth.ts`**: Relax CSP `connect-src` to `*` when externally bound

## Usage

```bash
# CLI flag
npx @agentmemory/agentmemory --host 0.0.0.0

# Or via env in ~/.agentmemory/.env
AGENTMEMORY_HOST=0.0.0.0
```

## Security Considerations

- Default behavior is **unchanged** (`127.0.0.1`)
- Explicit opt-in required via `--host` flag or `AGENTMEMORY_HOST` env
- Startup warning when binding externally without `AGENTMEMORY_SECRET`
- CSP and CORS only relaxed when `AGENTMEMORY_HOST=0.0.0.0`
- iii-engine config patched at runtime via temp file (original untouched)

## Summary by CodeRabbit

* **New Features**
  * Added a --host CLI option to set the server bind address.
  * Server can bind to non-local interfaces; host is now configurable.
  * Startup warning when binding to a non-local host without a secret.

* **Improvements**
  * CORS and content-security policies adapt based on chosen host (more permissive for 0.0.0.0).
  * Startup logging now shows the actual host used.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -14,6 +14,11 @@ export function createViewerNonce(): string {
 }
 
 export function buildViewerCsp(nonce: string): string {
+  const host = process.env.AGENTMEMORY_HOST;
+  const connectSrc =
+    host === "0.0.0.0"
+      ? "connect-src *"
+      : "connect-src 'self' http://localhost:* http://127.0.0.1:* ws://localhost:* ws://127.0.0.1:* wss://localhost:* wss://127.0.0.1:*";
   return [
     "default-src 'none'",
     "base-uri 'none'",
@@ -23,7 +28,7 @@ export function buildViewerCsp(nonce: string): string {
     `script-src 'nonce-${nonce}'`,
     "script-src-attr 'none'",
     "style-src 'unsafe-inline'",
-    "connect-src 'self' http://localhost:* http://127.0.0.1:* ws://localhost:* ws://127.0.0.1:* wss://localhost:* wss://127.0.0.1:*",
+    connectSrc,
     "img-src 'self'",
     "font-src 'self'",
   ].join("; ");
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -88,6 +88,13 @@ if (hostIdx !== -1 && args[hostIdx + 1]) {
 
 const skipEngine = args.includes("--no-engine");
 
+function isLoopbackHost(host: string): boolean {
+  if (host === "localhost" || host === "::1" || host === "::ffff:127.0.0.1") {
+    return true;
+  }
+  return host.startsWith("127.");
+}
+
 function getRestPort(): number {
   const url = process.env["AGENTMEMORY_URL"];
   if (url) {
@@ -259,7 +266,7 @@ async function startEngine(): Promise<boolean> {
   if (iiiBin && configPath) {
     const hostOverride = process.env["AGENTMEMORY_HOST"];
     const effectiveConfig =
-      hostOverride && hostOverride !== "127.0.0.1"
+      hostOverride && !isLoopbackHost(hostOverride)
         ? patchIiiConfigHost(configPath, hostOverride)
         : configPath;
     const s = p.spinner();
@@ -303,7 +310,7 @@ async function startEngine(): Promise<boolean> {
   if (iiiBin && configPath) {
     const hostOverride = process.env["AGENTMEMORY_HOST"];
     const effectiveConfig =
-      hostOverride && hostOverride !== "127.0.0.1"
+      hostOverride && !isLoopbackHost(hostOverride)
         ? patchIiiConfigHost(configPath, hostOverride)
         : configPath;
     const s = p.spinner();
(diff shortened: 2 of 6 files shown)


Qodo-Free-For-OSS:
Hi, The viewer server now supports binding to a configurable host, but its REST proxy still always forwards to 127.0.0.1. If iii-engine is configured (via --host) to bind only to a non-loopback interface, the viewer proxy will fail to reach the REST API.

**Severity:** action required | **Category:** correctness

**How to fix:** Proxy to configured host

> ### Issue description
> The viewer binds to `config.host`, but proxies REST calls to `127.0.0.1` unconditionally. When iii-engine is patched to bind to a non-loopback address, the viewer proxy can’t reach it.
> 
> ### Issue Context
> - `patchIiiConfigHost()` rewrites iii-engine host in YAML.
> - `proxyToRestApi()` still targets loopback.
> 
> ### Fix Focus Areas
> - src/viewer/server.ts[80-90]
> - src/viewer/server.ts[152-169]
> - src/cli.ts[113-127]
> 
> ### Suggested approach
> - Thread an `upstreamHost` into `proxyToRestApi()` (or compute inside):
>   - If `resolvedHost === "0.0.0.0"`, use `127.0.0.1` for upstream.
>   - Else use `resolvedHost` (and consider IPv6 formatting, e.g. `[::1]`).
> - Alternatively, constrain `--host` to only `127.0.0.1` and `0.0.0.0` if you want to guarantee loopback proxying.

We noticed a couple of other issues in this PR as well - happy to share if helpful.

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

Edison-A-N:
Thanks for the detailed reviews!

The `--host` flag is intentionally scoped to `0.0.0.0` — the use case is accessing agentmemory from another machine on the LAN. This is consistent with how `--host` works in Vite, Next.js, Playwright, etc., which simply pass the value to `server.listen()` without additional CORS/CSP machinery.

On the specific findings:

- **`isAllowedOrigin` port-only check** — Intentional. When bound to `0.0.0.0`, requests come from unpredictable LAN IPs. Restricting hostname to loopback would break the feature. `AGENTMEMORY_SECRET` provides the auth layer, and `0.0.0.0` is already an explicit opt-in.
- **CSP / displayHost / YAML patching concerns** — These all assume support for arbitrary bind addresses like `192.168.x.x`, which is not a targeted use case. Only `127.0.0.1` (default) and `0.0.0.0` are the two possible values.

Happy to add input validation rejecting other values and the `resolveEngineConfig()` DRY refactor if the maintainer wants them in this PR.

rohitg00:
Thanks @Edison-A-N — `--host` / `AGENTMEMORY_HOST` is genuinely useful for the LAN-dev-server case, and the security framing (default stays `127.0.0.1` per GHSA #03) is exactly right.

Two asks before this can land:

**1. Rebase.** Branch is CONFLICTING against current `main`. Several recent merges touched `src/index.ts`, `src/cli.ts`, and `src/auth.ts`. Could you rebase onto latest main and resolve?

**2. Loud warning when `--host` is non-loopback.** Right now binding to `0.0.0.0` exposes the REST API and viewer to the LAN. If `AGENTMEMORY_SECRET` isn't set, that's a write-anywhere endpoint (the audit endpoints are public; `/agentmemory/observe` is auth-gated only when secret is set). Please add a `console.warn` on startup when the resolved bind host isn't a loopback address AND `AGENTMEMORY_SECRET` isn't set — something like:

```
[agentmemory] WARNING: bound to <host> with no AGENTMEMORY_SECRET set.
              Anyone on the network can write to memory and read observations.
              Set AGENTMEMORY_SECRET, or bind to 127.0.0.1 (the default).
```

Once the rebase is in and the warning lands, happy to merge. Good fit for Q4 ("long-running deployment story") theme.

Edison-A-N:
Thanks @rohitg00 — both items addressed in [`fcc30eb`](https://github.com/rohitg00/agentmemory/pull/194/commits/fcc30eb3c6eef08b9b4d049304604e7bbd0480ea):

**1. Rebase ✅** — rebased onto latest `main`. Only conflict was the `node:fs`/`node:os` import block in `src/cli.ts`; resolved by taking the union of symbols actually used by both sides. `mergeable` is now `MERGEABLE`.

**2. Warning ✅** — reformatted to your suggested multi-line `console.warn` template, and tightened the trigger condition while I was in there:

- Added a small `isLoopbackHost()` helper that treats `localhost`, `::1`, `::ffff:127.0.0.1`, and the full `127.0.0.0/8` block as loopback (the previous code only checked literal `'127.0.0.1'`, so e.g. `--host 127.0.0.2` or `--host localhost` would have wrongly tripped the warning *and* triggered the iii-engine config patch).
- Replaced all 3 call sites (`startEngine()` x2 + the startup warning in `main()`) with `!isLoopbackHost(hostOverride)`.

Warning output now matches your template exactly:

```
[agentmemory] WARNING: bound to <host> with no AGENTMEMORY_SECRET set.
              Anyone on the network can write to memory and read observations.
              Set AGENTMEMORY_SECRET, or bind to 127.0.0.1 (the default).
```

`npm run build` ✅ and `npm test` is 836/838 — the 2 failures are in `test/auto-compress.test.ts` and reproduce on a clean `upstream/main` checkout, so they're pre-existing and unrelated to this PR.

Ready for re-review whenever you have a minute 🙏

KShepherd-Kustoms:
Is the ability to bind all the services for Agentmemory (REST API / Etc) to 0.0.0.0 going to be available soon? This is a blocker for me with Hermes Agent as I have Hermes in it's own Container and AgentMemory in it's own container and have no way to get them both to communicate (from what I can figure out)

PR #207, opened 2026-04-26
fix(replay): preserve real sessionId + surface 200-file scan cap (#202, #203)

### LINKED ISSUE #202: `parseJsonlText` ignores file's real `sessionId`, generates fresh random ID every call — breaks merge-on-reimport, inflates session count ~4×
### Summary

`src/replay/jsonl-parser.ts` `parseJsonlText(text, fallbackSessionId)` is supposed to use the JSONL file's own `sessionId` field and fall back to `fallbackSessionId` only when the file has none. The implementation does the opposite: the fallback is assigned first, and the loop's `!sessionId` guard then prevents the file's real id from ever being adopted. Every imported session gets a freshly generated `sess_…` id. Re-running `import-jsonl` against the same file produces a *new* session every time instead of merging into the existing one.

### Repro

```bash
# 1. Pick any Claude Code JSONL transcript
SAMPLE=~/.claude/projects/-Users-vlad-repos-english-teaser-website/a5013d66-997c-4761-ad45-951f6ddf3bd5.jsonl

# 2. Confirm the file actually has a sessionId
grep -oE '"sessionId":"[^"]*"' "$SAMPLE" | sort -u
# → "sessionId":"a5013d66-997c-4761-ad45-951f6ddf3bd5"   (single, consistent)

# 3. Import the file twice
for i in 1 2; do
  curl -s -X POST http://localhost:3111/agentmemory/replay/import-jsonl \
    -d "{\"path\":\"$SAMPLE\"}" \
    | jq -r '.sessionIds[]'
done
```

Output (observed):
```
sess_moe65b3v_4ce2bddf36e7      # call 1 — generated, NOT a5013d66...
sess_moe65b8g_0ec96c6b1e5e      # call 2 — different generated id
```

Expected: both calls return `a5013d66-997c-4761-ad45-951f6ddf3bd5`, the file's real id, and the second call merges into the existing session row.

### Real-world impact

On a `~/.claude/projects` tree with **2167 JSONL files** spanning **564 unique source sessionIds**, a single clean import lands **2166 sessions** (≈1 per file) instead of ~564. Files that represent the same logical Claude Code session — resumes, sub-agent forks, branching — never merge because each file gets a fresh random id. Re-running the importer multiplies the inflation: a second pass against the same files produced 4730 sessions in our case (≈2.2× per file).

The Replay tab, sessions list, and any "show me sessions about X" workflow show the inflated count. Search/recall on observation content still works (BM25 + vectors), so this is mostly a session-metadata problem, but it's the headline `import-jsonl` UX.

### Root cause

`src/replay/jsonl-parser.ts`, in `parseJsonlText`:

```ts
export function parseJsonlText(text: string, fallbackSessionId?: string): ParsedTranscript {
  // ...
  let sessionId = fallbackSessionId || "";   // ← already populated
  // ...
  for (const entry of entries) {
    if (entry.sessionId && !sessionId) sessionId = entry.sessionId;  // !sessionId is always false
    // ...
  }
}
```

`sessionId` is initialized from `fallbackSessionId`, so on every iteration `!sessionId` is false and the entry's real `sessionId` is never adopted.

### Suggested fix

Reverse the precedence — prefer the file's id, fall back only at the end:

```ts
let sessionId = "";
// ... loop unchanged: `if (entry.sessionId && !sessionId) sessionId = entry.sessionId;`
if (!sessionId) sessionId = fallbackSessionId || generateId("sess");
```

This makes `parseJsonlText` honor the field name `fallback`SessionId, and the existing merge logic in `src/functions/replay.ts` (`kv.get<Session>(KV.sessions, parsed.sessionId)` → update existing) starts working as advertised.

A regression test in `test/replay.test.ts` covering "file with sessionId X, parser returns X" and "two parses of the same file produce the same id" would catch this from coming back.

### LINKED ISSUE #203: `agentmemory import-jsonl` CLI silently caps at 200 files; no flag to override; no warning when truncated
### Summary

`agentmemory import-jsonl` (the CLI users actually run, per the README's "Already have older Claude Code JSONL transcripts you want to bring in?" section) walks the target directory with a hard-coded default `limit = 200`. The CLI doesn't expose a `--max-files` flag, the help text doesn't mention any cap, and the post-import summary doesn't warn when files were skipped. On any non-trivial `~/.claude/projects` tree the user gets a silent ~10% import.

The underlying API endpoint already accepts a `maxFiles` field — the cap is purely a CLI/UX limitation.

### Repro

```bash
# Tree with 2167 JSONL files
find ~/.claude/projects -name '*.jsonl' | wc -l
# → 2167

agentmemory import-jsonl
# CLI shows "scanning files…" then "View at http://localhost:3113 → Replay tab"
# No file count, no warning.

curl -s localhost:3111/agentmemory/sessions?limit=10000 | jq '.sessions | length'
# → 199    (≈ the 200 cap, after dedupe)
```

Workaround (only works if you know the API exists):

```bash
curl -s -X POST http://localhost:3111/agentmemory/replay/import-jsonl \
  -d '{"path": "~/.claude/projects", "maxFiles": 5000}'
# → success: true, but iii function timed out at 30s on a large tree
```

So the practical override path for any large tree is "iterate per project subdirectory and call the API directly" — not something a normal CLI user will discover.

### Root cause

`src/functions/replay.ts`:

```ts
async function findJsonlFiles(root: string, limit = 200): Promise<string[]> {
  // ...
  if (out.length >= limit) return;
  // ...
}
// ...
files = await findJsonlFiles(abs, data.maxFiles || 200);
```

The 200 default is an arbitrary safety cap — sensible for an HTTP endpoint with a 30s function timeout, but the CLI inherits it without ever exposing it or warning when it bites.

### Suggested fix

Three small changes that compound:

1. **Expose the flag in the CLI.** Add `--max-files <N>` to `agentmemory import-jsonl` (default e.g. 5000 to match real-world `~/.claude/projects` sizes). Update the help text in `src/cli.ts`.
2. **Warn when the cap is hit.** In the import summary, if `files.length === limit`, print: `⚠ Hit the ${limit}-file cap; ${found - imported} files were skipped. Re-run with --max-files=… or batch by subdirectory.`
3. **Document the per-subdir batching pattern.** A 200-file cap *does* make sense given the iii-engine 30s function timeout — under load a single 2000-file pass times out. The README's import section could mention "for very large trees, iterate by subdirectory" with a one-liner. Or, even better, have the CLI auto-batch when it detects a tree larger than `--max-files`.

### Sibling bug

This compounds badly with #(parser-sessionId-bug — also filed today). Each re-run with a different `--max-files` creates *new* phantom sessions instead of merging, so users who hit the silent cap and re-run end up with wildly inflated session counts. Together they turn `import-jsonl` from a one-shot operation into "1/10th imported, the rest invisible, and any retry doubles the noise."

rohitg00:
## Summary

Two bugs in the JSONL import path that combined to make \`agentmemory import-jsonl\` against a real \`~/.claude/projects\` tree quietly mis-import: ~10% of files scanned, each landing under a fresh random session id even when the file's own header named one.

### #202 — \`parseJsonlText\` ignores the file's sessionId

The parser pre-populated \`sessionId\` from \`fallbackSessionId\` and then guarded \`if (entry.sessionId && !sessionId)\`, so the file's real id was never adopted. Each call returned a freshly generated session, breaking merge-on-reimport and inflating session counts ~4× on real trees.

Fix: don't pre-populate \`sessionId\`; fall back to \`fallbackSessionId\` then \`generateId('sess')\` only after the loop.

### #203 — \`import-jsonl\` CLI silently caps at 200 files

\`findJsonlFiles\` had a hard 200-file cap (sensible for the 30s function timeout) but the CLI exposed no flag to override it and never warned when files were skipped. On a 2167-file tree users got a silent ~10% import.

Fix:
- \`findJsonlFiles\` now reports \`{files, truncated, discovered}\` so the handler can count beyond the cap without retaining the paths
- Handler returns \`discovered\`, \`truncated\`, \`maxFiles\` in the response
- CLI accepts \`--max-files <N>\` and \`--max-files=<N>\`
- CLI warns on truncation, suggests a larger cap

## Tests

- 3 new regression cases in \`test/replay.test.ts\`: file sessionId beats fallback, repeated parses produce the same id, fallback only when file has no id
- Full suite: 830/830 passing

Thanks @bloodcarter for the precise repros and root-cause pointers in #202/#203.

Closes #202
Closes #203

## Test plan

- [ ] \`agentmemory import-jsonl ~/.claude/projects\` on a tree >200 files — must surface a warning + skipped count
- [ ] \`agentmemory import-jsonl --max-files 5000 ~/.claude/projects\` — must scan beyond 200
- [ ] Re-run import twice on the same file — must merge into one Session (no duplicates)

## Summary by CodeRabbit

* **New Features**
  * CLI: added optional --max-files (both forms); imports now report discovered/skipped counts, truncation/traversal caps, and recommend batching or a higher max within the server-provided upper bound.
* **Bug Fixes**
  * Fixed CLI argument parsing so long-form flags consume their values correctly.
  * Enhanced server-side validation of maxFiles with clearer range errors; invalid CLI values emit warnings and are ignored.
* **Tests**
  * Added tests covering session ID precedence and fallback behavior.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -36,7 +36,8 @@ Commands:
   upgrade            Upgrade local deps + iii runtime (best effort)
   mcp                Start standalone MCP server (no engine required)
   import-jsonl [p]   Import Claude Code JSONL transcripts (default: ~/.claude/projects)
-                     Use --max-files <N> or --max-files=<N> to override the 200-file scan cap (default: 200)
+                     --max-files <N> | --max-files=<N>: override scan cap (default 200, max 1000;
+                     out-of-range is rejected; for trees >1000 files, batch by subdirectory)
 
 Options:
   --help, -h         Show this help
@@ -961,6 +962,11 @@ async function runMcp(): Promise<void> {
 }
 
 async function runImportJsonl(): Promise<void> {
+  // Long-form flags that take a value. Their value tokens must be
+  // consumed alongside the flag so they don't leak into positional
+  // args (e.g. `--port 3112 import-jsonl` would otherwise turn
+  // 3112 into pathArg).
+  const VALUE_FLAGS = new Set(["--port", "--tools"]);
   let maxFiles: number | undefined;
   const tail = args.slice(1);
   const positional: string[] = [];
@@ -985,6 +991,10 @@ async function runImportJsonl(): Promise<void> {
       } else {
         p.log.warn(`Ignoring --max-files=${raw}: expected a positive integer.`);
       }
+      continue;
+    }
+    if (VALUE_FLAGS.has(a)) {
+      i++;
       continue;
     }
     if (a.startsWith("-")) continue;
@@ -1046,7 +1056,9 @@ async function runImportJsonl(): Promise<void> {
       observations?: number;
       discovered?: number;
       truncated?: boolean;
+      traversalCapped?: boolean;
       maxFiles?: number;
+      maxFilesUpperBound?: number;
     } = {};
     if (text.length > 0) {
       try {
@@ -1086,12 +1098,30 @@ async function runImportJsonl(): Promise<void> {
     );
     if (json.truncated) {
       const cap = json.maxFiles ?? 200;
-      const skipped = (json.discovered ?? 0) - (json.imported ?? 0);
-      p.log.warn(
-        `Hit the ${cap}-file scan cap; ${skipped} of ${json.discovered ?? "?"} discovered file(s) were skipped. ` +
-          `Re-run with --max-files=<N> (e.g. --max-files=${Math.max((json.discovered ?? cap) + 100, cap * 2)}) ` +
-          `or batch by subdirectory.`,
-      );
+      const upper = json.maxFilesUpperBound ?? 1000;
+      const discovered = json.discovered ?? 0;
+      const skipped = discovered - (json.imported ?? 0);
+      const discoveredLabel = json.traversalCapped
+        ? `${discovered}+ (traversal halted at safety cap)`
+        : String(discovered);
+      const baseMsg = `Hit the ${cap}-file scan cap; ${skipped} of ${discoveredLabel} discovered file(s) were skipped.`;
+      // If we already saw more than the server's hard cap (or the
+      // walker stopped early), bumping --max-files won't help on its
+      // own — recommend batching by subdirectory.
+      if (discovered > upper || json.traversalCapped) {
+        p.log.warn(
+          `${baseMsg} Tree exceeds the server's --max-files limit of ${upper}; ` +
+            `batch by subdirectory (run import-jsonl once per project under ~/.claude/projects).`,
+        );
+      } else {
+        const suggested = Math.min(
+          Math.max((discovered || cap) + 100, cap * 2),
+          upper,
+        );
+        p.log.warn(
+          `${baseMsg} Re-run with --max-files=${suggested} (max ${upper}) or batch by subdirectory.`,
+        );
+      }
     }
     if (json.sessionIds && json.sessionIds.length > 0) {
       p.log.info(`View at ${getViewerUrl()} → Replay tab`);
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -36,6 +36,7 @@ Commands:
   upgrade            Upgrade local deps + iii runtime (best effort)
   mcp                Start standalone MCP server (no engine required)
   import-jsonl [p]   Import Claude Code JSONL transcripts (default: ~/.claude/projects)
+                     Use --max-files <N> to override the 200-file scan cap (default: 200)
 
 Options:
   --help, -h         Show this help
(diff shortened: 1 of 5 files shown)


Qodo-Free-For-OSS:
Hi, findJsonlFiles() enforces its traversal cap using `discovered`, but `discovered` is only incremented for `.jsonl` files, so trees with many non-`.jsonl` entries can still be walked almost entirely and hit function timeouts.

**How to fix:** Cap by visited entries

> ### Issue description
> `findJsonlFiles()` attempts to bound filesystem traversal via `traversalCap`, but it uses `discovered` (count of `.jsonl` files) to enforce that cap. In directories with many non-`.jsonl` files, the traversal can still run extremely long and may exceed function timeouts.
> 
> ### Issue Context
> We still want:
> - `discovered`: number of `.jsonl` files encountered (for reporting)
> - `files`: first `limit` `.jsonl` files
> - a hard bound on total work regardless of file extension
> 
> ### Fix Focus Areas
> - src/functions/replay.ts[203-242]
>   - Introduce a separate counter (e.g., `visitedEntries` / `walked`) incremented per directory entry processed (or per `lstat` attempt), and use THAT for the `traversalCap` checks.
>   - Keep `discovered++` for `.jsonl` only so reporting remains correct.
>   - Consider setting `truncated` true when the traversal cap is hit even if `discovered <= out.length` (i.e., stopped early for safety).

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers.*

Qodo-Free-For-OSS:
Hi, runImportJsonl() treats any non-dash token as a positional path, so values belonging to other flags (e.g. `--port 3112` or `--tools core`) can be incorrectly consumed as the import path.

**Severity:** remediation recommended | **Category:** correctness

**How to fix:** Skip values for known flags

> ### Issue description
> `runImportJsonl()` builds `positional` args by skipping only tokens that start with `-`. This misclassifies flag *values* (like the `3112` after `--port`) as positional arguments and can set `pathArg` incorrectly.
> 
> ### Issue Context
> The CLI already supports global flags like `--port` and `--tools` by scanning `args` earlier and setting env vars, but those tokens remain in `args` and are re-parsed by `runImportJsonl()`.
> 
> ### Fix Focus Areas
> - src/cli.ts[963-994]
>   - When encountering known flags that take a value (e.g., `--port`, `--tools`), increment `i` to skip the next token.
>   - Alternatively, refactor to a small shared arg parser that returns `{ positionals, flags }` and properly consumes flag values.
>   - Add a regression test (if CLI has tests) or at least manual note in PR description with example invocations.

---
*Found by [Qodo](https://qodo.ai) code review*

PR #209, opened 2026-04-27
fix: prevent process crash on iii-engine state::set timeout (#204)

### LINKED ISSUE #204: Service crashes with uncaught IIIInvocationError TIMEOUT on state::set (v0.9.3)
## Summary

The agentmemory service (`@agentmemory/agentmemory@0.9.3`) crashes intermittently with an uncaught `IIIInvocationError: TIMEOUT: invocation timed out after 30000ms` on `function_id: 'state::set'`. The rejection escapes through `iii-sdk/dist/index.mjs:405` and terminates the Node process. Under sustained write load (passive observation capture from CC hooks across multiple projects) we observed 5–15 crashes/hour. systemd auto-restart recovers in ~10s, but in-flight BM25/vector index updates that hadn't fired their 5s `IndexPersistence` debounce are lost across the crash boundary, breaking `memory_smart_search` recall for very recent saves.

## Environment

- Node.js v20.20.2
- Linux 6.8.0-106-generic, x86_64
- `@agentmemory/agentmemory` v0.9.3 installed via `npm install -g`
- `iii` engine v0.11.0 native binary
- Embedding provider: `openai`
- `AGENTMEMORY_AUTO_COMPRESS=true`, `CONSOLIDATION_ENABLED=true`, `GRAPH_EXTRACTION_ENABLED=true`
- Load profile: 5 Claude Code agents + ~1.7K observations/day (~75/hour avg) via plugin hooks

## Reproduction

1. Run service with the env vars above
2. Drive sustained write load: ~1 observation/sec via `POST /agentmemory/observe` (or via the plugin hooks under active CC sessions)
3. Within 1–10 minutes, the process exits with the trace below

## Crash log

```
[agentmemory] Ready. Triple-stream (BM25+Vector+Graph) search active.
... (40–90 s of normal operation, observations captured/compressed) ...
file:///usr/lib/node_modules/@agentmemory/agentmemory/node_modules/iii-sdk/dist/index.mjs:405
                                                reject(new IIIInvocationError({
                                                       ^

IIIInvocationError: TIMEOUT: invocation timed out after 30000ms
    at Timeout._onTimeout (file:///.../iii-sdk/dist/index.mjs:405:14)
    at listOnTimeout (node:internal/timers:581:17)
    at process.processTimers (node:internal/timers:519:7) {
  code: 'TIMEOUT',
  function_id: 'state::set',
  stacktrace: undefined
}

Node.js v20.20.2
agentmemory.service: Main process exited, code=exited, status=1/FAILURE
```

The same crash repeats with restart counter climbing (we see >40 starts/24h on a busy day).

## What we expect

The `state::set` invocation timing out shouldn't crash the whole process. Either:

- Catch and log the rejection (degrade gracefully — drop or queue the write), or
- Surface a configurable longer timeout / retry policy for KV writes, or
- Add `process.on('unhandledRejection', …)` at the entrypoint as a hard floor.

## Side effects observed

1. Recent BM25 / vector index additions (since the last `IndexPersistence.scheduleSave()` debounce flush — 5s) are lost across the crash, since they live only in memory until persisted via `state::set(KV.bm25Index, …)`.
2. `memory_smart_search` doesn't return content saved within ~30s of the crash, even though `kv.set(KV.memories, …)` itself completed.
3. Even after restart, OTel WebSocket reconnect loop ("WebSocket error: Unexpected server response: 404") spams logs with exponential backoff up to ~30s.

## Possibly related

- Whether the iii engine has an internal queue limit / write backpressure that surfaces as a 30s timeout under load — happy to share a journalctl dump if useful.
- `function_id: 'state::set'` is the only function_id we see crash; `state::get` and others time out gracefully.

## Suggested fixes (any one helps)

- [ ] Wrap `state::set` calls in `IndexPersistence.save()` and `kv.set(...)` paths with a `.catch()` that logs + drops, instead of letting the rejection propagate.
- [ ] Top-level `process.on('unhandledRejection', …)` handler in the service entrypoint so a single SDK timeout doesn't take down the whole memory mesh.
- [ ] Document the recommended `iii-engine` tuning for sustained-write workloads (e.g. `AGENTMEMORY_OBSERVE_QUEUE_LIMIT=…`) — if such knobs exist.

Happy to PR a `.catch()` wrapper if you point me at the preferred location.

rohitg00:
## Summary

Under sustained write load (Claude Code plugin hooks across multiple agents at ~75+ obs/hour), \`IndexPersistence\`'s 5s debounce flush could fire into an iii-engine queue that hadn't drained, hitting the SDK's 30s default timeout on \`state::set\`. The rejection escaped the \`setTimeout(() => this.save(), DEBOUNCE_MS)\` callback in \`src/state/index-persistence.ts\` because the returned promise was discarded — landing as an \`unhandledRejection\` that terminated the long-lived memory service.

Reproducible at ~5–15 crashes/hour on ~1.7K observations/day workloads (see #204 for trace).

## Two layers of defense

1. **\`IndexPersistence\` swallows kv.set rejections internally.** Scheduled-save callback now \`.catch()\`s the returned promise; \`save()\` wraps kv.set in try/catch. Failures logged via \`logger.warn\` with once-per-minute throttle so queue-pressure bursts don't spam the log. Recent index updates stay in memory and retry on next debounce flush.

2. **Top-level \`unhandledRejection\` handler in \`src/index.ts\`.** Logs and continues for any other path we missed — defense in depth, also throttled.

## Tests

- 2 new regression cases in \`test/index-persistence.test.ts\`: scheduled save with rejecting kv.set must not raise unhandledRejection; \`save()\` must resolve (not throw) when kv.set rejects
- Full suite: 829/829 passing

Thanks @bunke for the painstaking writeup — the journalctl trace pointing at \`iii-sdk/dist/index.mjs:405\` made the root cause obvious.

Closes #204

## Test plan

- [ ] Drive ~1 observation/sec via \`POST /agentmemory/observe\` for 10+ minutes — service must not exit
- [ ] Verify the throttled \`[agentmemory] warn index persistence: failed to save\` log appears under timeout pressure

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process stability by preventing crashes from unhandled asynchronous errors
  * Enhanced data persistence error handling with better recovery mechanisms and logging
  * Added throttled error messages to help diagnose timeout and persistence issues without log spam

* **Tests**
  * Added test coverage for error handling during data persistence operations

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -104,6 +104,25 @@ function hasGetMeter(
   );
 }
 
+// Top-level safety net for iii-engine invocation timeouts (issue #204).
+// Under sustained write load (e.g. Claude Code hooks across many
+// projects) `state::set` can occasionally exceed the SDK's 30s timeout.
+// We don't want one such timeout to terminate the long-lived memory
+// service — the rejection is surfaced to the relevant call site via
+// .catch() where it matters; everything else is logged-and-continued.
+// Throttle logs to avoid spamming on bursts.
+let lastUnhandledLogAt = 0;
+process.on("unhandledRejection", (reason) => {
+  const now = Date.now();
+  if (now - lastUnhandledLogAt < 60_000) return;
+  lastUnhandledLogAt = now;
+  const r = reason as { code?: string; function_id?: string; message?: string };
+  console.warn(
+    `[agentmemory] unhandledRejection (suppressed):`,
+    r?.code ? `${r.code} ${r.function_id ?? ""} ${r.message ?? ""}`.trim() : reason,
+  );
+});
+
 async function main() {
   const config = loadConfig();
   const embeddingConfig = loadEmbeddingConfig();
--- a/src/state/index-persistence.ts
+++ b/src/state/index-persistence.ts
@@ -2,11 +2,14 @@ import { SearchIndex } from "./search-index.js";
 import { VectorIndex } from "./vector-index.js";
 import type { StateKV } from "./kv.js";
 import { KV } from "./schema.js";
+import { logger } from "../logger.js";
 
 const DEBOUNCE_MS = 5000;
+const FAILURE_LOG_THROTTLE_MS = 60_000;
 
 export class IndexPersistence {
   private timer: ReturnType<typeof setTimeout> | null = null;
+  private lastFailureLogAt = 0;
 
   constructor(
     private kv: StateKV,
@@ -16,7 +19,13 @@ export class IndexPersistence {
 
   scheduleSave(): void {
     if (this.timer) clearTimeout(this.timer);
-    this.timer = setTimeout(() => this.save(), DEBOUNCE_MS);
+    // setTimeout discards the returned promise, so any rejection inside
+    // save() would surface as unhandledRejection and crash the process
+    // under sustained iii-engine write timeouts (issue #204). Funnel
+    // rejections through logFailure() instead.
+    this.timer = setTimeout(() => {
+      this.save().catch((err) => this.logFailure(err));
+    }, DEBOUNCE_MS);
   }
 
   async save(): Promise<void> {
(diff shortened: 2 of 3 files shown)


bunke:
Thanks!

PR #218, opened 2026-04-30
fix(viewer): prevent tab bar crush and fix graph canvas overflow

nagarjunr:
## Summary

Two CSS layout bugs in the single-file viewer SPA (`src/viewer/index.html`):

### Bug 1 — Tab bar compressed to ~8px (invisible)

**Root cause:** `.tab-bar` has `overflow-x: auto` for horizontal scrolling on small screens. The CSS overflow interaction rule means setting `overflow-x` to anything other than `visible` also forces `overflow-y: auto`. This zeroes the flex item's minimum height floor. Combined with the default `flex-shrink: 1` on all structural elements, the flexbox algorithm shrinks the tab bar to near-zero when content is tall.

**Fix:** Add `flex-shrink: 0` to all structural chrome elements (`.app-header`, `.tab-bar`, `.flag-banners`, `.viewer-footer`). Only the content view (`.view.active`, which already has `min-height: 0`) should participate in shrinking.

### Bug 2 — Graph canvas overflows its container (~235px)

**Root cause:** `.graph-container` used `height: calc(100vh - 130px)` which assumed 130px of fixed chrome above it. Since the flag-banners section was added the actual chrome is ~194px, making the canvas 64-235px taller than the visible view area. Users see only ~62% of the graph canvas.

**Fix:** Add `#view-graph { padding: 0; overflow: hidden }` and change `.graph-container` to `height: 100%`, removing the stale `calc()` and the `margin: -24px` padding-negation hack. Since the view is a flex item with `flex-grow: 1` the height resolves correctly against the available viewport space.

## Test plan

- [ ] Tab bar is fully visible at all viewport heights
- [ ] Switching tabs no longer crushes the tab bar
- [ ] Expanding the Feature Flags banner does not overlap the tab bar
- [ ] Graph canvas fills its container without overflow scrollbar
- [ ] Graph interaction (pan/zoom/drag) works within the full visible area
- [ ] All other tab views scroll normally within their content area

## Summary by CodeRabbit

* **Style**
  * Improved viewer layout spacing and alignment for better visual consistency
  * Enhanced graph tab display sizing to better utilize available space

### DIFF
--- a/src/viewer/index.html
+++ b/src/viewer/index.html
@@ -96,6 +96,7 @@
       align-items: center;
       justify-content: space-between;
       background: var(--bg);
+      flex-shrink: 0;
     }
     .app-header .brand {
       display: flex;
@@ -150,6 +151,7 @@
       border-bottom: 1px solid var(--border-light);
       background: var(--bg);
       overflow-x: auto;
+      flex-shrink: 0;
     }
     .tab-bar button {
       background: none;
@@ -346,10 +348,13 @@
     .btn-primary { background: var(--ink); color: var(--bg); border-color: var(--ink); }
     .btn-primary:hover { background: var(--ink-secondary); box-shadow: 3px 3px 0px 0px var(--ink-muted); }
 
+    #view-graph {
+      padding: 0;
+      overflow: hidden;
+    }
     .graph-container {
       display: flex;
-      height: calc(100vh - 130px);
-      margin: -24px;
+      height: 100%;
       border-top: 1px solid var(--border-light);
     }
     .graph-canvas-wrap {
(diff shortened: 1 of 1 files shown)


rohitg00:
Reviewed — clean CSS layout fix. The diagnoses are right:

- Tab bar overflow without `flex-shrink: 0` was crushing into the content area when sessions accumulate.
- Graph canvas using `calc(100vh - 130px)` with negative margins was a brittle hack; switching to `height: 100%` inside the `#view-graph` container makes it actually responsive.
- The footer + flag-banner `flex-shrink: 0` additions prevent the same crush pattern in those slots.

CSS-only, no runtime impact, no test coverage needed. Approving for merge — holding for @rohitg00 to push the button.

If you have screenshots of before/after the tab-bar fix specifically, mind dropping them in the PR body? Helpful for the CHANGELOG entry. Not blocking.

ZhangSanMo:
same issues, wait fix.

rohitg00:
I'm not totally sure this PR fixes it. I'll check today. We're swamped with PRs and commits that need a good look before merging.

Issue #143, opened 2026-04-14
bursting token allocation (re-open)


adrianricardo:
I installed a couple hours ago (after the PR was merged) but had a similar experience to https://github.com/rohitg00/agentmemory/issues/138 . I hit my usage limit from 0 in just 4 messages (claude pro)

rohitg00:
## What the Claude Code docs actually say

From [code.claude.com/docs/en/hooks.md](https://code.claude.com/docs/en/hooks.md):

> "For most events, stdout is written to the debug log but not shown in the transcript. The exceptions are \`UserPromptSubmit\` and \`SessionStart\`, where stdout is added as context that Claude can see and act on."

**PreToolUse is not an exception.** Plain stdout from a PreToolUse hook goes to the debug log and is never injected into the model's input context. The only way PreToolUse can push context into the model is via JSON output with \`hookSpecificOutput.additionalContext\`, which agentmemory does not do — it writes plain text. Same for PreCompact.

So my \`20 × 1000 tokens per message\` math was fiction. The pre-tool-use hook is a resource hog in terms of Node process spawns and REST round-trips to \`/agentmemory/enrich\`, but it's not burning your Pro tokens.

## What actually affects tokens

- **\`SessionStart\` stdout IS injected into the model context** (docs explicitly list it as the exception). So agentmemory is pushing ~1000-2000 tokens into the first model turn of every new session you start. Real, but it's once per session, not per tool call. On a 4-message conversation this is ~1-2K tokens total, not 80K.

## What's probably the real cause

The Register ran a piece on March 31 ([link](https://www.theregister.com/2026/03/31/anthropic_claude_code_limits/)) in which Anthropic confirmed: *"people are hitting usage limits in Claude Code way faster than expected. It's the top priority."* Claude Pro's quotas are documented as deliberately tight and were not sized for heavy Claude Code tool-call-per-message usage. A 4-message burn on Pro with zero plugins installed is a very plausible baseline right now.

I still believe agentmemory is contributing (~1-2K tokens per session via SessionStart), and I'm still going to ship a fix for that. But I don't want to claim that fix will restore your Pro allocation to normal when Anthropic is actively wrestling with the underlying Pro-cap sizing problem.

## Honest recommendations

1. **Short-term, if you're blocked**: \`/plugin disable agentmemory\` in Claude Code. That kills the SessionStart injection immediately. Saves ~1-2K tokens per session you start. Won't save you from the base Claude Pro tightness.
[...]

PR #222, opened 2026-05-01
fix: make session-start / subagent-start hooks fire-and-forget when no read needed (#221)

### LINKED ISSUE #221: session-start / subagent-start hooks block agent startup for up to 5s on slow/unreachable REST
### Problem

Two hook scripts block Claude Code's session/subagent startup waiting on REST responses they don't actually need:

**`src/hooks/session-start.ts`** — `await fetch(...)` with a 5000 ms timeout on every session start. When `AGENTMEMORY_INJECT_CONTEXT=false` (the default since 0.8.10, per #143) the response is discarded — `INJECT_CONTEXT && res.ok` is false, the body is never read. So the await is pure latency: if the REST endpoint is slow or unreachable, every session start eats up to 5 s of wall time before the agent can do its first turn.

**`src/hooks/subagent-start.ts`** — has a `// fire and forget` comment in the catch block, but the code actually `await`s a 2000 ms POST. The hook is purely passive telemetry; nothing reads the response. The comment and the code disagree.

### Impact

This becomes painful in any setup that spawns Claude sessions concurrently:

- Slack-bot orchestrators (one CC subprocess per Slack thread)
- Multi-agent harnesses where one parent dispatches several child agents
- CI / scripting that fans out `claude -p` jobs

When `agentmemory` is healthy, latency is ~50–150 ms per hook and unnoticeable. When it slows down (overloaded `iii-engine`, paused for compaction, etc.), every concurrent session start stacks the awaited timeout, which feeds back into the engine's load — a positive-feedback loop. We hit it on a host running ~10 parallel `claude -p` subprocesses; engine RSS climbed past 2 GB, OS OOM-killed it, systemd restarted, the `claude -p` subprocesses were still alive and immediately re-flooded the new instance. The 5 s blocked sessions were the multiplier.

### Reproduction

```bash
# Simulate an unreachable REST endpoint
time (echo '{"session_id":"x","cwd":"/tmp"}' | \
  AGENTMEMORY_URL=http://127.0.0.1:1 \
  AGENTMEMORY_INJECT_CONTEXT=false \
  node plugin/scripts/session-start.mjs)
# Expected: <500 ms (fire-and-forget when no inject)
# Observed: ~5000 ms (timeout)

time (echo '{"session_id":"x","cwd":"/tmp","agent_id":"a","agent_type":"b"}' | \
  AGENTMEMORY_URL=http://127.0.0.1:1 \
  node plugin/scripts/subagent-start.mjs)
# Expected: <500 ms (claims fire-and-forget)
# Observed: ~2000 ms (timeout)
```

### Proposed fix

- `session-start.ts` — fire-and-forget when `INJECT_CONTEXT=false`. When `INJECT_CONTEXT=true` keep the await (response is needed for stdout) but tighten the timeout (5000 → 1500 ms) so a slow server can't block the agent indefinitely.
- `subagent-start.ts` — actually fire-and-forget, matching its own comment. Drop timeout 2000 → 800 ms as a defensive cap.
- Drive-by: both files inline a copy of `isSdkChildContext`; `src/hooks/sdk-guard.ts` already exports it. Consolidate.

Happy to send a PR with tests measuring exit time under a black-hole `AGENTMEMORY_URL`. Filing this first per CONTRIBUTING ("Link the issue the PR resolves").

### Environment

- `agentmemory` 0.9.3 (also visible in 0.9.4 main on inspection)
- Node 20+
- Linux x86_64

bunke:
## Summary

Two startup hooks awaited POSTs whose responses they don't actually use:

- `src/hooks/session-start.ts` — `await fetch(...)` with a 5000 ms timeout on every session start. When `AGENTMEMORY_INJECT_CONTEXT=false` (the default since 0.8.10, per #143) the response is discarded — the body is never read. The await is pure latency: if the REST endpoint is slow or unreachable, every session start eats up to 5 s of wall time before the agent can do its first turn.
- `src/hooks/subagent-start.ts` — has a `// fire and forget` comment in the catch block, but the code actually `await`s a 2000 ms POST. The hook is purely passive telemetry; nothing reads the response. Comment and code disagreed.

Under load (parallel `claude -p` fan-out, e.g. a Slack-bot orchestrator dispatching many sessions) the awaited timeouts stacked up and starved the engine. We hit it on a host running ~10 parallel `claude -p` subprocesses; engine RSS climbed past 2 GB, OS OOM-killed it, systemd restarted, the subprocesses were still alive and immediately re-flooded the new instance. The 5 s blocked sessions were the multiplier.

## Three changes

1. **`session-start.ts`** — when `INJECT_CONTEXT=false`, `main()` returns immediately after firing the request; node exits when the bg request resolves or aborts. When `INJECT_CONTEXT=true` we still await (response body is needed for stdout) but with a tightened **1500 ms** ceiling instead of 5000 ms.
2. **`subagent-start.ts`** — actually fire-and-forget, matching its own comment. Timeout shrunk **2000 → 800 ms** as a defensive cap.

3. **`isSdkChildContext` helper + tsdown bundling** — the function now lives in a single canonical file `src/hooks/sdk-guard-internal.ts` (deliberately *outside* `hookEntries` in tsdown.config.ts). The three hooks that use it (`session-start.ts`, `subagent-start.ts`, `stop.ts`) and the recursion-guard test all import from it. To keep each compiled hook self-contained — i.e. NOT have rolldown extract the shared helper into a hash-named chunk like `sdk-guard-internal-<HASH>.mjs`, which is the exact failure mode #221 caused — `tsdown.config.ts` now builds each hook in its own single-entry pass with `outputOptions: { codeSplitting: false }`. Verified: `head dist/hooks/stop.mjs` shows the function inlined; `ls dist/hooks plugin/scripts | grep -E '\-[A-Za-z0-9_]{8,}\.mjs'` is empty.

## Tests

`test/hooks-fire-and-forget.test.ts` (new) — spins up a stalling HTTP server (accepts the connection and request body but never sends a response, forcing `fetch` to wait for its `AbortSignal.timeout`) and spawns the built hooks against it. Connection-refused URLs would resolve too fast to distinguish old vs. new behaviour. Three cases:

- `session-start` with `INJECT_CONTEXT` unset (default off) — `main()` returns without awaiting; total exit time **< 2000 ms** (would have been ≥ 5000 ms on old code)
- `session-start` with `INJECT_CONTEXT=true` — exits between **1000 ms and 3500 ms**, honouring the new 1500 ms cap
- `subagent-start` — fire-and-forget; exits **< 2000 ms** (was ≥ 2000 ms on old code)

Observed locally: 849 ms / 1576 ms / 852 ms.

## Risk

- When `INJECT_CONTEXT=true` and the server is slow, sessions may now start without prior-session context if the response doesn't arrive within 1.5 s. Trade-off accepted — context injection is a quality-of-life feature, not correctness.
- `subagent_start` observation events may be lost if the server is unreachable (best-effort telemetry; same as the existing silently-swallowed catch).
- Fire-and-forget caveat: with `INJECT_CONTEXT=false`, the hook process keeps running until the bg request resolves or aborts. The 800 ms timeout caps that wall-clock cost. The agent itself isn't blocked on it — Claude Code reads stdout once, which we're not writing to — but the hook process exit isn't instant.

## Test plan

- [ ] CI green on the new `hooks-fire-and-forget` suite
- [ ] No regression in existing `context-injection` suite (same spawn pattern; both hooks still register the session for observation tracking)
- [ ] Full vitest run (`npm test`): no regressions vs. main

Closes #221

## Summary by CodeRabbit

* **Performance**
  * Faster startup by making session and telemetry calls non-blocking when possible
  * Reduced subagent initialization timeout to 800ms

* **Bug Fixes**
  * More robust handling of slow or failing network calls during initialization to prevent hangs

* **Tests**
  * Added tests validating fire-and-forget behavior under slow REST dependencies


Qodo-Free-For-OSS:
Hi, plugin hook scripts now import `./sdk-guard-DI1NUOS9.mjs`, a hash-named build artifact that will change across rebuilds, causing runtime module resolution failures for session-start/subagent-start.

**How to fix:** Use stable sdk-guard module

> ### Issue description
> `plugin/scripts/session-start.mjs` and `plugin/scripts/subagent-start.mjs` import a hash-named helper (`sdk-guard-DI1NUOS9.mjs`). Hash-named chunks are not stable across rebuilds, so future rebuilds can break these hooks at runtime with `ERR_MODULE_NOT_FOUND`.
> 
> ### Issue Context
> Other compiled hook scripts inline `isSdkChildContext` rather than importing a shared chunk, likely to avoid exactly this build-churn problem.
> 
> ### Fix Focus Areas
> - plugin/scripts/session-start.mjs[1-5]
> - plugin/scripts/subagent-start.mjs[1-5]
> - plugin/scripts/sdk-guard-DI1NUOS9.mjs[1-30]
> 
> ### What to change
> - Avoid importing a hash-named module from runtime hook scripts.
> - Options:
>   - Re-inline the `isSdkChildContext` function in these two compiled scripts (consistent with other hooks), or
>   - Ensure the shared guard is emitted with a stable filename (e.g., `sdk-guard.mjs`) and update imports accordingly.
> - If keeping a shared module, ensure build tooling consistently emits it and that imports are stable across rebuilds.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

bunke:
Thanks for the catch — pushed b06073c which inlines `isSdkChildContext` in `session-start.ts` / `subagent-start.ts` (matching the convention of the 11 other hooks in the repo) and removes `src/hooks/sdk-guard.ts` (no remaining consumers). Rebuild now emits no `sdk-guard-*.mjs` chunk and each `plugin/scripts/*.mjs` hook is fully self-contained, so future rebuilds cannot break the import. PR test suite still green.

rohitg00:
Thanks @bunke — this is a great catch and the trace from the 10-parallel `claude -p` fan-out + OOM-restart cycle is exactly the kind of scenario built-in plugin hooks have to survive. Tests are well-shaped (the stalling HTTP server pattern is the right way to assert this — connection-refused would fire too fast to distinguish behavior).

Two issues to address before merge:

**1. Test breakage from the `sdk-guard.ts` deletion.**

`test/stop-hook-recursion-guard.test.ts` line 2 imports `isSdkChildContext` from `../src/hooks/sdk-guard.js`. Once you delete that file, this test fails to compile. Either:

- Re-export `isSdkChildContext` from one of the surviving hook files, OR
- Move the canonical definition into a non-hook location (e.g. `src/hooks/sdk-guard-internal.ts`) that's *not* picked up as a hook entry by tsdown — the per-hook bundling rule is what triggers the shared-chunk emission, so a util file outside the hooks-entry list should be fine.

The second option keeps DRY without re-introducing the build break you're solving.

**2. Stale references after deletion.**

- `src/hooks/stop.ts:3` says "Inlined — see src/hooks/sdk-guard.ts for canonical version." That file no longer exists. Update the comment to point at wherever the canonical version lives now (or drop the reference if you go full-inline).
- `src/hooks/stop.ts:36` says "would re-enter agent-sdk provider and loop (see sdk-guard.ts)." Same fix.

**3. PR body is inverted.**

The body says: *"They now import it from the shared `src/hooks/sdk-guard.ts`."* The diff does the **opposite** — deletes `sdk-guard.ts` and keeps inline copies. The inline-copy direction is correct (the tsdown shared-chunk reasoning in the new comment is solid), but the body needs to be flipped before merge so future blame on the change isn't confusing.

Once those land, the perf fix itself is clean and I'll approve. The 5 s → 1.5 s ceiling on `INJECT_CONTEXT=true` and the fire-and-forget conversion are both the right calls.

bunke:
@rohitg00 addressed the three review points — pushed two new commits:

**Polish 1 — `isSdkChildContext` DRY without re-introducing the build break.**
New canonical file `src/hooks/sdk-guard-internal.ts` (deliberately outside `hookEntries` in `tsdown.config.ts`). The three hooks that need it (`session-start.ts`, `subagent-start.ts`, `stop.ts`) and `test/stop-hook-recursion-guard.test.ts` all import from it.

I want to flag one nuance: putting the helper outside `hookEntries` is necessary but on its own *not sufficient*. Rolldown still extracts any module imported by 2+ entries into a hash-named shared chunk (verified empirically — got `sdk-guard-internal-NyA2RU0s.mjs` on the first attempt), which is the exact breakage #221 calls out. To actually keep each hook self-contained, I split `hookEntries` into one single-entry tsdown build per hook with `outputOptions: { codeSplitting: false }`. With that, `head dist/hooks/stop.mjs` now shows the function inlined and there is no shared chunk anywhere under `dist/hooks` or `plugin/scripts`. Build time goes from ~4 s to ~10 s — acceptable for a fix that's specifically about *not* emitting the chunk.

**Polish 2 — stale comments in `src/hooks/stop.ts`.** Removed the two `see sdk-guard.ts` references at lines 3 and 36; they pointed to a deleted file.

**Polish 3 — PR body.** Updated the "Two changes" section to "Three changes" with the corrected description of the helper-and-bundling change. Removed the inverted "they now import it from the shared `src/hooks/sdk-guard.ts`" sentence, since the diff does the opposite.

**Verification:**
- `rm -rf dist plugin/scripts && npm run build` → no `*-<HASH>.mjs` files in `dist/hooks/` or `plugin/scripts/`.
- `head dist/hooks/{stop,session-start,subagent-start}.mjs` → each shows the inlined `isSdkChildContext` body, no external import.
- `npx vitest run test/stop-hook-recursion-guard.test.ts` → 7/7 pass.
- `npx vitest run test/hooks-fire-and-forget.test.ts` → 3/3 pass.
- Full `npm test` shows 11 unrelated failures in `test/mcp-standalone.test.ts`; reproduce on `upstream/main` (verified by running the same suite there), so out of scope.

Plugin artifacts in `plugin/scripts/{session-start,stop,subagent-start}.mjs` are checked-in build outputs (npm publishes from there); regenerated those in the second commit so the published bundles match the source.

bunke:
Closing as superseded by #271 (merged 2026-05-10).

Same direction, same caps (1500 ms inject / 800 ms register), same fire-and-forget telemetry path, same black-hole TCP test pattern — thanks for the quick turnaround. Happy with the inlined-guard choice; the `sdk-guard-internal.ts` split we'd added here was solving a problem that doesn't exist once the helper stays inlined per the established hook pattern.

We'll pick up the fix downstream via the next plugin upgrade and drop our local `BOOKIMED-PATCH 2026-05-01` markers from `claude-plugin/scripts/{session-start,subagent-start}.mjs` at that point.

PR #223, opened 2026-05-02
feat(smart-search): add format parameter (full/narrative/compact) | 添加 format 参数

mechanic-Q:
## Summary | 概述

Add `format` parameter to `mem::smart-search` to align with `mem::search`, which already supports format-aware output (full / compact / narrative).

为 `mem::smart-search` 添加 `format` 参数，与已有的 `mem::search` 格式体系对齐（支持 full / compact / narrative 三种输出模式）。

## Motivation | 动机

When using agentmemory with Hermes Agent and other MCP clients in a real-world workflow, we found that `memory_smart_search` (which calls `mem::smart-search`) is the primary search tool exposed to LLMs. However it only returned compact results — obsId, title, type, score. For the LLM to actually *use* the search results, it needs to see the `narrative` field (the human-readable summary of each observation). 

Without this, every smart-search call required a follow-up expandIds round-trip just to read the narratives — doubling latency and token usage. Meanwhile `memory_search` (mem::search) already had format support. This inconsistency meant integrators had to choose between the smarter hybrid search OR readable results, but couldn't have both.

本补丁来自实际使用 agentmemory + Hermes Agent 的体验：`memory_smart_search` 是 LLM 使用的主要搜索工具，但只返回 compact 格式（obsId/title/type/score），LLM 无法直接读取记忆的 narrative 内容。每次搜索都需要额外 expandIds 调用来获取正文，导致延迟翻倍。而 `memory_search` 已有 format 支持。这个不一致让集成分叉——要么用智能搜索，要么用可读结果，无法兼得。

## Problem | 问题

`mem::search` supports a `format` parameter (`full` | `compact` | `narrative`) that controls the richness of returned results. `mem::smart-search` was hardcoded to compact-only output, creating an inconsistency between the two search functions.

## Changes | 改动

- Added `format?: string` to the input type (defaults to `"compact"` for backwards compatibility)
- Three format modes:
  - `compact` — minimal: obsId, sessionId, title, type, score, timestamp (existing behavior)
  - `narrative` — adds `narrative` field for LLM consumption (replaces the expandIds round-trip)
  - `full` — complete observation object with all fields
- Validates format against allowed values, returns error for invalid input
- Removes unused `CompactSearchResult` import

Default format is `"compact"`, preserving existing behavior. All existing callers work without changes.

## Related | 相关

`mem::search` format implementation: `src/functions/search.ts` L82-84

### DIFF
--- a/src/functions/smart-search.ts
+++ b/src/functions/smart-search.ts
@@ -1,6 +1,5 @@
 import type { ISdk } from "iii-sdk";
 import type {
-  CompactSearchResult,
   CompressedObservation,
   HybridSearchResult,
 } from "../types.js";
@@ -19,6 +18,7 @@ export function registerSmartSearchFunction(
       query?: string;
       expandIds?: Array<string | { obsId: string; sessionId: string }>;
       limit?: number;
+      format?: string;
     }) => {
 
       if (data.expandIds && data.expandIds.length > 0) {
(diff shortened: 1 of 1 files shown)


rohitg00:
Thanks @mechanic-Q — `format` as a query parameter is the right call. Saves a downstream client from having to expand obsIds when they really wanted the narrative inline.

Two small nits before merge:

**1. Bad-format response shouldn't echo the bad string.**

```ts
return { mode: format, results: [], error: "format must be one of 'full', 'compact', or 'narrative'" };
```

If a caller sends `format: "foobar"`, you return `mode: "foobar"`, which makes the response shape lie about itself. Either return `mode: "compact"` (the safe default) plus the error, or drop `mode` entirely on the error path. Same applies in the log line.

**2. `Array<Record<string, unknown>>` is loose.**

The three branches each produce a well-shaped variant. Mind defining a `SearchResultFormatted` discriminated union (or keeping `CompactSearchResult` and adding `NarrativeSearchResult` / `FullSearchResult`)? Otherwise downstream callers lose all type safety on what they get back.

Functionally clean otherwise. Once those are addressed, happy to merge.

PR #240, opened 2026-05-08
feat: add OpenAI-compatible LLM provider

fatinghenji:
## Summary

Adds a new `openai` LLM provider that uses raw fetch to call any OpenAI-compatible `/v1/chat/completions` endpoint.

## Motivation

Currently, AgentMemory only supports Anthropic, Gemini, OpenRouter, MiniMax, and the agent-sdk fallback for LLM-backed compression and summarization. Users with OpenAI API keys (or keys from OpenAI-compatible services like DeepSeek, SiliconFlow, Azure OpenAI, vLLM, LM Studio) cannot use their existing credentials for the LLM layer, even though the embedding layer already supports `OPENAI_API_KEY` via `OpenAIEmbeddingProvider`.

## Changes

- `src/types.ts`: Add `openai` to `ProviderType` union
- `src/providers/openai.ts`: New `OpenAIProvider` class using raw fetch (no SDK dependency)
  - Supports any `/v1/chat/completions` endpoint
  - Respects `OPENAI_API_KEY`, `OPENAI_BASE_URL`, `OPENAI_MODEL` env vars
- `src/providers/index.ts`: Wire `openai` case into `createBaseProvider()`
- `src/config.ts`:
  - Add `OPENAI_API_KEY` detection to `detectProvider()` (with `OPENAI_API_KEY_FOR_LLM` opt-out)
  - Add `OPENAI_API_KEY` to `detectLlmProviderKind()`
  - Add `openai` to `VALID_PROVIDERS` set
  - Update no-key warning message

## Supported Endpoints

| Service | OPENAI_BASE_URL | Notes |
|---------|----------------|-------|
| OpenAI official | https://api.openai.com (default) | |
| DeepSeek | https://api.deepseek.com/v1 | |
| SiliconFlow | https://api.siliconflow.cn/v1 | |
| Azure OpenAI | https://{resource}.openai.azure.com/openai/deployments/{deployment} | |
| vLLM / LM Studio | http://localhost:8000/v1 or http://localhost:1234/v1 | |
| Ollama | http://localhost:11434/v1 | With --enable-openai |

## Configuration Example

```env
# Embedding + LLM both via SiliconFlow
OPENAI_API_KEY=sk-...
OPENAI_BASE_URL=https://api.siliconflow.cn/v1
OPENAI_MODEL=deepseek-chat
OPENAI_EMBEDDING_MODEL=Qwen/Qwen3-Embedding-8B
OPENAI_EMBEDDING_DIMENSIONS=4096
EMBEDDING_PROVIDER=openai
```

## Backwards Compatibility

- Default behavior unchanged: `OPENAI_API_KEY` is now checked first in `detectProvider()`, but only activates when the key is present
- Users who only use `OPENAI_API_KEY` for embedding and prefer another LLM provider can set `OPENAI_API_KEY_FOR_LLM=false` to skip auto-detection
- No breaking changes to existing provider configurations

## Testing

- `npm run build` passes
- `npm test` passes: 838 tests, 75 test files

## Checklist

- [x] Build passes
- [x] Tests pass
- [x] No new dependencies
- [x] Backwards compatible
- [x] Follows existing provider patterns (raw fetch, env var config)

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI LLM provider support with compatibility for OpenAI-compatible APIs, including configurable base URL and model parameters.

* **Bug Fixes**
  * Increased network request timeouts for session management, memory consolidation, and summarization operations to improve reliability.

### DIFF
--- a/plugin/scripts/session-end.mjs
+++ b/plugin/scripts/session-end.mjs
@@ -28,7 +28,7 @@ async function main() {
 			method: "POST",
 			headers: authHeaders(),
 			body: JSON.stringify({ sessionId }),
-			signal: AbortSignal.timeout(5e3)
+			signal: AbortSignal.timeout(3e4)
 		});
 	} catch {}
 	if (process.env["CONSOLIDATION_ENABLED"] === "true") {
@@ -37,7 +37,7 @@ async function main() {
 				method: "POST",
 				headers: authHeaders(),
 				body: JSON.stringify({ olderThanDays: 0 }),
-				signal: AbortSignal.timeout(15e3)
+				signal: AbortSignal.timeout(6e4)
 			});
 		} catch {}
 		try {
@@ -48,7 +48,7 @@ async function main() {
 					tier: "all",
 					force: true
 				}),
-				signal: AbortSignal.timeout(3e4)
+				signal: AbortSignal.timeout(12e4)
 			});
 		} catch {}
 	}
@@ -56,7 +56,7 @@ async function main() {
 		await fetch(`${REST_URL}/agentmemory/claude-bridge/sync`, {
 			method: "POST",
 			headers: authHeaders(),
-			signal: AbortSignal.timeout(5e3)
+			signal: AbortSignal.timeout(3e4)
 		});
 	} catch {}
 }
--- a/plugin/scripts/stop.mjs
+++ b/plugin/scripts/stop.mjs
@@ -28,7 +28,7 @@ async function main() {
 			method: "POST",
 			headers: authHeaders(),
 			body: JSON.stringify({ sessionId }),
-			signal: AbortSignal.timeout(3e4)
+			signal: AbortSignal.timeout(12e4)
 		});
 	} catch {}
 }
--- a/src/config.ts
+++ b/src/config.ts
@@ -50,6 +50,16 @@ function hasRealValue(v: string | undefined): v is string {
 function detectProvider(env: Record<string, string>): ProviderConfig {
   const maxTokens = parseInt(env["MAX_TOKENS"] || "4096", 10);
 
+  // OpenAI-compatible: supports OpenAI, DeepSeek, SiliconFlow, Azure, vLLM, LM Studio
+  if (hasRealValue(env["OPENAI_API_KEY"]) && env["OPENAI_API_KEY_FOR_LLM"] !== "false") {
+    return {
+      provider: "openai",
+      model: env["OPENAI_MODEL"] || "gpt-4o-mini",
+      maxTokens,
+      baseURL: env["OPENAI_BASE_URL"],
+    };
+  }
+
   // MiniMax: Anthropic-compatible API, requires raw fetch to avoid SDK stainless headers
   if (hasRealValue(env["MINIMAX_API_KEY"])) {
     return {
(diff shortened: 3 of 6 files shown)


rohitg00:
Reviewed — the OpenAI-compatible provider itself is clean and well-scoped. Like the raw-fetch approach (avoids SDK stainless headers, works against any `/v1/chat/completions` endpoint including DeepSeek, SiliconFlow, vLLM, LM Studio). The `OPENAI_API_KEY_FOR_LLM=false` opt-out is a thoughtful escape hatch for users who already use `OPENAI_API_KEY` for embeddings only.

Two things before merge:

**1. Scope creep on the timeout bumps.**

The hook script timeout changes (5s→30s, 15s→60s, 30s→120s, 30s→120s) in `plugin/scripts/session-end.mjs` and `plugin/scripts/stop.mjs` are unrelated to the OpenAI provider and don't fit this PR's title. Also they're going **the opposite direction** of #222, which just landed in spirit (5s → 1.5s ceiling on `session-start.ts`) to fix engine starvation under parallel `claude -p` fan-out.

Two options:

- **Preferred** — drop the timeout changes from this PR. If OpenAI-compatible endpoints are slow enough that you needed more headroom, that's a real signal but it deserves its own PR with the use-case spelled out, since #222 just argued the opposite direction for `session-start`.
- If you want to keep them, please call out in the PR body why session-end/stop need 4-5× more time when session-start is moving the other way, and confirm with @bunke that this isn't going to undo the perf fix in #222.

**2. Update README env block + provider table.**

The OpenAI entry is missing from the LLM-provider table in README.md and from the env block. Once those land it's a clean v1 of OpenAI support and #185 closes cleanly with this as the successor.

Otherwise approving the provider code itself. Will hold the merge for @rohitg00 once the scope question is resolved.

flamerged:
I tested the Ollama side of this PR path against current Ollama Cloud/local behavior and I think this PR can close the Ollama use case if it adds one small passthrough.

Findings from 2026-05-12:

- Direct Ollama Cloud exposes OpenAI-compatible discovery: `GET https://ollama.com/v1/models` returns `200` with cloud models including `deepseek-v4-flash`, `kimi-k2.6`, `gpt-oss:120b`, etc.
- Direct native discovery also works: `GET https://ollama.com/api/tags` returns `200`.
- Local signed-in Ollama daemon exposes cloud models through the OpenAI-compatible endpoint at `http://127.0.0.1:11434/v1/chat/completions`.

The important edge case: for a thinking model (`kimi-k2.6:cloud`), this request returned reasoning but empty final content:

```json
{
  "model": "kimi-k2.6:cloud",
  "messages": [{"role":"user","content":"Reply exactly: ok"}],
  "max_tokens": 20,
  "stream": false
}
```

The response had `message.reasoning` populated and `message.content` empty. The same request worked when I added either:

```json
{"reasoning_effort":"none"}
```

or:

```json
{"reasoning":{"effort":"none"}}
```

Native Ollama `/api/chat` also worked with `think:false`.

So for Ollama Cloud via this OpenAI-compatible provider, I think we need one optional env passthrough such as:

```env
OPENAI_REASONING_EFFORT=none
```

and then include `reasoning_effort` in the `/v1/chat/completions` body when set. That would let users configure:

```env
OPENAI_API_KEY=ollama
OPENAI_BASE_URL=http://127.0.0.1:11434/v1
OPENAI_MODEL=kimi-k2.6:cloud
OPENAI_REASONING_EFFORT=none
```

or direct cloud:

```env
OPENAI_API_KEY=<ollama cloud key>
OPENAI_BASE_URL=https://ollama.com/v1
OPENAI_MODEL=kimi-k2.6
OPENAI_REASONING_EFFORT=none
```

Without that, this PR may work for non-thinking models but fail AgentMemory's compression/summarization path on some Ollama Cloud thinking models because the provider treats empty `message.content` as an unexpected response.

fatinghenji:
Closing in favor of a clean rebased PR that addresses all review feedback.

PR #268, opened 2026-05-10
Add Ollama embedding provider

jnibarger01:
## Summary

Adds native Ollama embedding support so agentmemory can use locally downloaded Ollama embedding models without routing through the OpenAI-compatible provider. The default local embedding model is `nomic-embed-text:latest`, with known dimensions for both `nomic-embed-text` and `mxbai-embed-large`.

This also fixes the Windows-specific build script failure by replacing the POSIX shell copy tail with a small Node asset-copy script. While validating the full suite on Windows, three unrelated-looking failure clusters were traced to test portability and isolation assumptions, so this branch fixes those as part of making the suite merge-ready.

## Root Cause And Fix

The embedding provider factory had no first-class Ollama provider, so local Ollama embeddings required pretending the endpoint was OpenAI-compatible. The new provider calls Ollama's native `/api/embed` endpoint, resolves dimensions from the configured model or explicit override, and stays behind the existing dimension guard.

The full test suite failures were not caused by the Ollama provider. `compress-file` and `obsidian-export` used POSIX path expectations in tests while production code resolves paths through Node on Windows. Those tests now compare resolved and joined paths instead of hard-coded separators. `mcp-standalone` unit tests could accidentally proxy to a live server on localhost, so they now force local fallback while the dedicated proxy test file continues to cover proxy behavior.

The package build script used `cp`, `mkdir -p`, POSIX redirection, and `true`, which fails under Windows shells after `tsdown` completes. The build now runs `tsdown && node scripts/copy-build-assets.mjs`.

## Validation

- `npm test -- test/compress-file.test.ts` passed, 7 tests.
- `npm test -- test/obsidian-export.test.ts` passed, 11 tests.
- `npm test -- test/mcp-standalone.test.ts` passed, 24 tests.
- `npm test -- test/mcp-standalone-proxy.test.ts` passed, 7 tests.
- `npm test -- test/embedding-provider.test.ts` passed, 25 tests.
- `npm run test:integration` passed, 17 tests.
- `npm run build` passed on Windows.
- `npm test` passed, 78 files and 862 tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -885,6 +885,10 @@ Create `~/.agentmemory/.env`:
 
 # Embedding provider (auto-detected, or override)
 # EMBEDDING_PROVIDER=local
+# EMBEDDING_PROVIDER=ollama
+# OLLAMA_BASE_URL=http://localhost:11434
+# OLLAMA_EMBEDDING_MODEL=nomic-embed-text:latest
+# OLLAMA_EMBEDDING_DIMENSIONS=768          # nomic-embed-text=768, mxbai-embed-large=1024
 # VOYAGE_API_KEY=...
 # OPENAI_API_KEY=sk-...
 # OPENAI_BASE_URL=https://api.openai.com   # Override for Azure / vLLM / LM Studio / proxies
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "agentmemory": "dist/cli.mjs"
   },
   "scripts": {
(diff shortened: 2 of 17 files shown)


jnibarger01:
Closing this draft rather than keeping a dirty/conflicting PR open. Current state: draft, merge conflicts, and Vercel failing; if Ollama embedding support is still wanted, it should be revived as a fresh branch rebased on current main with CI/Vercel sorted.

PR #271, opened 2026-05-10
fix: cap session-start/subagent-start hook latency (#221)

### LINKED ISSUE #221: session-start / subagent-start hooks block agent startup for up to 5s on slow/unreachable REST

rohitg00:
## Summary
- `session-start.ts` awaited a 5000ms POST and discarded the response whenever `AGENTMEMORY_INJECT_CONTEXT=false` (the default since 0.8.10) — pure blocking latency on every Claude Code session start.
- `subagent-start.ts` had a `// fire and forget` comment but the code awaited a 2000ms POST.
- Under concurrent fan-out (Slack-bot orchestrators, multi-agent harnesses, fanned `claude -p` jobs) the awaited timeouts stack and create the positive feedback loop the reporter described — leading to iii-engine OOM-kill.

## Fix
- `session-start` — fire-and-forget on the telemetry path. Cap the inject path at **1500ms** (down from 5000ms) so a slow server can't block the agent indefinitely when stdout is consumed.
- `subagent-start` — actually fire-and-forget, matching the existing comment. Cap at **800ms**.

## Live e2e (black-hole TCP listener: accepts, never replies)
| hook | before | after |
|---|---|---|
| `session-start` (no inject) | 5.05s | **0.85s** |
| `session-start` (inject=true) | 5.05s | **1.55s** |
| `subagent-start` | 2.05s | **0.87s** |

## Notes
- Bundled artifacts in `plugin/scripts/` regenerated via `npx tsdown`.
- The shared `isSdkChildContext` guard is intentionally inlined in each hook source (matches the established pattern: every `tsdown` hook entry compiles to a single self-contained `.mjs`).
- Full suite: 856 passing.

Closes #221.

## Summary by CodeRabbit

* **Performance**
  * Optimized timeout handling for startup operations with separate thresholds for telemetry versus response-dependent requests.
  * Improved startup responsiveness by making telemetry requests non-blocking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/271)

### DIFF
--- a/plugin/scripts/session-start.mjs
+++ b/plugin/scripts/session-start.mjs
@@ -8,6 +8,8 @@ function isSdkChildContext(payload) {
 const INJECT_CONTEXT = process.env["AGENTMEMORY_INJECT_CONTEXT"] === "true";
 const REST_URL = process.env["AGENTMEMORY_URL"] || "http://localhost:3111";
 const SECRET = process.env["AGENTMEMORY_SECRET"] || "";
+const INJECT_TIMEOUT_MS = 1500;
+const REGISTER_TIMEOUT_MS = 800;
 function authHeaders() {
 	const h = { "Content-Type": "application/json" };
 	if (SECRET) h["Authorization"] = `Bearer ${SECRET}`;
@@ -25,18 +27,29 @@ async function main() {
 	if (isSdkChildContext(data)) return;
 	const sessionId = data.session_id || `ses_${Date.now().toString(36)}`;
 	const project = data.cwd || process.cwd();
+	const url = `${REST_URL}/agentmemory/session/start`;
+	const init = {
+		method: "POST",
+		headers: authHeaders(),
+		body: JSON.stringify({
+			sessionId,
+			project,
+			cwd: project
+		})
+	};
+	if (!INJECT_CONTEXT) {
+		fetch(url, {
+			...init,
+			signal: AbortSignal.timeout(REGISTER_TIMEOUT_MS)
+		}).catch(() => {});
+		return;
+	}
 	try {
-		const res = await fetch(`${REST_URL}/agentmemory/session/start`, {
-			method: "POST",
-			headers: authHeaders(),
-			body: JSON.stringify({
-				sessionId,
-				project,
-				cwd: project
-			}),
-			signal: AbortSignal.timeout(5e3)
+		const res = await fetch(url, {
+			...init,
+			signal: AbortSignal.timeout(INJECT_TIMEOUT_MS)
 		});
-		if (INJECT_CONTEXT && res.ok) {
+		if (res.ok) {
 			const result = await res.json();
 			if (result.context) process.stdout.write(result.context);
 		}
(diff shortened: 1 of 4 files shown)


bunke:
Confirming this works in production after upgrading our plugin install to v0.9.9. Verified locally against a black-hole TCP listener (accepts, never replies):

```
session-start (INJECT_CONTEXT=false):  5.073s → 0.874s
subagent-start:                        2.075s → 0.874s
```

Matches the PR description. Thanks for the rapid turnaround — this closes the OOM-loop we hit on 2026-05-01 (originally filed as #221).

PR #273, opened 2026-05-10
feat: opt-in fail-loud when AGENTMEMORY_URL is unreachable

TotalZack:
## Problem

`@agentmemory/mcp`'s standalone shim probes `/agentmemory/livez` and **silently falls back to local in-memory mode** on probe failure. Users on multi-machine self-hosted setups (Tailscale, Coolify) have no signal that their memories are landing in a local store rather than the configured shared backend.

## Fix

Two new env vars, both default-off (no behavior change for existing users), plus a visible stderr line on the previously-silent fallback path:

| Env var | Default | Effect |
|---|---|---|
| `AGENTMEMORY_REMOTE_REQUIRED` | unset | When `1` / `true` / `yes` / `on`: probe failure throws a structured `RemoteUnreachableError` instead of falling back to local. Recommended for production multi-machine deployments. |
| `AGENTMEMORY_LIVEZ_TIMEOUT_MS` | 2000 ms (same default as `AGENTMEMORY_PROBE_TIMEOUT_MS` since v0.9.7) | Override the probe timeout. Takes precedence over `AGENTMEMORY_PROBE_TIMEOUT_MS` so existing deployments that already set this env var continue to work unchanged. |

Plus: when `REMOTE_REQUIRED` is unset and probe fails, the shim now writes a single line to `stderr` — with `url`, `probeTimeoutMs`, and a hint at both env vars.

## Three operator-friendly modes

| Setup | Configuration |
|---|---|
| Indie dev (default, unchanged) | env unset → try remote, fall back to local with a visible stderr line if probe fails |
| Strict production multi-machine | `AGENTMEMORY_REMOTE_REQUIRED=1` → throw `RemoteUnreachableError` if probe fails |
| Local-only | `AGENTMEMORY_URL` unset or unreachable, no `REMOTE_REQUIRED` → local-only (existing behavior) |

## Changes vs v0.9.9

- `src/mcp/rest-proxy.ts`: `remoteRequired()` helper + `RemoteUnreachableError` class + merged `probeTimeoutMs()` that checks `AGENTMEMORY_LIVEZ_TIMEOUT_MS` first, then `AGENTMEMORY_PROBE_TIMEOUT_MS`; `resolveHandle()` throws on probe failure when `REMOTE_REQUIRED` is set; fallback path emits `process.stderr.write` (matching upstream's v0.9.7 style) instead of `logger.warn`
- `test/mcp-standalone-proxy.test.ts`: 3 vitest cases covering each operator mode

## Compatibility

- No behavior change for users with no env vars set
- `AGENTMEMORY_LIVEZ_TIMEOUT_MS` is a transparent alias — deployments using the old name keep working
- Rebased onto v0.9.9

## Summary by CodeRabbit

* **Documentation**
  * Updated changelog and README with guidance and example env settings for remote-backend resilience.

* **New Features**
  * Added AGENTMEMORY_LIVEZ_TIMEOUT_MS to customize remote health-probe timeout.
  * Added AGENTMEMORY_REMOTE_REQUIRED to opt into fail‑loud behavior instead of silent local fallback.
  * Enhanced stderr warning when remote /livez is unreachable (includes timeout and env var hints).

* **Tests**
  * Added tests covering timeout, fallback, and fail‑loud behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/273)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Added (TotalZack/agentmemory fork — B1 patch, pending upstream PR #273)
+
+- **`AGENTMEMORY_REMOTE_REQUIRED` env var** — when set to `1` / `true` / `yes` / `on`, the standalone MCP shim refuses to fall back to in-memory local mode if the `/agentmemory/livez` probe fails at startup. Throws a structured `RemoteUnreachableError` instead. Default off, so existing users see no behavior change. Recommended for production multi-machine deployments.
+- **`AGENTMEMORY_LIVEZ_TIMEOUT_MS` env var** — overrides the probe timeout (alias for upstream's `AGENTMEMORY_PROBE_TIMEOUT_MS`; takes precedence). Default 2000 ms (same as upstream since v0.9.7). Useful on high-latency networks (Tailscale across regions, slow Coolify boots) where the default trips the silent-fallback path even when the backend is healthy.
+- **Visible stderr line on silent local fallback** — when `AGENTMEMORY_URL` points at a remote backend but the `/livez` probe fails and `REMOTE_REQUIRED` is unset, the shim now emits a warning on stderr (with `url`, `probeTimeoutMs`, and a hint at the two env vars).
+
 ## [0.9.9] — 2026-05-11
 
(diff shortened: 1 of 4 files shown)


TotalZack:
Just to clarify the **Real-world validation** section in the PR body — the Hetzner / Coolify / Tailscale specifics there describe one concrete dev environment that was used to reproduce both the silent-fallback bug and verify the fix end-to-end. The bug + fix apply to **any** self-hosted setup where the agentmemory backend sits behind a private overlay network (Tailscale, WireGuard, ZeroTier, in-cluster service mesh, etc.), where 500 ms is sometimes not enough probe headroom. The fix is platform-neutral.

Folgecommit `882037e` addresses the three CodeRabbit nitpicks (full env var name in CHANGELOG, fallback wording in README, neutral wording for `RemoteUnreachableError.message`). Pushed as a separate commit so the original B1 work and the review-feedback fixes stay readable in history.

PR #314, opened 2026-05-12
fix(cli): add --data-dir flag + AGENTMEMORY_DATA_DIR so engine state lives outside repos

### LINKED ISSUE #303: Engine writes data/ to caller's cwd — pollutes user repos when started from project root
## Summary

`npx @agentmemory/agentmemory` starts `iii-engine` with the *current working directory* as the state-store location. When the server is launched from inside a project (which is the default for the Claude Code plugin auto-start path, since Claude Code's cwd is the user's repo), the engine writes a `data/` directory directly into the user's git working tree:

```
data/state_store.db/mem%3Aaccess.bin
data/state_store.db/mem%3Ahealth.bin
data/state_store.db/mem%3Amemories.bin
data/state_store.db/mem%3Aobs%3A<session-id>.bin
data/state_store.db/mem%3Aindex%3Abm25.bin
data/stream_store/stream%3Amem-live%3A<session-id>.bin
data/stream_store/stream%3Amem-live%3Aviewer.bin
```

These show up immediately in `git status` and are easy to accidentally commit. They also leak session ids and memory contents into whatever repo the user happened to be in when the server first started.

## Repro

```bash
cd ~/some-project    # any git repo
npx -y @agentmemory/agentmemory   # let it become healthy
# in another shell:
cd ~/some-project && git status
# => Untracked: data/
```

## Environment

- macOS 14 (darwin 23.6.0), x86_64
- agentmemory `0.9.9`
- iii-engine `0.11.2` (the version agentmemory pins)
- Node from `npx`, no Docker fallback

## Workaround

Start the server from a dedicated directory:

```bash
mkdir -p ~/.agentmemory
cd ~/.agentmemory && npx -y @agentmemory/agentmemory
```

This works, but it's fragile because:
1. The Claude Code plugin install instructions (\"run `npx @agentmemory/agentmemory` in a separate terminal\") don't mention cwd, so a typical user who pastes the install block into their project terminal will hit this.
2. There's no CLI flag or env var to override the data location (`--help` lists `--port`, `--no-engine`, `--tools`, `--verbose`, but nothing for state).
3. If/when any hook or skill auto-spawns the engine, it will inherit Claude Code's cwd (the repo) and recreate the problem.

## Proposed fix

Make the state-store location explicit and default it outside any repo. A couple of equivalent shapes:

1. **CLI flag + env var**, defaulting to `~/.agentmemory/data` (or platform-appropriate: `$XDG_DATA_HOME/agentmemory` on Linux, `~/Library/Application Support/agentmemory` on macOS):
   ```bash
   agentmemory --data-dir ~/.agentmemory/data
   AGENTMEMORY_DATA_DIR=~/.agentmemory/data npx @agentmemory/agentmemory
   ```
2. **Auto-relocate** when started inside a git repo: detect `.git` in cwd or any ancestor, and if found, use `~/.agentmemory/data` instead of `./data`. Log a one-line warning so users aren't surprised.

(2) alone would catch every accidental case without needing users to set anything; (1) is still worth having for users who want to pin the location.

Happy to send a PR if you'd like — let me know which shape you prefer.

## Why this matters

Plugin auto-installation is the headline path in the README (\"`/plugin marketplace add rohitg00/agentmemory` and `/plugin install agentmemory`\"). Anyone following that path from inside their repo gets engine state written into their working tree on the first run. The current default basically guarantees a footgun for the recommended install flow.

mvanhorn:
## Summary

Adds a `--data-dir` flag and `AGENTMEMORY_DATA_DIR` env var so the engine writes its state outside the caller's cwd. Defaults to a platform-appropriate user dir, so the documented `npx @agentmemory/agentmemory` flow no longer drops a `data/` tree into whatever repo you happen to be in.

## What changed

- `src/cli-data-dir.ts` — new resolver: flag > env > platform default. `~` expansion, `${VAR}` interpolation, `.` and `./data` honored as explicit opt-ins for repo-local state.
- `src/cli.ts` — `--data-dir <path>` parsed alongside `--port`; writes a runtime `iii-config.yaml` into the resolved dir so `state_store.db` and `stream_store` live there too. `--help` documents the flag.
- `docker-compose.yml` — fallback mount becomes `${AGENTMEMORY_DATA_DIR:-iii-data}` to `/data`.
- `README.md` — short section on the new flag and env var.
- `test/cli-data-dir.test.ts` — covers flag-wins, env-fallback, default-not-cwd, repo-local opt-in, `~` expansion.

Default by platform:
- macOS: `~/Library/Application Support/agentmemory`
- Linux: `$XDG_DATA_HOME/agentmemory` (falls back to `~/.local/share/agentmemory`)
- Windows: `~/AppData/Local/agentmemory`

## Testing

`npm test` and `npm run build` couldn't run locally because `vitest` / `tsdown` aren't installed in this sandbox (`npm install` blocked by network). The new tests are written against the existing vitest shape used elsewhere in `test/`; CI / CodeRabbit will exercise them.

Fixes #303

Signed-off-by: Matt Van Horn <mvanhorn@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Added persistent data directory configuration via `--data-dir` or `AGENTMEMORY_DATA_DIR`, with OS-specific defaults and `--instance`-scoped storage.
  * The CLI now generates a runtime `iii-config.yaml` so databases and stream files live under the resolved directory (with a one-time relocation notice when applicable).
* **Deployment**
  * Docker Compose mounts the engine `/data` path using the `AGENTMEMORY_DATA_DIR`-based location instead of a fixed named volume.
* **Documentation**
  * README updated with default locations and override instructions; the environment reference now includes `AGENTMEMORY_DATA_DIR`.
* **Bug Fixes**
  * `import-jsonl` now treats `--data-dir` as a value-taking option.
* **Tests**
  * Added unit tests covering precedence, defaults, and legacy/relocation behavior.


HaleTom:
Please be aware of #871 which is about moving the `bin` directory to the XDGBDS directory: `$HOME/.local/bin`.

rohitg00:
hey, please rebase. thanks

HaleTom:
Would you consider moving the default install directory to $XDG_DATA_HOME/agentmemory also?

Otherwise there would be two directories (`~/.agentmemory` would be the other, which doesn't conform to the XDGBDS).

mvanhorn:
Thanks @HaleTom - good news, the PR already does this. On Linux the default resolves to `$XDG_DATA_HOME/agentmemory` when set, falling back to `~/.local/share/agentmemory` (the XDG spec default) - see `platformDefaultDataDir` in `src/cli-data-dir.ts`. macOS uses `~/Library/Application Support/agentmemory` (platform-native), and Windows uses `%APPDATA%/agentmemory`, with `~/.agentmemory` only as a last-resort when APPDATA is unset. So no `~/.agentmemory` on Linux/macOS. (The `~/.agentmemory/.env` you may be thinking of is the separate hermes integration's config path, not this data dir.) Happy to switch macOS to XDG too if you'd prefer that over Application Support. Thanks also for the #871 pointer.

rohitg00:
The core fix is right. Confirmed on current main: the engine spawns with the caller's cwd and the bundled `iii-config.yaml` uses relative `./data/state_store.db`, so `npx @agentmemory/agentmemory` drops a `data/` tree into whatever repo you run it in. A `--data-dir` / `AGENTMEMORY_DATA_DIR` with a platform default is the right shape for #303.

Two things before it can land:

1. Migration / back-compat. Changing the default from cwd `./data` to a platform user dir orphans existing users' state: they boot into a fresh empty store and their memories look gone. Please add an adopt-legacy step, if a `./data` (or the prior default) already exists, use or migrate it, and fall back to the new platform default only for fresh installs.

2. Rebase. The branch is conflicting and predates the current CLI. Config is now resolved through `findIiiConfig()` (env > cwd > ~/.agentmemory > bundled), and there is a `--instance` multi-daemon mode plus a private engine install under `~/.agentmemory/bin`. The resolver needs to reconcile with `findIiiConfig()`, and `--data-dir` should compose with `--instance` so two daemons never share one state store.

Supply chain and scope look clean (no dependency, postinstall, or workflow changes). Solid DX win once the migration path is in.

mvanhorn:
Pushed 8242fab with the adopt-legacy step: if a ./data store from the prior default exists, it keeps being used before the platform-default kicks in, so nobody's memories vanish on upgrade. Added tests covering both the changed-file and no-op paths. Build and vitest pass locally (5/5).

rohitg00:
@mvanhorn conflicts are still there.

mvanhorn:
Rebased onto latest main and resolved the conflicts. The data-dir handling is re-ported onto the refactored engine-start flow: the runtime iii config (pointing at the relocated data dir) now feeds startIiiBin and the pinned-version path, and my old fallback loop is dropped in favor of main's pickCompatibleIii. Build and tests pass locally. Should be clean to review now.

mvanhorn:
Regenerated the skills reference (`npm run skills:gen`) so `AGENTMEMORY_DATA_DIR` shows up — the stale autogen block in `REFERENCE.md` was what the check was failing on. Focused data-dir tests pass 5/5; the remaining suite failures are the sandbox loopback/home-write ones that also reproduce on `main`.

mvanhorn:
Fixed the broken anchor — the heading slug is `#local-models-ollama--lm-studio--vllm` (the removed slashes leave double hyphens), so the link resolves now.

rohitg00 (changes requested):
Thanks for turning this around, @mvanhorn. The adopt-legacy commit is exactly the migration path I was worried about last time, and the rebase cleared the conflicts. Core idea is right and I want it in. Three things holding me back, all about what happens on real setups:

**1. The legacy `./data` pickup is too eager.** `resolveDataDir` adopts any `data/` folder in the cwd (`existsSync(resolve(cwd, "data"))`), but loads of repos already have a `data/` dir that has nothing to do with us (ML/data projects especially). Someone runs `npx @agentmemory/agentmemory` in one of those and we quietly start writing `state_store.db` / `stream_store` / `iii-config.yaml` into their folder. That's worse than the bug we're fixing. Can we only adopt it when it's actually ours, e.g. `data/state_store.db` (or `data/iii-config.yaml`) exists?

**2. Docker users lose their volume.** cli.ts sets `process.env.AGENTMEMORY_DATA_DIR` to the resolved absolute path unconditionally, even for the default. So in docker-compose, `${AGENTMEMORY_DATA_DIR:-iii-data}` never falls back to the `iii-data` named volume, it bind-mounts the platform dir instead. Anyone already on the docker path suddenly boots against an empty dir with their old memories stranded in the named volume. I think we only want to export `AGENTMEMORY_DATA_DIR` when the user actually passed `--data-dir` / set the env, so the default keeps the volume.

**3. The XDG relocation is too broad.** It fires whenever cwd is inside any git repo and `XDG_DATA_HOME` is set, without checking XDG actually points inside the worktree. So `XDG_DATA_HOME=/mnt/data` run from a normal repo gets XDG ignored plus a warning that says it's "inside a git worktree" when it isn't. Can we scope it to when the XDG path is really under the git root?

Two smaller ones, fine as follow-ups: `--data-dir` doesn't move the image store (`IMAGES_DIR` is still hardcoded to `~/.agentmemory/images`, nothing on the worker reads `AGENTMEMORY_DATA_DIR`), and `renderIiiConfig` rewrites the two `file_path` lines by exact string match, so a custom `~/.agentmemory/iii-config.yaml` would silently keep writing to `./data`.

Requesting changes on 1-3, but this is close and I appreciate you sticking with it.

mvanhorn:
All three addressed in 9f5f449.

1. Legacy `./data` pickup is now keyed on the directory actually being ours: `resolveDataDir` adopts it only when `data/state_store.db` or `data/iii-config.yaml` exists, rather than on `data/` existing at all. Your ML-repo example was the right one to worry about.

2. `cli.ts` only exports `AGENTMEMORY_DATA_DIR` when the resolution source is `flag` or `env`, so the default no longer shadows `${AGENTMEMORY_DATA_DIR:-iii-data}` in docker-compose and existing users keep their named volume.

3. The XDG relocation now requires the XDG path to actually be under the git root, via a `relative()`-based containment check rather than just "cwd is inside a repo and XDG_DATA_HOME is set". `XDG_DATA_HOME=/mnt/data` from a normal repo is left alone, with no misleading worktree warning.

I left the two you called follow-ups alone: `IMAGES_DIR` still does not move with `--data-dir`, and `renderIiiConfig` still rewrites the `file_path` lines by exact string match. Happy to take either in a separate PR.

Test run: 1417 passed, 5 failed. The 5 are in `test/hook-project.test.ts` and fail identically on clean HEAD; they assert `resolveProject` returns `"agentmemory"`, which only holds when the checkout directory is named that. Mine is not, so they are an artifact of where I cloned it. The two new `cli-data-dir` cases pass.

rohitg00:
Thanks @mvanhorn for your contribution!

rohitg00 (approved):

mvanhorn:
Thank you for merging the --data-dir flag. Appreciate you taking a change that keeps engine state out of the caller's cwd.

PR #315, opened 2026-05-12
fix(integrations): warn when bearer auth crosses plaintext HTTP to non-loopback

### LINKED ISSUE #275: CRITICAL: Plugin sends auth token over plaintext HTTP
## Summary
The OpenCode plugin () defaults  to  (plaintext HTTP). The auth secret is sent as a Bearer token on every request. If the environment variable is not set, all traffic — including potentially sensitive data from observations, memories, and LLM prompts — travels unencrypted.

## Impact
- Auth tokens and session data exposed on the wire
- Man-in-the-middle attacks possible on shared networks

## Fix
Default to  or enforce HTTPS when a custom URL is not explicitly configured.

mvanhorn:
## Summary

Each of the three plugin clients now warns once when a bearer token would be sent over plaintext HTTP to a non-loopback host. Loopback HTTP stays silent (it's not actually wire-exposed). HTTPS stays silent. An opt-in `AGENTMEMORY_REQUIRE_HTTPS=1` env var turns the warning into a startup error for users who want a guarantee.

## What changed

- `integrations/openclaw/plugin.mjs`: HTTP + non-loopback + secret -> one-time `console.warn`. Hard-fail when `AGENTMEMORY_REQUIRE_HTTPS=1`.
- `integrations/pi/index.ts` + new `integrations/pi/security.ts`: same guard, factored into a small module so the test can target it directly.
- `integrations/hermes/__init__.py`: same guard using `warnings.warn` + `RuntimeError` for the hard-fail path.
- `test/integration-plaintext-http.test.ts`: covers loopback silence, remote HTTP warning fires once, HTTPS silence, hard-fail when env is set.

Default behavior is unchanged. The warning fires only when the configuration is actually dangerous (HTTP + non-loopback + secret).

## Why not raise by default

`http://localhost:3111` is the documented default and is loopback-only — the kernel routes those packets without touching any NIC. Hard-failing every plaintext-localhost user would break the install flow for no security gain.

## Testing

- `node --check` on the modified `.mjs` file
- `python3 -m py_compile` on the modified Python plugin
- Targeted smoke checks on each plugin's auth path
- `npm test` couldn't run locally because `vitest` isn't in the sandbox; the new test follows the existing `test/*.test.ts` shape, CI/CodeRabbit will exercise it

Fixes #275

Signed-off-by: Matt Van Horn <mvanhorn@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Plaintext HTTP Bearer-token safety checks added across integrations: loopback hosts bypass, remote HTTP emits a one-time warning to stderr or raises when AGENTMEMORY_REQUIRE_HTTPS=1; enforcement runs at initialization and before requests.

* **Tests**
  * Added integration tests covering loopback, remote, HTTPS, warn-once, and fail-on-require semantics across integrations.

* **Documentation**
  * READMEs updated to document AGENTMEMORY_REQUIRE_HTTPS behavior.

* **Chore**
  * .gitignore updated to ignore Python bytecode cache.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/315)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ data/
 package-lock.json
 pnpm-lock.yaml
 yarn.lock
+integrations/hermes/__pycache__/
--- a/integrations/hermes/README.md
+++ b/integrations/hermes/README.md
@@ -105,6 +105,7 @@ The plugin auto-detects the running server and hooks into the Hermes agent loop.
 |---|---|---|
 | `AGENTMEMORY_URL` | `http://localhost:3111` | agentmemory server URL |
 | `AGENTMEMORY_SECRET` | (none) | Auth token for protected instances |
+| `AGENTMEMORY_REQUIRE_HTTPS` | (off) | When set to `1`, refuse to send the bearer token over plaintext HTTP to a non-loopback host. Sends only when `AGENTMEMORY_URL` is `https://...` or points at `localhost`/`127.0.0.1`/`::1`. With this off, the plugin warns once on stderr but still sends. |
 
 The plugin reads `~/.agentmemory/.env` (or `$XDG_CONFIG_HOME/agentmemory/.env`) at import time and populates any missing values into the process environment via `os.environ.setdefault`. Anything you set in the shell takes precedence; the file is only used to fill gaps. This means `hermes memory status` reports the plugin as available even when the agentmemory service is launched by systemd or another process manager that loads `~/.agentmemory/.env` directly without exporting it to the Hermes CLI shell (#250).
 
--- a/integrations/hermes/__init__.py
+++ b/integrations/hermes/__init__.py
@@ -11,10 +11,12 @@ from __future__ import annotations
 
 import json
 import os
+import sys
 import threading
 import time
 from pathlib import Path
-from typing import Any
+from typing import Any, Callable
+from urllib.parse import urlparse
 from urllib.request import Request, urlopen
 from urllib.error import URLError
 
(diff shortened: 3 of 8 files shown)


rohitg00:
Shipped in v0.9.12 ([release notes](https://github.com/rohitg00/agentmemory/releases/tag/v0.9.12)). Thanks @mvanhorn — symmetric three-runtime guard (Python / mjs / TypeScript), warn-once semantic, and the AGENTMEMORY_REQUIRE_HTTPS=1 escalation knob is exactly the right shape for hardened deployments. Hardening commits we layered before merge (IPv6 / LAN-IP / lookalike-hostname edge-case tests + env-var docs in pi/hermes README) preserved every line of your design — the guard logic itself didn't need a change.

mvanhorn:
Thanks @rohitg00 - really appreciate the merge and the extra hardening pass. The IPv6 / LAN-IP / lookalike-hostname coverage tightens the policy in exactly the gaps a single-loopback-string check would miss; good additions. Glad the env-var escape hatch landed cleanly for the hardened-deployment path.

rohitg00:
Thanks for your contribution and whatever you do for community, I use /last30days daily.

mvanhorn:
Thanks @rohitg00, really glad /last30days is in your rotation. The bearer-over-HTTP warn was a fun one to track down; reach out anytime if there's something else useful to add on the integrations side.

mvanhorn:
@rohitg00 thanks for the merge. Warning when a bearer token would hit non-loopback plaintext HTTP is exactly the kind of polite hardening that prevents accidental exposure without breaking the dev-loop on localhost.

PR #380, opened 2026-05-14
fix(cli): authenticate apiFetch when AGENTMEMORY_SECRET is set

eleboucher:
The doctor and status commands call protected endpoints (config/flags, health) via apiFetch but never send the bearer header. Deployments that set AGENTMEMORY_SECRET get 401 responses, parsed as flags=null,  which the doctor renders as 'LLM provider: not set' and 'Embedding provider: not set' regardless of actual server-side config.

## Summary by CodeRabbit

* **New Features**
  * API requests now support optional authentication when credentials are configured.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/380)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -473,7 +473,13 @@ async function main() {
 
 async function apiFetch<T = unknown>(base: string, path: string, timeoutMs = 5000): Promise<T | null> {
   try {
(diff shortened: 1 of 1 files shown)


eleboucher:
@rohitg00 can you give a look please?

PR #386, opened 2026-05-15
fix: pass plugin mcp env overrides

### LINKED ISSUE #375: Feature: Plugin should inherit MCP server env vars from environment
## Problem

The Claude Code plugin hardcodes the MCP server config to `localhost:3111`. When agentmemory is deployed remotely (e.g., Kubernetes cluster accessed via reverse proxy, or cluster-internal DNS), users must add a duplicate entry in `.mcp.json` with the correct `AGENTMEMORY_URL` and `AGENTMEMORY_SECRET` env vars.

This causes a permanent `/doctor` warning:
```
MCP server "agentmemory" skipped — same command/URL as already-configured "agentmemory"
```

Both entries are needed because:
- **Plugin**: provides hooks (observation capture, session lifecycle)
- **`.mcp.json`**: provides MCP server with correct URL/auth env vars

## Requested Behavior

Plugin should support env var overrides for its MCP server config, either:
1. Read `AGENTMEMORY_URL` and `AGENTMEMORY_SECRET` from the environment when available instead of hardcoding localhost
2. Detect an existing `.mcp.json` entry and skip its own MCP server definition (without the warning)
3. Allow plugin config to specify env var overrides

## Use Cases

- Kubernetes deployment with cluster-internal DNS (`http://agentmemory.agentmemory.svc.cluster.local:3111`)
- Remote server behind reverse proxy with API key auth
- Any non-localhost deployment where hooks are still needed

LaplaceYoung:
## Summary
- Pass `AGENTMEMORY_URL` and `AGENTMEMORY_SECRET` through the bundled plugin MCP server config.
- Keep `http://localhost:3111` as the default MCP server target when `AGENTMEMORY_URL` is unset.
- Add regression coverage for the plugin MCP env block.
- Document the remote/protected Claude Code launch path.

Fixes #375

## Test plan
- `npm test -- --run test/codex-plugin.test.ts`
- `npx tsdown`
- `git diff --check`

## Notes
- `npm run build` reaches the tsdown compile step on Windows, then the shell copy tail exits on `true` because the script uses Unix shell syntax.
- `npm test` on Windows reports path-sensitive failures in `test/obsidian-export.test.ts` and `test/compress-file.test.ts`; 894 tests pass, including `test/codex-plugin.test.ts`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified installation snippet on how to run agentmemory in remote/protected deployments by setting AGENTMEMORY_URL and AGENTMEMORY_SECRET; notes these values are forwarded to the bundled MCP server (default URL shown).
* **Tests**
  * Added an automated test to verify the environment variable entries for agentmemory are present and correctly passed through.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/386?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,6 +355,8 @@ Implementation details live in `src/cli.ts` (see `runUpgrade` around the `src/cl
 Install agentmemory: run `npx @agentmemory/agentmemory` in a separate terminal to start the memory server. Then run `/plugin marketplace add rohitg00/agentmemory` and `/plugin install agentmemory` — the plugin registers all 12 hooks, 4 skills, AND auto-wires the `@agentmemory/mcp` stdio server via its `.mcp.json`, so you get 51 MCP tools (memory_smart_search, memory_save, memory_sessions, memory_governance_delete, etc.) without any extra config step. Verify with `curl http://localhost:3111/agentmemory/health`. The real-time viewer is at http://localhost:3113.
 ```
 
+For remote or protected deployments, launch Claude Code with `AGENTMEMORY_URL` and `AGENTMEMORY_SECRET` set. The plugin passes both values through to its bundled MCP server, with `http://localhost:3111` as the default URL.
+
 ### Codex CLI (Codex plugin platform)
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ Implementation details live in `src/cli.ts` (see `runUpgrade` around the `src/cl
 Install agentmemory: run `npx @agentmemory/agentmemory` in a separate terminal to start the memory server. Then run `/plugin marketplace add rohitg00/agentmemory` and `/plugin install agentmemory` — the plugin registers all 12 hooks, 4 skills, AND auto-wires the `@agentmemory/mcp` stdio server via its `.mcp.json`, so you get 51 MCP tools (memory_smart_search, memory_save, memory_sessions, memory_governance_delete, etc.) without any extra config step. Verify with `curl http://localhost:3111/agentmemory/health`. The real-time viewer is at http://localhost:3113.
 ```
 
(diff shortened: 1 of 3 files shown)


rohitg00:
Solid fix and the README addition is great. One small concern before merge: the `${VAR:-default}` bash-style fallback syntax — Claude Code's MCP loader expands `${VAR}` (verified in the wild) but I'm not confident it strips the `:-default` suffix. If a user on Linux without `AGENTMEMORY_URL` set in their environment gets the literal string `${AGENTMEMORY_URL:-http://localhost:3111}` passed to the MCP shim's `process.env["AGENTMEMORY_URL"]`, the shim's own fallback (https://github.com/rohitg00/agentmemory/blob/main/src/mcp/standalone.ts#L44) catches it because the unknown URL hostname won't resolve — but it's an awkward failure path.

Could you simplify to:

```json
"env": {
  "AGENTMEMORY_URL": "${AGENTMEMORY_URL}",
  "AGENTMEMORY_SECRET": "${AGENTMEMORY_SECRET}"
}
```

The shim already defaults to `http://localhost:3111` when `AGENTMEMORY_URL` is empty (https://github.com/rohitg00/agentmemory/blob/main/src/mcp/standalone.ts#L44), so we don't need the bash-style fallback. Same goes for SECRET — the shim treats empty as 'no auth'. Update the test assertion accordingly. Happy to merge once that's in.

LaplaceYoung:
Updated the plugin MCP env templates to use plain `${AGENTMEMORY_URL}` and `${AGENTMEMORY_SECRET}`, with the localhost fallback handled in the shim. I also adjusted the related test and README note.

Verified locally with:

```sh
npm test -- --run test/codex-plugin.test.ts
```

rohitg00:
Thanks @LaplaceYoung — plugin .mcp.json was indeed missing the env-passthrough that the connect adapters already had (util.ts AGENTMEMORY_MCP_BLOCK). Cleanly closes the remote-deployment gap. Merged. Closes #375.

PR #412, opened 2026-05-15
fix: support AGENTMEMORY_EMBEDDING_PROVIDER and xenova/transformers aliases

### LINKED ISSUE #395: Bug: Local embedding stuck at "none" (BM25-only) in v0.9.13 despite correct xenova config
### Environment
- **agentmemory version**: `v0.9.13` (You can check with `agentmemory --version`)
- **Node.js version**: v24.15.0
- **OS**: Windows 11
- **iii version**: 0.11.2
- **Installation method**: `npm install -g @agentmemory/agentmemory`
### Steps to Reproduce 
1. Install the local embedding library: `npm install -g @xenova/transformers`
2. Set the embedding provider to `xenova`:
   - **Method A (environment variable)**: `export AGENTMEMORY_EMBEDDING_PROVIDER=xenova` (or `set` on Windows)
   - **Method B (.env file)**: Create `~/.agentmemory/.env` file with the line `AGENTMEMORY_EMBEDDING_PROVIDER=xenova`
3. Start the server: `agentmemory`
4. Check the startup logs.

### Expected Behavior 
- The startup log should show `Embedding provider: xenova` and successfully load the `all-MiniLM-L6-v2` model.
- The documentation [mentions](https://github.com/rohitg00/agentmemory#embedding-providers) that `Local (recommended)` is free and should work offline.

### Actual Behavior 
- The startup log shows: `Embedding provider: none (BM25-only mode)`
- It remains in BM25-only mode and never attempts to download or use the local model.

### Additional Context 
- I also tried setting `AGENTMEMORY_EMBEDDING_PROVIDER` to `local` or `transformers`, but it still shows `none`.
- The `@xenova/transformers` library can be required successfully (`node -e "const t = require('@xenova/transformers'); console.log('OK');"` outputs `OK`), so the dependency itself is fine.
- I confirmed this is unrelated to the LLM provider; the issue persists even when an `ANTHROPIC_API_KEY` is set.
### Full startup log (without any LLM API key)

```text
T  agentmemory
|
*  iii-engine is running
[agentmemory] No LLM provider key found (ANTHROPIC_API_KEY, GEMINI_API_KEY, OPENROUTER_API_KEY, MINIMAX_API_KEY). LLM-backed compression and summarization are DISABLED — using no-op provider. This is the safe default: the agent-sdk fallback used to spawn Claude Agent SDK child sessions which inherit Claude Code's plugin hooks and cause infinite Stop-hook recursion (#149 follow-up). To opt in to the agent-sdk fallback anyway, set both AGENTMEMORY_AUTO_COMPRESS=true AND AGENTMEMORY_ALLOW_AGENT_SDK=true — but be aware it will burn your Claude Pro allocation and may still recurse if you use it from inside Claude Code itself.
[agentmemory] Starting worker v0.9.13...
[agentmemory] Engine: ws://localhost:49134
[agentmemory] Provider: noop (noop)
[agentmemory] Embedding provider: none (BM25-only mode)
[agentmemory] REST API: http://localhost:3111/agentmemory/*
[agentmemory] Streams: ws://localhost:3112
[OTel] Traces initialized: engine=ws://localhost:49134, service=agentmemory
[OTel] Metrics initialized: interval=30000ms
[agentmemory] Consolidation pipeline: registered (CONSOLIDATION_ENABLED=false)
[agentmemory] Auto-compress: OFF (default, #138) — observations indexed via zero-LLM synthetic compression. Set AGENTMEMORY_AUTO_COMPRESS=true to opt-in to LLM-powered summaries (uses your API key).
[agentmemory] Context injection: OFF (default, #143) — hooks capture observations but do not inject context into Claude Code's conversation. Set AGENTMEMORY_INJECT_CONTEXT=true to opt-in (warning: expect your Claude Pro allocation to drain faster).
[agentmemory] v0.6 advanced retrieval: sliding-window, query-expansion, temporal-graph, retention-scoring
[agentmemory] Orchestration layer: actions, frontier, leases, routines, signals, checkpoints, flow-compress, mesh, branch-aware, sentinels, sketches, crystallize, diagnostics, facets
[iii] Worker registered with ID: 0012e2a6-d01f-40fb-98d0-697fca152d67
[agentmemory] Ready. BM25+Graph search active.
[agentmemory] Endpoints: 107 REST + 51 MCP tools + 6 MCP resources + 3 MCP prompts
[agentmemory] Auto-forget: enabled (every 60m)
[agentmemory] Lesson decay sweep: enabled (every 24h)
[agentmemory] Viewer: http://localhost:3113

AyushKar2005:
so i was looking at issue #395 and traced it down to `detectEmbeddingProvider` in config.ts. it was reading `EMBEDDING_PROVIDER` but the docs and startup logs all say to set `AGENTMEMORY_EMBEDDING_PROVIDER` — so the env var was just never being read.

also noticed that even if someone set it correctly to `xenova` or `transformers`, it would still fall through to null because the switch in index.ts only handles `local`. added an alias that maps both to `local`.

tested locally, embedding provider now correctly resolves instead of falling back to BM25-only mode.

Closes #395

## Summary by CodeRabbit

* **Improvements**
  * Added support for an additional override for embedding provider selection and normalized provider names for consistent behavior.
  * Improved tool discovery so the server can be configured to always expose the full tool set, even when proxying or falling back to local tooling.
* **Chores**
  * Server configuration updated to allow forcing proxy behavior and preconfiguring service endpoints for local development.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/412?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/plugin/.mcp.json
+++ b/plugin/.mcp.json
@@ -2,7 +2,12 @@
   "mcpServers": {
     "agentmemory": {
       "command": "npx",
-      "args": ["-y", "@agentmemory/mcp"]
+      "args": ["-y", "@agentmemory/mcp"],
+      "env": {
+        "AGENTMEMORY_URL": "http://127.0.0.1:3111",
+        "AGENTMEMORY_TOOLS": "all",
+        "AGENTMEMORY_FORCE_PROXY": "1"
+      }
     }
   }
 }
--- a/src/config.ts
+++ b/src/config.ts
@@ -182,8 +182,14 @@ export function detectEmbeddingProvider(
   env?: Record<string, string>,
 ): string | null {
   const source = env ?? getMergedEnv();
-  const forced = source["EMBEDDING_PROVIDER"];
-  if (forced) return forced;
+const forced = source["AGENTMEMORY_EMBEDDING_PROVIDER"] || source["EMBEDDING_PROVIDER"];
+
+if (forced) {
+
+  if (forced === "xenova" || forced === "transformers") return "local";
+  
+  return forced;
+}
 
   if (source["GEMINI_API_KEY"]) return "gemini";
   if (source["OPENAI_API_KEY"]) return "openai";
--- a/src/mcp/standalone.ts
+++ b/src/mcp/standalone.ts
@@ -378,11 +378,23 @@ export async function handleToolsList(): Promise<{ tools: unknown[] }> {
           `[@agentmemory/mcp] tools/list: remote response shape: ${shape}\n`,
         );
       }
-      if (remote && Array.isArray(remote.tools)) {
+if (remote && Array.isArray(remote.tools)) {
         if (debug) {
           process.stderr.write(
             `[@agentmemory/mcp] tools/list: returning ${remote.tools.length} tools from server\n`,
           );
+        }
+        const allTools = getAllTools();
+        if (
+          process.env["AGENTMEMORY_TOOLS"] === "all" &&
+          remote.tools.length < allTools.length
+        ) {
+          if (debug) {
+            process.stderr.write(
+              `[@agentmemory/mcp] tools/list: AGENTMEMORY_TOOLS=all override — returning ${allTools.length} tools instead of ${remote.tools.length} from server\n`,
+            );
+          }
+          return { tools: allTools };
         }
         return { tools: remote.tools };
       }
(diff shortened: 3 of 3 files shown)


rohitg00:
The `src/config.ts` core fix is great and closes #395 cleanly — but the PR bundles three separate concerns that should land as three PRs:

**1. Accept `AGENTMEMORY_EMBEDDING_PROVIDER` + xenova/transformers alias (keep this here).** Real bug, valid fix. Two follow-ups before merge:
- Indentation: `+const forced = ...` and `+if (forced) {` are left-margin in your diff (should be 2-space inside the function body). `npm run build` will fail on style.
- Add a unit test in `test/config.test.ts`: `detectEmbeddingProvider({ AGENTMEMORY_EMBEDDING_PROVIDER: 'xenova' })` returns `'local'`, `detectEmbeddingProvider({ EMBEDDING_PROVIDER: 'gemini' })` still returns `'gemini'`.

**2. `plugin/.mcp.json` env injection (drop from this PR).** Same territory as #386. Please pull these out into a separate PR so we don't have two overlapping changes to the plugin manifest.

**3. `src/mcp/standalone.ts` AGENTMEMORY_TOOLS=all override (drop from this PR).** Worth reviewing on its own — v0.9.7 (PR #146 era) already added server-side delegation for tools/list. Need to understand whether your override hides a real bug or papers over expected subsetting. Separate PR + explanation in the description please.

I'll merge the config.ts piece once it's split + tested. The other two are valuable too but need their own scope.

PR #414, opened 2026-05-15
feat: time-range filtering for memory_recall, memory_smart_search, memory_sessions (#392)

### LINKED ISSUE #392: Feature Request: Support time range filtering for memory_recall and memory_sessions
## Summary

The `memory_recall` and `memory_sessions` tools currently lack the ability to filter results by time range. This makes it difficult to retrieve memories or sessions from a specific period (e.g., "what did I work on last week?" or "show me sessions from May 1-7").

## Current Behavior

- `memory_sessions` accepts **no parameters** — it only lists the most recent sessions with no filtering capability.
- `memory_recall` supports keyword/semantic search via the `query` parameter but has **no time-based filtering**.
- `memory_smart_search` similarly lacks time range parameters.

The only workaround is to include time-related keywords in the query string (e.g., `query="May 2026"`), which relies on semantic matching and is unreliable.

## Proposed Solution

Add optional time range parameters to the relevant tools:

### `memory_sessions`
```python
memory_sessions(
    start_time: str | None = None,  # ISO 8601, e.g., "2026-05-01T00:00:00Z"
    end_time: str | None = None,    # ISO 8601, e.g., "2026-05-07T23:59:59Z"
    limit: int = 50
)
```

### `memory_recall`
```python
memory_recall(
    query: str,
    start_time: str | None = None,
    end_time: str | None = None,
    format: str = "full",
    limit: int = 10,
    token_budget: int | None = None
)
```

### `memory_smart_search`
```python
memory_smart_search(
    query: str,
    start_time: str | None = None,
    end_time: str | None = None,
    expandIds: str | None = None,
    limit: int = 10
)
```

## Use Cases

1. **Weekly/Monthly summaries**: "Show me all sessions from the past week" — currently impossible without time filtering.
2. **Debugging timeline**: "What happened between 2pm and 4pm yesterday?" — useful for incident investigation.
3. **Periodic reviews**: "List all memories created in Q1" — for quarterly reviews and knowledge management.
4. **Reducing noise**: When searching with a broad query, time range filtering helps narrow down results to the relevant period instead of returning matches from months ago.

## Additional Context

Time-based filtering is a standard feature in most memory/logging systems. Adding this would significantly improve the usability of agentmemory for time-aware retrieval scenarios.

RajvardhanPatil07:
Closes #392.

## Summary

Adds optional ISO 8601 `start_time` and `end_time` arguments (both inclusive) to `memory_recall`, `memory_smart_search`, and `memory_sessions` so agents can answer "what did I work on last week?" without keyword guessing. `memory_sessions` also gains a `limit` argument and now sorts by `startedAt` descending.

The same arguments flow through every layer that exposes these tools: REST API, MCP server, and the `@agentmemory/mcp` shim (both proxy and local-fallback modes).

## Behaviour

- `start_time` / `end_time` are optional ISO 8601 strings (e.g. `2026-05-01T00:00:00Z` or `2026-05-01`). Both bounds are inclusive.
- Any single bound is accepted (open-ended window).
- Sessions are matched by **lifetime overlap** with the window — a session that started Apr 30 and ended May 2 still matches `2026-05-01..2026-05-07`. Active sessions (no `endedAt`) are treated as still running.
- Bad input (unparseable date, `start_time > end_time`, non-string) is rejected with HTTP 400 / `code: invalid_time_range` *before* retrieval runs, so a malformed window never costs you a search.
- Recall stability under filtering: `mem::search` over-fetches 10× from the BM25 index, `HybridSearch` over-fetches BM25/vector/graph by 5×, and the diversify-by-session cap rises to 5/session when a time range is active. This is the same pattern already used for `project` / `cwd` filtering; recall@K does not collapse on narrow windows.

## Acceptance criteria from #392

| Criterion | Status |
|---|---|
| `memory_sessions(start_time, end_time, limit)` | ✅ |
| `memory_recall(query, start_time, end_time, format, limit, token_budget)` | ✅ |
| `memory_smart_search(query, start_time, end_time, expandIds, limit)` | ✅ |
| Reduces noise on broad queries | ✅ (filter applied to candidate pool, not just final top-K) |

### REST

| Method | Path | What changed |
|---|---|---|
| `GET` | `/agentmemory/sessions` | New `?start_time=&end_time=&limit=` query params. |
| `POST` | `/agentmemory/search` | New `start_time` / `end_time` in body. |
| `POST` | `/agentmemory/smart-search` | New `start_time` / `end_time` in body. |

### MCP tools (server + shim)

- `memory_recall`, `memory_smart_search`, `memory_sessions` schemas extended.
- Server (`src/mcp/server.ts`) validates the range and forwards to `mem::search` / `mem::smart-search` / direct `kv.list(KV.sessions)` + `filterSessionsByTime`.
- Shim (`@agentmemory/mcp`, `src/mcp/standalone.ts`):
  - **Proxy mode**: forwards the arguments — POST body for `memory_smart_search` and `memory_recall`, query string for `memory_sessions`.
  - **Local InMemoryKV fallback** (no server reachable): applies the filter against `memory.createdAt` for recall and `session.startedAt` / `endedAt` for sessions, so the 7-tool fallback set still respects the new arguments.

### Internals

- New shared utility `src/state/time-filter.ts`:
  - `parseTimeRange(input) -> TimeRange | null` (throws `TimeRangeError` with an HTTP-friendly `code` field).
  - `inTimeRange(timestamp, range)` — defensive against malformed timestamps in the store.
  - `filterSessionsByTime(sessions, range)` — lifetime-overlap predicate + descending sort.
- `HybridSearch.search` / `searchWithExpansion` now accept `options.timeRange`. `tripleStreamSearch` over-fetches each stream by 5× under filtering, bumps the retrieval depth to `max(limit*5, 100)`, raises the diversify-by-session cap to 5, and applies `inTimeRange` post-enrichment.

## Examples

```bash
# All sessions whose lifetime overlapped May 1–7
curl -s "http://localhost:3111/agentmemory/sessions?start_time=2026-05-01T00:00:00Z&end_time=2026-05-07T23:59:59Z&limit=50"

# Recall against a window
curl -s -X POST http://localhost:3111/agentmemory/smart-search \
  -H "content-type: application/json" \
  -d '{"query":"auth refactor","start_time":"2026-05-01T00:00:00Z","end_time":"2026-05-07T23:59:59Z"}'
```

The same arguments are accepted via MCP `tools/call`.

## What was tested

- 38 new tests across two files:
  - `test/time-filter.test.ts` (31 tests): `parseTimeRange` happy/sad paths (valid ISO, date-only, single-bound, whitespace, unparseable, `start > end`, non-string, equal bounds), `inTimeRange` (inclusivity, malformed timestamps, single-bound), `filterSessionsByTime` (lifetime overlap, descending sort, active sessions), `mem::search` (no-filter / both-bound / one-bound / error cases), `mem::smart-search` (forwards `timeRange`, validates input, returns 400-style error).
  - `test/time-filter-mcp-shim.test.ts` (7 tests): proxy-mode forwarding for all three tools, validation rejects bad input *before* the proxy call goes out, local InMemoryKV fallback applies the filter for `memory_sessions` and `memory_recall`.
- Full suite: **992/992 tests pass** (954 baseline + 38 new). `npm run build` clean.

## Out of scope / follow-ups

- Date-math shortcuts like `last_week` / `since:24h`. The issue asks for ISO 8601 strings; a future enhancement could add a small parser layer on top of `parseTimeRange`.
- Time filtering for the other 50 tools (`memory_timeline` already takes an anchor; `memory_audit` could grow `dateFrom`/`dateTo` mirroring `GovernanceFilter`). Tracked separately if desired.
- A dedicated `mem::sessions` iii function. Today the session list is materialized inline in `api::sessions` and the MCP `memory_sessions` handler; promoting it to a function would let the iii console invoke it directly. Left for a follow-up to keep this PR scoped.

## Summary by CodeRabbit

* **New Features**
  * ISO 8601 time-range filtering (inclusive start_time/end_time) for memory recall, smart search, and sessions; invalid ranges return HTTP 400 with an error code.
  * Sessions endpoint added/updated with limit (default 50, max 1000), sorted by start time desc, and active sessions treated as running.

* **Documentation**
  * README and changelog updated with parameter behavior and endpoint examples.

* **Tests**
  * Added comprehensive tests for parsing, filtering, API and proxy behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/414)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Added
+
+- **Time-range filtering on `memory_recall`, `memory_smart_search`, and `memory_sessions`** ([#392](https://github.com/rohitg00/agentmemory/issues/392)). All three tools now accept optional ISO 8601 `start_time` / `end_time` arguments (both inclusive) so agents can answer "what did I work on last week?" without keyword guessing. `memory_sessions` also gains a `limit` parameter (default 50, max 1000) and now sorts by `startedAt` descending. Sessions are matched by lifetime overlap with the window — a session that started Apr 30 and ended May 2 still matches `2026-05-01..2026-05-07`; active sessions (no `endedAt`) are treated as still running. Bad input (unparseable date or `start_time > end_time`) returns a 400 with `code: invalid_time_range` *before* retrieval runs.
+  - REST: `POST /agentmemory/search`, `POST /agentmemory/smart-search` accept `start_time` / `end_time` in the body. `GET /agentmemory/sessions` accepts `?start_time=&end_time=&limit=`.
+  - MCP server: `memory_recall`, `memory_smart_search`, `memory_sessions` schemas extended; the same arguments are forwarded to `mem::search` / `mem::smart-search`.
+  - MCP shim (`@agentmemory/mcp`): proxy mode forwards the new arguments to the running server; the local `InMemoryKV` fallback applies the filter against `memory.createdAt` (recall) and `session.startedAt` / `endedAt` (sessions).
+  - Search/recall over-fetches (10× in `mem::search`, 5× in `HybridSearch`) and bumps the diversify-by-session cap to 5/session when a time range is set, so recall@K does not collapse on narrow windows. `mem::search` filtering by `project` / `cwd` is unchanged.
+
 ## [0.9.15] — 2026-05-15
 
 DevEx overhaul. Four PRs landed simultaneously rebuilding the first-run experience to SkillKit-grade polish: splash banner + interactive agent grid + provider picker + smart-defaults preferences, `agentmemory connect <agent>` to automate native-plugin install for 8 agents, interactive `doctor` v2 with Fix/Skip/More/Quit prompts and a `--all` auto-fix flag, `agentmemory remove` for clean uninstall with destruction-plan confirmation, plus five silent-killer fixes around viewer port collisions, engine version-mismatch detection, `stop --force` override, adopt-on-attach state recording, and an npx-to-global-install hint.
--- a/README.md
+++ b/README.md
@@ -708,6 +708,22 @@ Fused with Reciprocal Rank Fusion (RRF, k=60) and session-diversified (max 3 res
 
 BM25 tokenizes Greek, Cyrillic, Hebrew, Arabic, and accented Latin out of the box. For Chinese / Japanese / Korean memories, install the optional segmenters (`npm install @node-rs/jieba tiny-segmenter`) to split CJK runs into word-level tokens; without them, agentmemory soft-falls to whole-run tokenization and prints a one-time hint on stderr.
 
+### Time-range filtering
+
+`memory_recall`, `memory_smart_search`, and `memory_sessions` all accept optional ISO 8601 `start_time` / `end_time` bounds (both inclusive). Use them to answer "what did I work on last week?" without falling back to keyword guessing.
+
+```bash
+# All sessions whose lifetime overlapped May 1–7
+curl -s "http://localhost:3111/agentmemory/sessions?start_time=2026-05-01T00:00:00Z&end_time=2026-05-07T23:59:59Z&limit=50" | jq
+
+# Recall against a specific window
+curl -s -X POST http://localhost:3111/agentmemory/smart-search \
+  -H "content-type: application/json" \
+  -d '{"query":"auth refactor","start_time":"2026-05-01T00:00:00Z","end_time":"2026-05-07T23:59:59Z"}'
+```
+
+The same arguments work via MCP tool calls (`memory_smart_search`, `memory_recall`, `memory_sessions`). For sessions, the filter checks lifetime overlap with the window — a session that started April 30 and ended May 2 still matches `2026-05-01..2026-05-07`. Active (no `endedAt`) sessions are treated as still running. Bad input (unparseable date or `start_time > end_time`) is rejected with a 400 / `code: invalid_time_range` before the search runs, so you don't pay for retrieval on a malformed window.
+
 ### Embedding providers
 
 agentmemory auto-detects your provider. For best results, install local embeddings (free):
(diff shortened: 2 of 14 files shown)


MarvinFS:
Heads-up on a related shim bug you'll likely touch in the same file: #440. The validator in `src/mcp/standalone.ts` for `memory_recall` / `memory_smart_search` currently only extracts `query` and `limit` and drops everything else - including the `format` parameter that `tools-registry.ts` advertises as defaulting to `"full"`. Worse, both tools are aliased to `/agentmemory/smart-search` (compact-by-design), so even when callers pass `format:"full"` it never reaches the daemon that would honor it.

Net effect through MCP today: titles + scores, no bodies. That defeats the whole point of having a memory system. A ranked list of "you have a memory called 'chatterbox tts version plan'" without the actual plan is just an index - the agent still has to make a second call to get the content, except the second call (`expandIds`) is ALSO dropped by the same validator, so there's no working path to content from inside MCP. You have to drop to direct REST to get bodies, which kills the abstraction.

Since your PR is already adding `start_time` / `end_time` to that same validator and proxy handler, you're in the right spot to either:

1. Split `memory_recall` from `memory_smart_search` in the validator+proxy so `memory_recall` routes to `/agentmemory/search` with `format`, while `memory_smart_search` keeps its compact-then-expand pattern (with `expandIds` actually forwarded).
2. Or just forward all advertised params verbatim in your changes so future param additions don't keep silently disappearing.

#440 has the suggested diff. Not asking you to scope-creep this PR - just flagging so whoever reviews catches the overlap. If you'd rather keep this PR tight, totally fair, and I can open a separate PR after yours merges.

The bigger point is the framing: a memory tool whose primary output is "here are 5 memory titles that might be relevant" instead of "here's what you stored" defeats the entire abstraction. Ranking without content forces every caller into a round-trip that currently doesn't exist in the MCP surface.

RajvardhanPatil07:
Good catch, thanks. I folded #440 into this PR since it overlaps the same MCP shim path.

This now splits `memory_recall` from `memory_smart_search`: recall proxies to `/agentmemory/search` with `format`/`token_budget`/time bounds, smart search keeps `/agentmemory/smart-search` and forwards `expandIds`. I also loosened the MCP schema/server path so expandIds-only calls work for the compact-then-expand flow.

Validated with `npm test` and `npm run build`.

MarvinFS:
Quick follow-up: I went ahead and built a local patch that splits `memory_recall` from `memory_smart_search` in `src/mcp/standalone.ts` and forwards `format` / `token_budget` / `expandIds` through to the daemon. Tested end-to-end against a real installation and `memory_recall` with `format:"full"` now returns full observation bodies (narrative, facts, rules, concepts) instead of just titles + scores.

51 insertions / 4 deletions, all in `src/mcp/standalone.ts`. The diff and rationale are in #440. Happy to open a separate PR after yours merges so we don't conflict, since you're already adding `start_time`/`end_time` to the same validator. Or if you want, the validator split makes adding your time-range fields cleaner too - `memory_recall` and `memory_smart_search` end up as separate cases that can extract whatever each tool's schema declares.

Either way, just wanted to confirm the proposed fix in #440 actually works in practice. Running against our patched build now.

RajvardhanPatil07:
Thanks for testing this end-to-end, that’s really helpful.

The validator split sounds like the right direction to me, especially if `memory_recall` and `memory_smart_search` now need to forward different schema-specific fields. I’m happy to keep my `start_time` / `end_time` change scoped for now, and if your PR lands first I can rebase onto the split and add the time-range fields in the cleaner separate case.

Feel free to open the PR when ready and link it here. If mine merges first, yours should still be straightforward to rebase since the overlap is limited to `src/mcp/standalone.ts`.

Issue #149, opened 2026-04-15
429 rate limit errors when AUTO_COMPRESS uses agent-sdk provider in a heavy Claude Code session


SChinomi:
## What happened

1. Installed agentmemory v0.8.11, configured hooks in Claude Code's `settings.json` (SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, Stop, SubagentStop)
2. Set `AGENTMEMORY_AUTO_COMPRESS=true` in `~/.agentmemory/.env`
3. Intended to use Gemini as the LLM provider, but the API key variable name was incorrect (`GOOGLE_API_KEY` instead of `GEMINI_API_KEY`), so `detectProvider()` fell through to `agent-sdk`
4. Started agentmemory — startup log showed `Provider: agent-sdk (claude-sonnet-4-20250514)`
5. Continued working in a heavy Claude Code session (Opus model, multiple subagents, frequent tool calls)
6. Within minutes, agentmemory logs showed compression failures:
   ```
   error Compression failed {"obsId":"obs_mo0n4cse_80b6e1a62a8f","error":"Claude Code returned an error result: API Error: Server is temporarily limiting requests (not your usage limit) · Rate limited"}
   ```
7. Claude Pro subscription usage jumped from ~30% to 100%
## Environment

- agentmemory v0.8.11
- iii-engine v0.11.0
- macOS (Apple Silicon)
- Claude Code with Pro subscription (Opus model)
- Heavy session: multiple subagents, ~50+ tool calls

PR #418, opened 2026-05-15
Security: Bearer token sent over plaintext HTTP

tuanaiseo:
## Problem

The hook scripts (stop.ts, post-tool-use.ts, etc.) send Bearer tokens in Authorization headers to REST_URL. When AGENTMEMORY_URL defaults to 'http://localhost:3111', tokens are transmitted over plaintext HTTP. While localhost traffic typically doesn't leave the machine, this could be intercepted by local malware or other processes on multi-tenant systems.

**Severity**: `high`
**File**: `src/hooks/stop.ts`

## Solution

Enforce HTTPS or validate that the URL uses https:// protocol. Consider adding a configuration check that errors if AGENTMEMORY_SECRET is set but AGENTMEMORY_URL uses http:// (non-loopback). The integrations/pi/security.ts already has a guard for this pattern - consider centralizing this logic.

## Changes

- `src/hooks/stop.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security validation to prevent sensitive authentication credentials from being transmitted over unsecured network connections. When an authorization secret is configured, the application now enforces HTTPS encryption for all remote connections or restricts to local loopback addresses only, ensuring bearer tokens are never accidentally exposed via unencrypted HTTP.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/418?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/hooks/stop.ts
+++ b/src/hooks/stop.ts
@@ -12,6 +12,23 @@ function isSdkChildContext(payload: unknown): boolean {
 const REST_URL = process.env["AGENTMEMORY_URL"] || "http://localhost:3111";
 const SECRET = process.env["AGENTMEMORY_SECRET"] || "";
 
+function isSecureUrl(url: string): boolean {
+  try {
+    const parsed = new URL(url);
+    if (parsed.protocol === "https:") return true;
+    const hostname = parsed.hostname.toLowerCase();
+    if (hostname === "localhost" || hostname === "127.0.0.1" || hostname === "::1") return true;
+    return false;
+  } catch {
+    return false;
+  }
+}
+
+if (SECRET && !isSecureUrl(REST_URL)) {
+  console.error("Error: AGENTMEMORY_SECRET is set but AGENTMEMORY_URL uses insecure http:// protocol. Use https:// or a loopback address.");
+  process.exit(1);
+}
+
 function authHeaders(): Record<string, string> {
   const h: Record<string, string> = { "Content-Type": "application/json" };
   if (SECRET) h["Authorization"] = `Bearer ${SECRET}`;


rohitg00:
Thanks for catching this. Two changes needed before merge:

1. **Scope is too narrow.** The fix only patches `src/hooks/stop.ts`, but the same `AGENTMEMORY_SECRET + REST_URL` pattern lives in 12 hook files: `session-start.ts`, `session-end.ts`, `prompt-submit.ts`, `pre-tool-use.ts`, `post-tool-use.ts`, `post-tool-failure.ts`, `pre-compact.ts`, `subagent-start.ts`, `subagent-stop.ts`, `notification.ts`, `task-completed.ts`, `stop.ts`. The guard needs to fire on all of them or none — otherwise a token still leaks via `session-start` even when `stop` refuses.

2. **Align with the existing pattern.** `integrations/pi/security.ts` already implements this via `createPlaintextBearerAuthGuard()` + `AGENTMEMORY_REQUIRE_HTTPS=1` opt-in (closes #149's design call). Hard-failing with `process.exit(1)` on every hook call is a behavioural break for users who deploy behind a TLS-terminating reverse proxy where the worker URL appears as `http://internal-host:3111` but the bearer is end-to-end safe.

Suggested approach: extract a shared `src/hooks/auth-guard.ts` helper that mirrors `integrations/pi/security.ts:createPlaintextBearerAuthGuard` — emits a one-shot warning by default, throws only when `AGENTMEMORY_REQUIRE_HTTPS=1` is set. Import it into all 12 hooks at module-init time so the behaviour is consistent.

I'll merge once the guard covers every hook file and the opt-in env var matches the existing surface.

rohitg00:
Hey @tuanaiseo — thanks for the report. Two changes needed before this can land:

**1. Module-level `process.exit(1)` breaks build + test paths.**

```ts
// shipped in this PR (top of stop.ts):
if (SECRET &amp;&amp; !isSecureUrl(REST_URL)) {
  process.exit(1);  // ← runs at import, not at hook invocation
}
```

Anything that imports `stop.ts` (vitest collection, type-check inspection, bundlers walking the entry tree) triggers `process.exit(1)` before the test or build can run. Move the guard inside `main()` so it fires on actual hook execution, not on module load.

**2. Partial mitigation.**

`stop.ts` is one of 12 hook scripts. The other 11 (`session-start.mjs`, `prompt-submit.mjs`, `pre-tool-use.mjs`, etc.) still POST the bearer to `REST_URL` in plaintext. Patching only one hook creates a false sense of safety while leaving the same attack vector open on every other code path.

The PR body itself says the right shape ("centralize this check in a shared utility module imported by all hooks") — please deliver that shape:

- New `src/hooks/_secure-url.ts` with the `isSecureUrl` + `assertSecureBearerTransport` helpers
- Import it from all 12 hook scripts in `src/hooks/*.ts`
- Guard fires inside `main()` (or the equivalent entry function) on each one
- Regression test asserting one of the hooks exits non-zero when `SECRET` is set + `REST_URL` is plaintext

Happy to review again once the centralization is done.

PR #442, opened 2026-05-17
fix(viewer): allow non-loopback bind via AGENTMEMORY_VIEWER_HOST

### LINKED ISSUE #434: Fly deployment proxy doesn't work
I tried to follow README.md in `deploy/fly` but I can't access dashboard via `fly proxy`

```
fly proxy 3113:3113 --app "$APP"
Proxying localhost:3113 to remote [xxx.internal]:3113
````

```
❯ curl localhost:3113
curl: (56) Recv failure: Connection reset by peer
❯ curl 127.0.0.1:3113
curl: (56) Recv failure: Connection reset by peer
```

### PR DESCRIPTION
**Pull request by wyh0626**
## Summary

Closes #434. `startViewerServer` hardcodes `server.listen(port, "127.0.0.1")`. On Fly, traffic from `fly proxy` enters via the `fly-local-6pn` WireGuard interface, not loopback, so the kernel RSTs the connection. That's the `curl: (56) Recv failure: Connection reset by peer` the issue reports.

## Fix

A two-line helper:

```ts
export function resolveViewerHost(): string {
  return process.env.AGENTMEMORY_VIEWER_HOST?.trim() || "127.0.0.1";
}
```

`startViewerServer` calls it once and uses the result. The default stays `127.0.0.1`, so every local install is unaffected and the v0.9.12 plaintext-bearer guard (which assumes bearer tokens only ride loopback) still holds.

The Fly image opts in by setting `AGENTMEMORY_VIEWER_HOST=::` in `deploy/fly/Dockerfile`. That makes the viewer listen on `fly-local-6pn` (IPv6) as well as loopback. `fly.toml` is untouched: port 3113 is still not declared as a Fly service, so the viewer remains reachable only through `fly proxy` (mTLS over WireGuard).

## Verify

Unit:

```bash
npx vitest run test/viewer-host.test.ts
# 6/6 pass
```

End-to-end on Fly. I built the patched image locally with `npm pack` + a tarball install in the Dockerfile, deployed to a Singapore machine, then ran the exact reproducer from the issue:

```bash
fly proxy 3113:3113 --app "$APP"

curl -I http://localhost:3113/
# HTTP/1.1 404 Not Found
# (before fix: curl: (56) Recv failure: Connection reset by peer)

curl -s http://localhost:3113/ | head -c 80
# <!DOCTYPE html>...<title>agentmemory viewer</title>

curl -s http://localhost:3113/agentmemory/livez
# {"service":"agentmemory","status":"ok"}
```

HEAD on `/` returns 404 because the viewer only registers a GET handler there; GET serves the dashboard HTML. The signal is that the TCP+HTTP handshake completes at all, no RST. App and volume torn down after.

## Tests

6 cases in `test/viewer-host.test.ts`:

- `resolveViewerHost` defaults to `127.0.0.1` when the env var is unset
- same when it's an empty string
- returns the configured value when set
- trims surrounding whitespace
- `startViewerServer` actually binds to `127.0.0.1` by default
- `startViewerServer` binds to `::1` when `AGENTMEMORY_VIEWER_HOST=::1`

## Summary by CodeRabbit

* **New Features**
  * Configurable viewer network binding for non-loopback deployments; non-loopback binds require an allowed-hosts list and a Bearer token for API access while static assets stay public
  * In-browser auth prompt that stores a session token and auto-attaches it to API requests

* **Documentation**
  * Expanded Fly.io deployment guide covering networking, allowed-hosts, auth requirements, security warning, and secret-retrieval fallback

* **Tests**
  * Added tests for host binding, inbound auth gating, token storage, and related runtime behaviors

### DIFF
--- a/deploy/fly/README.md
+++ b/deploy/fly/README.md
@@ -75,6 +75,31 @@ fly proxy 3113:3113 --app "$APP"
 viewer's bearer token still has to ride a loopback connection on your
 laptop — the v0.9.12 plaintext-bearer guard stays satisfied.
 
+The entrypoint sets `AGENTMEMORY_VIEWER_HOST=::` **only when it detects
+Fly's runtime variables** (`FLY_APP_NAME` / `FLY_ALLOC_ID`). That makes
+the viewer listen on the machine's `fly-local-6pn` WireGuard interface
+as well as loopback so `fly proxy` can reach it. The same branch
+preseeds `VIEWER_ALLOWED_HOSTS=localhost:3113,127.0.0.1:3113,[::1]:3113`,
+which are the Host headers `fly proxy 3113:3113` actually emits on
+your laptop.
+
+When `AGENTMEMORY_VIEWER_HOST` is non-loopback the viewer enforces two
+extra guards: it refuses to start unless `VIEWER_ALLOWED_HOSTS` is
+explicitly set, and every request to `/agentmemory/*` must present
+`Authorization: Bearer $AGENTMEMORY_SECRET`. Static HTML and the
+favicon are still served unauthenticated. Browser-based viewer UX
+through `fly proxy` is therefore limited until the embedded UI learns
+to send the bearer — use `curl` (with the bearer) against the REST
+endpoints for now.
+
+> **Security warning.** Setting `AGENTMEMORY_VIEWER_HOST=0.0.0.0` or
+> `::` turns the viewer into a network-reachable proxy that signs every
+> upstream call with `AGENTMEMORY_SECRET`. Never enable that outside a
+> network you trust (Fly's WireGuard mesh in this template), and never
+> set it in a plain `docker run -p 3113:3113 …` on a shared host — the
+> entrypoint deliberately skips the override when Fly env vars are
+> absent so a plain Docker pull stays loopback-only.
+
 ## Rotate the HMAC secret
 
 ```bash
--- a/deploy/fly/README.md
+++ b/deploy/fly/README.md
@@ -51,7 +51,15 @@ fly logs --app "$APP" | grep -A1 AGENTMEMORY_SECRET=
 
 You will see exactly one line of the form `AGENTMEMORY_SECRET=<64 hex chars>`.
 Copy it into your client environment (`~/.bashrc`, Claude Desktop config,
-etc.). The secret is never printed again on subsequent boots.
+the viewer unlock prompt, etc.). The secret is never printed again on
+subsequent boots.
+
+If the first-boot log line is no longer available, read the persisted
+secret from the mounted volume:
+
+```bash
+fly ssh console --app "$APP" -C "sh -lc 'cat /data/.hmac'"
+```
 
 ## Verify the deployment
 
@@ -87,10 +95,11 @@ When `AGENTMEMORY_VIEWER_HOST` is non-loopback the viewer enforces two
 extra guards: it refuses to start unless `VIEWER_ALLOWED_HOSTS` is
 explicitly set, and every request to `/agentmemory/*` must present
 `Authorization: Bearer $AGENTMEMORY_SECRET`. Static HTML and the
-favicon are still served unauthenticated. Browser-based viewer UX
-through `fly proxy` is therefore limited until the embedded UI learns
-to send the bearer — use `curl` (with the bearer) against the REST
-endpoints for now.
+favicon are still served unauthenticated. If a proxied viewer request
+gets a 401, the browser UI prompts for `AGENTMEMORY_SECRET` and stores
+it in session storage so subsequent viewer API calls include the bearer.
+Use the value printed in the first-boot logs or read `/data/.hmac`
+inside the machine.
 
 > **Security warning.** Setting `AGENTMEMORY_VIEWER_HOST=0.0.0.0` or
 > `::` turns the viewer into a network-reachable proxy that signs every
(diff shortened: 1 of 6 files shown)


rohitg00:
Hey @wyh0626 — appreciate the contribution, but blocking on a security regression before this can land.

The viewer has no inbound auth — its only defense beyond loopback bind is the Host header allowlist in `buildAllowedHosts()`. Binding to `::` (or `0.0.0.0`) exposes a privileged proxy that holds the agentmemory REST bearer token in `process.env.AGENTMEMORY_SECRET` and forwards arbitrary requests upstream. Any LAN peer that spoofs `Host: localhost:3113` lands the same trust level as a local user.

Specifically:

1. **`deploy/fly/Dockerfile` bakes `AGENTMEMORY_VIEWER_HOST=::`** — any user who pulls this image and runs `docker run -p 3113:3113` outside the documented fly setup silently exposes the proxy. The ENV needs to move to an entrypoint that asserts the fly-only context (or a runtime guard).

2. **No Host-allowlist tightening when bind is non-loopback.** Either auto-derive `VIEWER_ALLOWED_HOSTS` from `AGENTMEMORY_VIEWER_HOST` + the configured fly app domain, or refuse to start unless `VIEWER_ALLOWED_HOSTS` is explicitly set.

3. **No inbound auth.** Even with the allowlist tightened, the viewer needs to require `AGENTMEMORY_SECRET` on its own inbound path when bound to non-loopback (not just on the upstream REST call). Reuse the existing HMAC pattern from `src/auth.ts` — same Bearer header, timing-safe compare.

4. **README needs an explicit warning** that `AGENTMEMORY_VIEWER_HOST=0.0.0.0` exposes a bearer-authorized proxy without inbound auth by default.

This PR also currently conflicts with main on `src/viewer/server.ts`. Rebase + add the 4 guards above and I'll review again.

wyh0626:
Updated this PR against the latest `main` and tightened the Fly viewer path.

Changes made:

1. Kept the default viewer behavior unchanged
   - The viewer still binds to `127.0.0.1` by default.
   - Local runs, plain Docker, and non-Fly deployments are unaffected unless they explicitly set `AGENTMEMORY_VIEWER_HOST` to a non-loopback address.

2. Added explicit non-loopback viewer authentication
   - When `AGENTMEMORY_VIEWER_HOST` is set to `0.0.0.0` or `::`, the viewer now requires `AGENTMEMORY_SECRET`.
   - API/proxy requests must include `Authorization: Bearer <AGENTMEMORY_SECRET>`.
   - Static viewer HTML remains loadable so the browser can show the unlock prompt.

3. Tightened Host allowlisting for non-loopback binds
   - Loopback-derived default hosts are no longer trusted when the viewer binds to a non-loopback address.
   - Non-loopback mode now requires explicit `VIEWER_ALLOWED_HOSTS`.

4. Improved operator-facing errors
   - Missing `AGENTMEMORY_SECRET` now explains how to fix it:
     either unset `AGENTMEMORY_VIEWER_HOST` to return to the safe loopback bind, or set `AGENTMEMORY_SECRET`.
   - Missing `VIEWER_ALLOWED_HOSTS` now explains that non-loopback binds require explicit trusted Host headers and gives a concrete example.

5. Avoided silent 403s from port fallback
   - In non-loopback mode, the viewer no longer retries on `EADDRINUSE`.
   - This avoids drifting from `3113` to `3114+` while `VIEWER_ALLOWED_HOSTS` still contains the old port.
   - Loopback/default mode keeps the previous retry behavior.

6. Made startup logs clearer
   - Non-loopback viewer startup logs now include the bind host, that inbound Bearer auth is required, and the configured allowed Host headers.

7. Added browser unlock flow
   - The viewer shows an `AGENTMEMORY_SECRET` prompt when API calls return 401.
   - The token is stored in `sessionStorage` only.

8. Updated Fly docs
   - Documented how to get the generated secret from first-boot logs or `/data/.hmac`.
   - Documented the viewer auth prompt and Fly proxy flow.

imcvampire:
@wyh0626 is there any update on this PR? Is there anything I could help you push it forward?

wyh0626:
@rohitg00 friendly ping on this one 🙏

All four points from your security review are addressed in the latest revision:

1. **Inbound auth** — when bound to a non-loopback host, the viewer now requires `AGENTMEMORY_SECRET` and validates `Authorization: Bearer <secret>` with a timing-safe compare (reusing the `src/auth.ts` HMAC pattern). Static HTML still loads so the browser can show the unlock prompt.
2. **Dockerfile** — non-loopback bind no longer silently exposes the proxy; missing `AGENTMEMORY_SECRET` / `VIEWER_ALLOWED_HOSTS` now fails startup with an actionable error instead of starting unprotected.
3. **Host allowlist** — loopback-derived defaults are dropped on non-loopback binds; an explicit `VIEWER_ALLOWED_HOSTS` is now required. Also removed the `EADDRINUSE` port fallback in this mode so the bind can't drift off the allowlisted port.
4. **README** — added the explicit warning about exposing a bearer-authorized proxy.

Default behavior is unchanged: still binds to `127.0.0.1`, local/plain-Docker/non-Fly setups are unaffected.

Happy to rebase or make any further changes if something's still blocking. Thanks for the review!

wyh0626:
> @wyh0626 is there any update on this PR? Is there anything I could help you push it forward?

@imcvampire thanks for the nudge! The PR is fully updated — all of the maintainer's review points are resolved and CI is green, so it's just waiting on a re-review.

Issue #371, opened 2026-05-14
Consolidate OpenAI LLM + embedding providers into one class


rohitg00:
## Why

Once #307 lands, `OpenAIProvider` (LLM, raw fetch against `/v1/chat/completions`) and `OpenAIEmbeddingProvider` (embeddings, raw fetch against `/v1/embeddings`) will share:

- The same `OPENAI_API_KEY` env
- The same `OPENAI_BASE_URL` env
- The same authorization header pattern
- The same supported-endpoints list (OpenAI, DeepSeek, SiliconFlow, vLLM, LM Studio, Ollama via `/v1`)

…but live in two separate classes with duplicated transport boilerplate (headers, error handling, fetch wrapper, retry / timeout when we add it). #199 proposed merging these.

## Proposal

A single `src/providers/openai.ts` exposes both surfaces from one class:

```ts
export class OpenAIProvider implements MemoryProvider, EmbeddingProvider {
  async compress(systemPrompt: string, userPrompt: string): Promise<string> { ... }
  async summarize(systemPrompt: string, userPrompt: string): Promise<string> { ... }
  async embed(text: string): Promise<Float32Array> { ... }
  async embedBatch(texts: string[]): Promise<Float32Array[]> { ... }
}
```

Internal `call()` helper takes a path (`"/v1/chat/completions"` or `"/v1/embeddings"`) and a body shape. Both factories (`createBaseProvider` in `src/providers/index.ts`, the embedding factory) instantiate the same class.

Expected delta: ~50 net LOC reduction once dup transport collapses.

## Out of scope

- Adding new provider features. Pure refactor.
- Azure-shaped URL handling (separate follow-up).

## Effort

Small — ~1 day implementation + tests.

## Closes
Supersedes #199.

PR #467, opened 2026-05-17
fix(embeddings): allow custom OpenAI embedding path

honor2030:
## Summary
- Adds `OPENAI_EMBEDDING_PATH` so OpenAI-compatible embedding providers can use custom embedding routes without duplicating `/v1`.
- Trims trailing slashes from `OPENAI_BASE_URL` before appending the embedding path.
- Documents the new setting in `.env.example` and README.

## Notes
- Related to the OpenAI-compatible provider work in #462, but this is a separate small slice: configurable embedding route path only.
- Default behavior remains `https://api.openai.com/v1/embeddings`.

## Test Plan
- [x] RED: `npm test -- --run test/embedding-provider.test.ts` failed because `OPENAI_BASE_URL=http://localhost:11434/v1/` still produced `/v1//v1/embeddings`
- [x] GREEN: `npm test -- --run test/embedding-provider.test.ts` → 18 passed
- [x] `npm run build`
- [x] `npm test` → 91 files passed, 1008 tests passed

## Summary by CodeRabbit

* **New Features**
  * Support for customizing the OpenAI embeddings endpoint path via a new OPENAI_EMBEDDING_PATH environment variable, improving compatibility with proxies or alternate routing.

* **Documentation**
  * README and example configuration updated to document OPENAI_EMBEDDING_PATH, including normalization guidance (leading slash) and usage examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/467?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -74,6 +74,7 @@
 # COHERE_API_KEY=...                             # General-purpose embeddings
 
 # Reuses OPENAI_API_KEY / OPENAI_BASE_URL above when EMBEDDING_PROVIDER=openai.
+# OPENAI_EMBEDDING_PATH=/v1/embeddings          # Override when a proxy expects a different embeddings route
 # OPENAI_EMBEDDING_MODEL=text-embedding-3-small  # Embedding model when EMBEDDING_PROVIDER=openai
 # OPENAI_EMBEDDING_DIMENSIONS=1536               # Required when the model is not in the known-models table
 
--- a/README.md
+++ b/README.md
@@ -1090,6 +1090,7 @@ Create `~/.agentmemory/.env`:
 # VOYAGE_API_KEY=...
 # OPENAI_API_KEY=sk-...
 # OPENAI_BASE_URL=https://api.openai.com   # Override for Azure / vLLM / LM Studio / proxies
+# OPENAI_EMBEDDING_PATH=/v1/embeddings     # Override when a proxy expects a different embeddings route
 # OPENAI_EMBEDDING_MODEL=text-embedding-3-small
 # OPENAI_EMBEDDING_DIMENSIONS=1536        # Required when the model is not in the known-models table
 
--- a/src/providers/_openai-shared.ts
+++ b/src/providers/_openai-shared.ts
@@ -104,13 +104,15 @@ export function buildEmbeddingUrl(
   baseUrl: string,
   isAzure: boolean,
   azureApiVersion: string,
+  embeddingPath?: string,
 ): string {
+  const route = embeddingPath ?? "/embeddings";
   if (isAzure) {
     return azureStyleOf(baseUrl) === "legacy"
-      ? legacyAzureUrl(baseUrl, "/embeddings", azureApiVersion)
-      : v1AzureUrl(baseUrl, "/embeddings");
+      ? legacyAzureUrl(baseUrl, route, azureApiVersion)
+      : v1AzureUrl(baseUrl, route);
   }
-  return `${baseUrl}/v1/embeddings`;
+  return embeddingPath ? `${baseUrl}${route}` : `${baseUrl}/v1/embeddings`;
 }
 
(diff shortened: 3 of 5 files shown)


honor2030:
Rebased this branch on current `main` and resolved the conflict with the shared OpenAI/Azure URL helper changes.

Follow-up included:
- kept Azure embedding URL behavior routed through `buildEmbeddingUrl(...)`
- preserved `OPENAI_EMBEDDING_PATH` for OpenAI-compatible custom embedding routes
- added the requested no-leading-slash regression case for `OPENAI_EMBEDDING_PATH="embeddings"`

Verification:
- `PATH=/opt/homebrew/bin:$PATH npx vitest run test/embedding-provider.test.ts test/openai-shared.test.ts --reporter=verbose` — 40 passed
- `PATH=/opt/homebrew/bin:$PATH npm run build` — passed
- `PATH=/opt/homebrew/bin:$PATH npm test -- --reporter=dot` — 1036 passed
- `git diff --check origin/main...HEAD` — passed

GitHub Actions `test (20)` and `test (22)` are green on the pushed commit. The remaining Vercel status is the external fork deployment authorization gate.

rohitg00:
Hey @honor2030 — this PR currently conflicts with main because PR #371 (merged after this one was opened) refactored `src/providers/embedding/openai.ts` to use shared `normalizeBaseUrl` + `buildEmbeddingUrl` helpers from `_openai-shared.ts`. The lines this PR patches (`${this.baseUrl}/v1/embeddings`, `getEnvVar("OPENAI_BASE_URL") || DEFAULT_BASE_URL`) no longer exist on main.

There's also overlap with #503 (efenex's PR), which adds `OPENAI_EMBEDDING_BASE_URL` + `OPENAI_EMBEDDING_API_KEY` via the existing helper. That covers the 'point embedding at a different OpenAI-compat endpoint' use case cleanly.

Two paths forward:

1. **Close as superseded by #503** — if your need is fully solved by setting a different base URL/key for embeddings, #503 is the cleaner integration with the post-#371 architecture.
2. **Rebase + reshape** — if `OPENAI_EMBEDDING_PATH` (the path-only knob, not just the URL) is still needed for your case, rebase onto current main and add the path override on top of `buildEmbeddingUrl` in `_openai-shared.ts`. Worth a tracked issue describing the use case first — what endpoint shape does the path-only knob unlock that a full URL override doesn't?

Let me know which one fits your need.

skywalker-35:
Thanks for this PR! I hit the same underlying issue from a different angle — see #646 for details.

**Context:** My `OPENAI_BASE_URL` is `https://open.bigmodel.cn/api/paas/v4` (Zhipu/智谱). The base URL already has a version segment (`/v4`), but unlike standard OpenAI, it does NOT use `/v1/` at all. The current `buildEmbeddingUrl()` unconditionally appends `/v1/embeddings`, producing `v4/v1/embeddings` → 404.

Adding `OPENAI_EMBEDDING_PATH` as a workaround helps, but it requires users to know about this env var and set it correctly. A more robust approach would be to detect whether the base URL already ends with a version path segment (`/v\d+$`) and skip `/v1/` in that case:

```js
return /\/v\d+$/.test(baseUrl) ? `${baseUrl}/embeddings` : `${baseUrl}/v1/embeddings`;
```

This fixes all three known cases:
- Standard OpenAI (`api.openai.com`) → `/v1/embeddings` ✅
- Zhipu (`open.bigmodel.cn/api/paas/v4`) → `/v4/embeddings` ✅
- DeepSeek/SiliconFlow/Ollama (`host/v1`) → `/v1/embeddings` ✅ (no double `/v1/v1`)
- Azure (already handled via `isAzure` branch) ✅

Would you be open to incorporating this into your PR instead of (or in addition to) `OPENAI_EMBEDDING_PATH`?

PR #472, opened 2026-05-17
fix(summarize): chunk large sessions to fit LLM context window

efenex:
## Summary

JSONL-imported sessions can have far more observations than the 500-cap `MAX_OBS_PER_SESSION` that constrains native sessions. `mem::summarize` previously built one prompt containing every observation and shipped it as a single LLM call, which exceeded the provider's context window for sessions >~7,000 observations and returned an unhelpful 400 from upstream - silently leaving large bulk-imported sessions out of the semantic tier.

This PR makes `mem::summarize` chunk + map-reduce when needed, with per-chunk retry, skip on persistent failure, and parallel batching.

## Problem

- Native flow caps sessions at `MAX_OBS_PER_SESSION=500` via `mem::observe`'s guard (`src/functions/observe.ts:128`).
- JSONL import (`mem::replay::import-jsonl` in `src/functions/replay.ts`) writes directly to KV, bypassing the guard. Real-world imports produce sessions with 5k–100k+ observations.
- `mem::summarize` then builds one prompt with every observation and ships it via `provider.summarize(...)`. Sessions above ~7k obs exceed the provider's context window; the upstream returns a generic 400; the summary never lands; the session is invisible to `consolidate-pipeline → semantic / reflect`.

## Approach

Map-reduce inside `mem::summarize`:

  - Sessions ≤ `SUMMARIZE_CHUNK_SIZE` (default **400**) take the legacy single-call path with **zero behavior change and zero overhead**.
  - Larger sessions:
    1. Split into chunks of `SUMMARIZE_CHUNK_SIZE`
    2. Summarize each chunk with the existing per-session prompt, in parallel batches of `SUMMARIZE_CHUNK_CONCURRENCY` (default **6**)
    3. Merge partials via a new `REDUCE_SYSTEM` / `buildReducePrompt`

### Robustness

  - **Per-chunk retry-once** on transient parse failures or provider errors
  - **Persistently-failing chunks are skipped** (not propagated). A flaky chunk doesn't waste 30+ already-completed LLM calls
  - **Bail with `too_many_chunks_skipped` only if >50%** of chunks fail — guards against producing a misleadingly thin merged summary

### Configuration

  | env var | default | purpose |
  |---|---|---|
  | `SUMMARIZE_CHUNK_SIZE` | 400 | obs per chunk; ≈ 50k tokens at ~110 tok/obs (fits 128k-window models) |
  | `SUMMARIZE_CHUNK_CONCURRENCY` | 6 | parallel chunk LLM calls; high-RPM providers can allow 100+ |

## Validation (live, against a real bulk-imported corpus)

  | session size | chunks | wallclock | result |
  |---|---|---|---|
  | 5,392 obs | 14 | 39s | quality 100, 5 decisions, 30 files |
  | 10,704 obs | 27 | 34s | quality 100, 6 decisions, 56 files |
  | 105,966 obs | 265 (`c=50`) | ~12 min | persisted server-side |
  | 52-session bulk backfill | — | ~25 min | **25 new semantic facts + 6 new reflect insights** in `consolidate-pipeline` |

  ## Known limit

  The iii-engine has a hardcoded ~180s function-invocation timeout (in the Rust binary; not exposed in `iii-config.yaml`). Sessions large enough that chunked summarize wallclock exceeds that will see the HTTP client receive a timeout/500 response, even though the handler completes and persists server-side. Raising `SUMMARIZE_CHUNK_CONCURRENCY` to 50+ pushes the cliff past any realistic session size on a generous-RPM provider.

  The true fix is an async-job pattern (return jobId, poll for result). Out of scope here.

## Operator tool

  `scripts/backfill-imported-sessions.sh` walks `jsonl-import`-tagged sessions and POSTs `mem::summarize` per session, with:

  - `--project-pattern REGEX`, `--skip-agents`, `--tag NAME` (filter)
  - `--max-obs N`, `--limit N` (scope)
  - `--dry-run` (preview cost, no POSTs)
  - `--debug-on-error` (dump failed-session metadata + obs + provider response for diagnosis)
  - Cost estimator (configurable per-1M rates) so operators see ~$ before committing

## Test plan

  - `npx vitest run test/summarize.test.ts` - 9 cases
    - small session → single-call path preserved
    - large session → N chunk calls + 1 reduce
    - `SUMMARIZE_CHUNK_SIZE` env override respected
    - flaky chunk: parse-fail → retry → succeed → no skip
    - persistently-broken chunk: skipped, reduce runs on survivors
    - too many skips → `too_many_chunks_skipped` bailout
    - provider error after retry → skipped (not propagated)
    - chunks run in parallel batches per `SUMMARIZE_CHUNK_CONCURRENCY`
  - Live: multi-session bulk backfill on a real corpus

  ## Files

  - `src/prompts/summary.ts` — add `REDUCE_SYSTEM` + `buildReducePrompt`
  - `src/functions/summarize.ts` — chunking, retry, skip, parallelism (single-call path unchanged)
  - `test/summarize.test.ts` — new
  - `.env.example` — document the two new env vars
  - `.gitignore` — ignore `agentmemory-debug/` and `data-*/` (operator artefacts)
  - `scripts/backfill-imported-sessions.sh` — new, companion operator tool

## Summary by CodeRabbit

* **New Features**
  * Chunked summarization for large sessions with parallel processing and partial-summary merging.
* **Documentation**
  * Documented two optional environment variables to tune chunk size and concurrency for summarization.
* **Chores**
  * Updated ignore patterns for data/debug directories.
  * Added a backfill utility for importing and summarizing past sessions (dry-run and debug modes).
* **Tests**
  * Added comprehensive tests covering chunking, retries, skipping, and concurrency behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/472?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -98,6 +98,8 @@
 # AGENTMEMORY_GRAPH_WEIGHT=0.2                   # Graph traversal bonus on smart-search ranking
 # TOKEN_BUDGET=2000                              # Max tokens injected via mem::context per session
 # MAX_OBS_PER_SESSION=500                        # Per-session observation cap before consolidation kicks in
+# SUMMARIZE_CHUNK_SIZE=400                       # When mem::summarize sees a session larger than this, it chunks observations and map-reduces (chunk-summarize → reduce-merge) to stay within the LLM's context window. Default 400 ≈ 50k tokens per chunk at ~110 tok/obs. Native sessions are capped by MAX_OBS_PER_SESSION; chunking primarily matters for bulk-imported jsonl sessions, which bypass that cap.
+# SUMMARIZE_CHUNK_CONCURRENCY=6                  # Parallel chunk LLM calls during chunked summarize. Default 6 fits ~100-chunk sessions under iii's 180s function-invocation timeout at typical ~8s/call. High-throughput providers (Novita, DeepInfra, DeepSeek) commonly allow 100+ concurrent — bump this for very large imported sessions.
 
 # -----------------------------------------------------------------------------
 # 5. Behaviour flags
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ dist/
 plugin/scripts/*.map
 plugin/scripts/*.d.mts
 data/
+data-*/
+agentmemory-debug/
 .gstack/
 
 # Lock files — never commit (see feedback_no_lockfiles memory)
--- a/scripts/backfill-imported-sessions.sh
+++ b/scripts/backfill-imported-sessions.sh
@@ -64,17 +64,26 @@ for bin in curl jq; do
   command -v "$bin" >/dev/null || { echo "missing dependency: $bin" >&2; exit 1; }
 done
 
+# Curl timeout profiles. Metadata reads (livez, sessions list, observations
+# pull for debug dumps) should fail fast and retry transient blips. The LLM
+# work calls (summarize, consolidate) intentionally have no --retry and a
+# wide --max-time: each call can legitimately take minutes for chunked
+# summarize on large sessions, and retrying a half-finished LLM job is
+# expensive both in dollars and in duplicated server-side work.
+META_CURL_OPTS=(--connect-timeout 10 --max-time 30 --retry 2 --retry-delay 1)
+WORK_CURL_OPTS=(--connect-timeout 10 --max-time 1800)
+
 echo "agentmemory backfill — server: $URL"
 [[ "$DRY_RUN" == 1 ]] && echo "DRY RUN: no POSTs will be made."
 
 # --- liveness ---
-if ! curl -fsS "$URL/agentmemory/livez" >/dev/null; then
+if ! curl -fsS "${META_CURL_OPTS[@]}" "$URL/agentmemory/livez" >/dev/null; then
   echo "server not reachable at $URL (try: npx @agentmemory/agentmemory)" >&2
   exit 1
 fi
 
 # --- collect session ids ---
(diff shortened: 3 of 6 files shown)


efenex:
Thanks for the review — pushed [86c0622](https://github.com/rohitg00/agentmemory/commit/86c0622) addressing all four points:

- **curl timeouts**: added `META_CURL_OPTS` (connect 10s / max 30s + retry) and `WORK_CURL_OPTS` (connect 10s / max 1800s, no retry) profiles. Split because retrying an in-flight LLM call doubles the cost and the server-side handler keeps running anyway.
- **debug filename sanitization**: `tr -c 'A-Za-z0-9._-' '_'` on the session id before joining with `DEBUG_DIR`.
- **observations URL encoding**: switched to `curl --get --data-urlencode "sessionId=$id"`.
- **jq parse guard**: wrapped summarize + consolidate response parsing in `jq -e . </dev/null 2>&1` first; bad responses log \`invalid_json_response\` instead of aborting the loop.
- **test mock path**: \`vi.mock("./audit.js", ...)\` → \`"../src/functions/audit.js"\`. The old path resolved to a nonexistent module, so the mock was a silent no-op (tests passed anyway because `safeAudit` writes to mocked KV).

9/9 tests still pass.

rohitg00:
Merged + shipping in v0.9.21. Thanks @efenex — sessions >7k obs were silently failing at the LLM context limit, so this restores the consolidation pipeline for power users. Backfill script under scripts/ is a one-time operator helper for users hitting the pre-fix bug.

PR #473, opened 2026-05-17
fix(visibility): surface lessons in smart-search + tally per-store in diagnose

efenex:
## Summary

Two related UX gaps in the memory layer's reflection surfaces. A consumer that calls `memory_lesson_save` and gets `success:true` reasonably expects to find the lesson via `memory_smart_search` ("did my save land?") and to see it counted in `memory_diagnose` ("what's actually in the store?"). Today neither is true:

- **smart-search** queries only `KV.observations` — lessons are invisible
- **diagnose**'s `memories` category counts only `KV.memories` — a store with thousands of lessons / summaries / semantic facts can read as "memories: 0"

The trust-shock that prompted this fix: an MCP consumer saved a lesson, got `success:true`, then could not surface it through `memory_smart_search` (got only `obs_*` IDs), nor see any sign of it in `memory_diagnose` (reported `memories: 0`). Hypothesized the save was disconnected. It was actually persisted — just invisible to the most-prominent retrieval and health surfaces.

## A: smart-search returns lessons

- New optional `project` and `includeLessons` (default `true`) params on `mem::smart-search`
- Delegates lesson scoring to existing `mem::lesson-recall` via `sdk.trigger` — no duplicate scoring logic; future scoring tweaks in `mem::lesson-recall` auto-propagate
- Lessons come back in a separate `lessons` field on the response, **not merged into `results`**. Existing consumers reading `result.results` are unaffected. New consumers can read `result.lessons` (typed `CompactLessonResult[]`)
- Content truncated to 240 chars in compact mode (full content via `mem::lesson-recall` directly when needed)
- Lesson-recall failures are soft-handled: log warn, return empty `lessons`, observation results still flow

## B: diagnose adds per-store tally categories

- New categories: `lessons`, `summaries`, `semantic`, `procedural`, `crystals`, `insights`
- Mirrors the existing `memories` pattern: count + light consistency check (confidence range for scored memories; non-empty title/narrative/steps for the rest)
- Each new category added to `ALL_CATEGORIES` so `--categories lessons` filtering works as expected
- Tombstoned lessons (`deleted: true`) are excluded from the live count and reported separately

## Backward compatibility

- `mem::smart-search`: response shape additive — adds optional `lessons` field. Existing readers of `results` see no behavior change.
- `mem::diagnose`: empty-system pass count goes from 8 → 14 (8 original + 6 new). Existing tests for the existing categories untouched.

## Test plan

- `npx vitest run test/smart-search.test.ts test/diagnostics.test.ts` — 43 cases pass

New cases:

| file | added |
|---|---|
| `test/smart-search.test.ts` | lesson inclusion, content-preview truncation, `includeLessons=false` opt-out, project filter passthrough, soft-fail on recall error, soft-fail on non-success response |
| `test/diagnostics.test.ts` | pass + warn cases for each of the 6 new categories, categories filter accepts new names |

## Files

- `src/types.ts` — new `CompactLessonResult` interface
- `src/functions/smart-search.ts` — lesson recall + merge (single-call observation path + expand mode unchanged)
- `src/functions/diagnostics.ts` — six new category blocks inserted before the existing `mesh` block
- `test/smart-search.test.ts` — 6 new cases
- `test/diagnostics.test.ts` — 7 new cases; empty-system pass count bumped 8→14

## Out of scope

- Merging lessons into the ranked `results` array (would require normalizing observation vs lesson scoring to a comparable scale — currently the BM25/vector hybrid score and the lesson confidence/recency score live on different bases). Keeping them as a separate `lessons` field avoids forcing that decision and gives the caller full flexibility to render however they want.
- Including summaries, semantic facts, or insights in `smart-search` results. Same scoring-comparability concern, plus most consumers will want to query those tiers explicitly via their dedicated endpoints rather than incidentally in a search call.

## Summary by CodeRabbit

* **New Features**
  * Diagnostics expanded to cover lessons, summaries, semantic, procedural, crystals, and insights with per-category checks and pass/warn reporting
  * Smart search compact responses optionally include lesson recall results (enabled by default; can be disabled)

* **Types**
  * Added a compact lesson result shape to standardize lesson previews in responses

* **Tests**
  * Added comprehensive tests for new diagnostics categories and lesson inclusion behavior in smart search

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/473?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/functions/diagnostics.ts
+++ b/src/functions/diagnostics.ts
@@ -374,12 +374,21 @@ export function registerDiagnosticsFunction(sdk: ISdk, kv: StateKV): void {
         const live = lessons.filter((l) => !l.deleted);
         let lessonIssues = 0;
         for (const l of live) {
-          if (l.confidence < 0 || l.confidence > 1) {
+          // Number.isFinite rejects NaN / Infinity / non-numbers; a
+          // corrupted row passing those would silently survive the < / >
+          // range check (e.g. NaN < 0 is false, NaN > 1 is false, so the
+          // bad row would be "healthy") and skew memory_lesson_recall's
+          // scoring downstream. Surface as warning.
+          if (
+            !Number.isFinite(l.confidence) ||
+            l.confidence < 0 ||
+            l.confidence > 1
+          ) {
             checks.push({
               name: `lesson-bad-confidence:${l.id}`,
               category: "lessons",
               status: "warn",
-              message: `Lesson ${l.id} has confidence ${l.confidence} (expected 0..1)`,
+              message: `Lesson ${l.id} has confidence ${l.confidence} (expected finite number in 0..1)`,
               fixable: false,
             });
             lessonIssues++;
@@ -400,7 +409,10 @@ export function registerDiagnosticsFunction(sdk: ISdk, kv: StateKV): void {
         const summaries = await kv.list<SessionSummary>(KV.summaries);
         let summaryIssues = 0;
         for (const s of summaries) {
-          if (!s.title || !s.title.trim()) {
+          // typeof guard before .trim() — a corrupted row with title=null
+          // or title=42 would otherwise throw and abort the whole diagnose
+          // run before later categories get checked.
+          if (typeof s.title !== "string" || s.title.trim().length === 0) {
             checks.push({
               name: `summary-missing-title:${s.sessionId}`,
               category: "summaries",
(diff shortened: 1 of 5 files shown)


efenex:
Thanks for the review — pushed [bbfaa16](https://github.com/rohitg00/agentmemory/commit/bbfaa16):

- All 5 confidence checks (lesson / semantic / insight) now prefix with `Number.isFinite(...)` so `NaN` and `Infinity` surface as warnings instead of silently passing the range check
- `summary.title` and `crystal.narrative` use `typeof x !== "string"` before `.trim()` so corrupted rows produce a warn rather than throwing and aborting the whole `mem::diagnose` run
- Error messages updated to read "finite number in 0..1"
- 4 new test cases under `defensive row-shape handling` in `test/diagnostics.test.ts`, including one that explicitly verifies a later category check still executes when an earlier one had a bad row

34/34 tests pass.

rohitg00:
Merged + shipping in v0.9.21. Thanks @efenex — closes the lesson round-trip with #458: save → indexed → recalled via smart-search → counted in diagnose. The trust-shock pattern (save success but invisible to recall) was a real contract violation.

PR #489, opened 2026-05-18
fix(mcp): preserve recall format option

### LINKED ISSUE #440: MCP `memory_recall` is aliased to smart-search and drops `format` param - full content unreachable via MCP
## Summary

In `@agentmemory/mcp` v0.9.17 the standalone proxy makes full memory content unreachable through MCP. Three related defects in `src/mcp/standalone.ts`:

1. **`memory_recall` and `memory_smart_search` are aliased to the same endpoint.** Both hit `/agentmemory/smart-search`, which is compact-by-design (returns `{obsId, sessionId, title, type, score, timestamp}` only). `memory_recall` should hit `/agentmemory/search`, which honors the `format` parameter.
2. **The `format` parameter is advertised but silently dropped.** `src/mcp/tools-registry.ts:27-30` declares `memory_recall` accepts `format: full | compact | narrative` defaulting to `full`. The validator in `src/mcp/standalone.ts` only extracts `query` and `limit` - `format` never reaches the daemon.
3. **`smart-search`'s `expandIds` parameter is unreachable via MCP.** The validator doesn't extract it either, so the documented compact-then-expand pattern can only be used via direct REST.

Net effect: through MCP, callers only ever see titles + scores. The actual content IS in the database - calling `/agentmemory/search` directly with `{"query":"...", "format":"full"}` returns full observations including narrative, facts, rules, and concepts.

## Reproduction

```bash
# Through MCP - compact only
{ "tool": "memory_recall", "arguments": { "query": "chatterbox", "format": "full" } }
# Returns: { mode: "compact", results: [{obsId, title, score, ...}] }  - no content

# Through REST directly - full content
curl -X POST -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"query":"chatterbox","limit":5,"format":"full"}' \
  https://daemon/agentmemory/search
# Returns: { format: "full", results: [{observation: {title, concepts, facts: [full text], ...}}] }
```

## Source locations

- Validator that drops `format`: `src/mcp/standalone.ts` lines 113-120 (case `memory_recall` / `memory_smart_search`)
- Proxy handler that calls wrong endpoint: `src/mcp/standalone.ts` lines 162-168
- Tool schema declaring `format`: `src/mcp/tools-registry.ts` lines 13-50
- Working REST handler honoring `format`: `src/functions/search.ts` line 178

## Suggested fix

Split the two cases in both the validator and the proxy handler:

```typescript
// validate()
case "memory_recall": {
  const query = args["query"];
  if (typeof query !== "string" || !query.trim()) throw new Error("query is required");
  v.query = query.trim();
  v.limit = parseLimit(args["limit"]);
  v.format = (args["format"] as string) || "full";
  v.tokenBudget = typeof args["token_budget"] === "number" ? args["token_budget"] : undefined;
  return v;
}
case "memory_smart_search": {
  // unchanged + add expandIds extraction
  v.query = (args["query"] as string)?.trim();
  v.limit = parseLimit(args["limit"]);
  v.expandIds = normalizeList(args["expandIds"]);
  return v;
}

// handleProxy()
case "memory_recall": {
  const result = await handle.call("/agentmemory/search", {
    method: "POST",
    body: JSON.stringify({
      query: v.query,
      limit: v.limit,
      format: v.format,
      token_budget: v.tokenBudget,
    }),
  });
  return textResponse(result, true);
}
case "memory_smart_search": {
  const result = await handle.call("/agentmemory/smart-search", {
    method: "POST",
    body: JSON.stringify({ query: v.query, limit: v.limit, expandIds: v.expandIds }),
  });
  return textResponse(result, true);
}
```

The `handleLocal()` path should mirror the same split so the in-process fallback also returns content rather than just titles.

## Workaround for users

Until this ships, hit the REST API directly. A small bash helper:

```bash
curl -sk -H "Authorization: Bearer $AGENTMEMORY_SECRET" -H "Content-Type: application/json" \
  -d '{"query":"...","limit":5,"format":"full"}' \
  "$AGENTMEMORY_URL/agentmemory/search"
```

This returns the full observation including `facts[]`, `narrative`, `concepts`, etc. - which is what every caller of `memory_recall` actually wants.

## Why this matters

The whole point of a memory system is to return the memories. Returning rankings without content forces every caller into a second tool call to retrieve bodies - except that second tool call doesn't exist in the MCP surface (`expandIds` is also dropped), so today the only working path is direct REST. That defeats MCP integration for every downstream agent.

honor2030:
## Summary
- Preserve the optional `format` argument when the standalone MCP shim validates `memory_recall` / `memory_smart_search` calls.
- Forward the validated format through the proxy smart-search request and keep invalid values rejected at the boundary.

## Test Plan
- RED: `npm test -- test/mcp-standalone.test.ts` failed before the implementation when `format` was omitted from the proxied body.
- GREEN: `npm test` → 91 files / 1009 tests passed.
- GREEN: `git diff --check`

Closes #440

## Release Notes

* **New Features**
  * Added format customization to memory search operations, allowing users to choose between "compact" and "full" result formats for greater control over search output detail levels.
  * Format option defaults to "compact" when not specified.
  * Format preference is consistently applied across all search pathways.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/489?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/mcp/standalone.ts
+++ b/src/mcp/standalone.ts
@@ -64,11 +64,19 @@ function normalizeList(value: unknown): string[] {
 
 const DEFAULT_LIMIT = 10;
 const MAX_LIMIT = 100;
+type SearchFormat = "compact" | "full";
+
 function parseLimit(raw: unknown, fallback = DEFAULT_LIMIT): number {
   if (typeof raw !== "number" && typeof raw !== "string") return fallback;
   const n = Number(raw);
   if (!Number.isFinite(n) || n <= 0) return fallback;
   return Math.min(Math.floor(n), MAX_LIMIT);
+}
+
+function parseSearchFormat(raw: unknown): SearchFormat | undefined {
+  if (typeof raw !== "string") return undefined;
+  const format = raw.trim().toLowerCase();
+  return format === "compact" || format === "full" ? format : undefined;
 }
 
 function textResponse(payload: unknown, pretty = false): {
@@ -89,6 +97,7 @@ interface Validated {
   files?: string[];
   query?: string;
   limit?: number;
+  format?: SearchFormat;
   memoryIds?: string[];
   reason?: string;
 }
@@ -118,6 +127,7 @@ function validate(toolName: string, args: Record<string, unknown>): Validated {
       }
       v.query = query.trim();
       v.limit = parseLimit(args["limit"]);
+      v.format = parseSearchFormat(args["format"]);
       return v;
     }
     case "memory_sessions": {
@@ -161,9 +171,14 @@ async function handleProxy(
     }
     case "memory_recall":
     case "memory_smart_search": {
+      const body: { query?: string; limit?: number; format?: SearchFormat } = {
+        query: v.query,
+        limit: v.limit,
+      };
+      if (v.format) body.format = v.format;
       const result = await handle.call("/agentmemory/smart-search", {
         method: "POST",
-        body: JSON.stringify({ query: v.query, limit: v.limit }),
+        body: JSON.stringify(body),
       });
       return textResponse(result, true);
     }
@@ -227,6 +242,7 @@ async function handleLocal(
     case "memory_smart_search": {
       const query = (v.query || "").toLowerCase();
       const limit = v.limit ?? DEFAULT_LIMIT;
+      const format = v.format ?? "compact";
       const all =
         await kvInstance.list<Record<string, unknown>>("mem:memories");
       const results = all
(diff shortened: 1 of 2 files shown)


rohitg00:
Closed as superseded by #516 (merged today), which is the more complete fix:

- Routes `memory_recall` through `/agentmemory/search` (this PR still routed through `/smart-search`)
- Forwards both `format` and `token_budget` (this PR handled only `format`)
- Default `format: 'full'` matching the schema (this PR defaulted to `'compact'`)

Both PRs target the same file + functions, so they're mutually exclusive. Thanks @honor2030 for surfacing #440 — leaving it open as a comment trail in case anyone else hits the issue and lands here from search.

PR #494, opened 2026-05-18
fix: parse self-closing graph entities

### LINKED ISSUE #492: Graph extraction drops self-closing entity tags
## Bug

Graph extraction can drop valid entities when the LLM returns self-closing `<entity .../>` tags.

## Environment

- OS: Windows
- Node: v24.14.1
- agentmemory: 0.9.18
- Area: `mem::graph-extract`

## Repro

Use a graph extraction provider response shaped like this:

```xml
<entities>
  <entity type="file" name="src/index.ts" />
  <entity type="function" name="main"><property key="lang">typescript</property></entity>
</entities>
<relationships>
  <relationship type="uses" source="src/index.ts" target="main" weight="0.9" />
</relationships>
```

## Expected

Both entities should become graph nodes, and the relationship should be created.

## Actual

The self-closing entity is ignored by `parseGraphXml`, so the relationship cannot connect because its source node is missing.

## Notes

The current parser accepts explicit closing tags like `<entity ...></entity>` but not self-closing XML tags. This can make graph stats look empty or incomplete even when the LLM provider and extraction flag are working.

Rex57:
## Summary

- Accept self-closing graph `<entity .../>` tags in `parseGraphXml`.
- Keep existing explicit closing-tag entity parsing and property parsing behavior.
- Add regression coverage for a self-closing entity that participates in a relationship.

Fixes #492.
Related to #559: fixes one parser-loss path where real provider output with self-closing `<entity .../>` tags can parse as `0 nodes / 0 edges`.

## Verification

- `npx tsdown` passed in isolated fork clone.
- `npm run build` reached tsdown build completion but failed afterward on Windows shell-only copy fallback (`true` is not recognized / shell syntax). No live AgentMemory service was started.
- Tests not run locally per local environment constraint; full `npm test` intentionally skipped to avoid any accidental interaction with the live AgentMemory setup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XML entity parsing to support both self-closing and standard tag formats.

* **Tests**
  * Added test case for entity parsing with self-closing XML tags to ensure robust format handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/494?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/functions/graph.ts
+++ b/src/functions/graph.ts
@@ -27,12 +27,12 @@ function parseGraphXml(
   const now = new Date().toISOString();
 
   const entityRegex =
-    /<entity\s+type="([^"]+)"\s+name="([^"]+)"[^>]*>([\s\S]*?)<\/entity>/g;
+    /<entity\s+type="([^"]+)"\s+name="([^"]+)"[^>]*(?:\/>|>([\s\S]*?)<\/entity>)/g;
   let match;
   while ((match = entityRegex.exec(xml)) !== null) {
     const type = match[1] as GraphNode["type"];
     const name = match[2];
-    const propsBlock = match[3];
+    const propsBlock = match[3] ?? "";
     const properties: Record<string, string> = {};
 
     const propRegex = /<property\s+key="([^"]+)">([^<]*)<\/property>/g;
--- a/test/graph.test.ts
+++ b/test/graph.test.ts
@@ -106,6 +106,32 @@ describe("Graph Functions", () => {
     expect(edges[0].type).toBe("uses");
   });
 
+  it("graph-extract accepts self-closing entity tags", async () => {
+    mockProvider.compress.mockResolvedValueOnce(`<entities>
+<entity type="file" name="src/index.ts"/>
+<entity type="function" name="main"><property key="lang">typescript</property></entity>
+</entities>
+<relationships>
+<relationship type="uses" source="src/index.ts" target="main" weight="0.9"/>
+</relationships>`);
+
+    const result = (await sdk.trigger("mem::graph-extract", {
+      observations: [testObs],
+    })) as { success: boolean; nodesAdded: number; edgesAdded: number };
+
+    expect(result.success).toBe(true);
+    expect(result.nodesAdded).toBe(2);
+    expect(result.edgesAdded).toBe(1);
+
+    const nodes = await kv.list<GraphNode>("mem:graph:nodes");
+    expect(nodes.some((n) => n.name === "src/index.ts")).toBe(true);
+    expect(nodes.some((n) => n.name === "main")).toBe(true);
+
+    const edges = await kv.list<GraphEdge>("mem:graph:edges");
+    expect(edges).toHaveLength(1);
+    expect(edges[0].type).toBe("uses");
+  });
+
   it("graph-query with search returns matching nodes", async () => {
     await sdk.trigger("mem::graph-extract", { observations: [testObs] });
 


rohitg00:
Caught a regression on local test merge. The new self-closing test case fails because the regex still has a greedy quantifier:

```ts
// shipped in this PR:
/<entity\s+type="([^"]+)"\s+name="([^"]+)"[^>]*(?:\/>|>([\s\S]*?)<\/entity>)/g
//                                            ^^^^^ greedy — eats across the next entity
```

Traced standalone with the test payload (`<entity ... />\n<entity ... ><property/></entity>`): match 1 captures the first entity's name correctly, then `[^>]*` greedily consumes the `/` of `/>` and tries the alternation. The first alternative `\/>` fails (no `/` left to match), so it falls through to the second alternative `>(...)<\/entity>` — which then matches `>\n<entity ...><property/></entity>`, swallowing entity #2 entirely. Result: 1 match instead of 2.

**One-char fix**: lazy quantifier instead of greedy.

```diff
- /<entity\s+type="([^"]+)"\s+name="([^"]+)"[^>]*(?:\/>|>([\s\S]*?)<\/entity>)/g
+ /<entity\s+type="([^"]+)"\s+name="([^"]+)"[^>]*?(?:\/>|>([\s\S]*?)<\/entity>)/g
```

Verified with the same node trace: lazy returns 2 matches.

Push that and the test suite should go green. Thanks for the catch + the regression test — that test is exactly the kind of payload that exposes the bug.

jonathanzhan1975:
I hit this same parser failure while investigating #559 / graph rebuild behavior on a local v0.9.20-style store. This PR is the right fix direction, and I think the real provider output makes the case stronger.

A live graph extraction response that source `parseGraphXml` would parse as `0 nodes / 0 edges` was not empty. It looked like this, sanitized:

```xml
<entities>
  <entity type="agent" name="Agent"/>
  <entity type="system" name="Memory System"/>
  <entity type="component" name="MCP Server Agent Memory"/>
  <entity type="concept" name="Chinese Input"/>
  <entity type="concept" name="English Memories"/>
  <entity type="tool" name="Glob"/>
  <entity type="directory" name="<local-project-dir>"/>
  <entity type="file_type" name="Markdown files"/>
</entities>
<relationships>
  <relationship type="uses" source="Agent" target="Glob" weight="0.9"/>
  <relationship type="searches_in" source="Glob" target="<local-project-dir>" weight="0.9"/>
  <relationship type="searches_for" source="Glob" target="Markdown files" weight="0.9"/>
</relationships>
```

So self-closing entity support is not just a synthetic edge case: it is enough to turn real non-empty provider output into `0/0` today.

One follow-up this PR may or may not want to include: the same real response also contains out-of-schema entity/relationship types (`agent`, `system`, `component`, `directory`, `file_type`, `searches_in`, `searches_for`, etc.). Even after the self-closing regex fix, those need either normalization to the supported graph enums or explicit rejection/fallback, otherwise graph extraction may stay brittle. Happy to split that into a separate PR if you'd rather keep #494 narrowly scoped to self-closing tags.

This also cross-links to #559, because parser loss can present as an empty graph even when the provider returned usable graph-shaped XML.

rohitg00:
Thanks @Rex57 — regex change is minimal (alternation between self-closing and explicit-close) and `propsBlock ?? ""` guard handles the missing capture group. Tests cover the new path. Merged. Closes #492; partial fix for #559.

PR #495, opened 2026-05-18
fix: end Codex sessions on Stop hook

Rex57:
## Summary

- Wire Codex `Stop` to run the existing `session-end.mjs` hook after `stop.mjs`.
- Preserve the current best-effort summarize behavior while closing the session lifecycle for Codex, which does not use a separate `SessionEnd` entry in `hooks.codex.json`.
- Add regression coverage that the Codex Stop hook includes both summarize and session-end commands.

Fixes #493.
Related: #308.

## Verification

- `npx tsdown` passed in isolated fork clone.
- `node -e "JSON.parse(require('fs').readFileSync('plugin/hooks/hooks.codex.json','utf8')); console.log('hooks.codex.json OK')"` passed.
- Tests not run locally per local environment constraint; full `npm test` intentionally skipped to avoid any accidental interaction with the live AgentMemory setup.

## Summary by CodeRabbit

* **Chores**
  * Enhanced session cleanup process configuration.

* **Tests**
  * Added verification tests for session cleanup functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/495?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

Issue #493, opened 2026-05-18
Codex Stop hook does not mark sessions ended


Rex57:
## Bug

Codex plugin sessions can stay active in the AgentMemory viewer after Codex stops.

## Environment

- OS: Windows
- Node: v24.14.1
- agentmemory: 0.9.18
- Agent: Codex CLI plugin installed through `/plugin`

## Repro

1. Start `agentmemory`.
2. Start a Codex session with the AgentMemory plugin enabled.
3. End the Codex session.
4. Open the AgentMemory viewer.

## Expected

The session should be marked ended automatically.

## Actual

The session can remain active until manually ending it in the viewer.

## Root cause

`plugin/hooks/hooks.codex.json` wires Codex `Stop` to `plugin/scripts/stop.mjs`. That script only calls `/agentmemory/summarize`; it does not call `/agentmemory/session/end`.

`plugin/scripts/session-end.mjs` already has session-end behavior, but Codex does not expose/use a separate `SessionEnd` hook in `hooks.codex.json`.

## Suggested fix

Have the Stop hook end the session after best-effort summarize. This keeps existing summary behavior while closing the viewer session lifecycle for Codex.

Related: #308 covers broader abandoned-session recovery. This issue is the narrower Codex Stop-hook lifecycle path.

rohitg00:
Re-opening — the Stop-hook chain shipped in v0.9.19 (#495) was reverted via #501 because Codex Stop fires before the conversation actually finishes, marking sessions completed too early. Separate SessionEnd surface is what we want here. Tracking for v0.9.20+.

PR #501, opened 2026-05-18
revert: remove Codex Stop hook session-end

Rex57:
## Summary
- Reverts #495 because Codex Stop can fire before the overall conversation is truly finished.
- Restores the Codex Stop hook to summarize-only behavior.
- Removes the test that required session-end.mjs to run from Stop.

## Why
After testing in Codex, adding session-end.mjs to Stop marks sessions completed too early while later observations can still arrive. A separate SessionEnd hook is not surfaced by Codex /hooks for this plugin, so the safe behavior is to avoid ending sessions from Stop.

## Verification
- npx vitest run test/codex-plugin.test.ts

Refs #493

## Summary by CodeRabbit

* **Chores**
  * Simplified Stop hook by removing session-end command execution

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/501?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/plugin/hooks/hooks.codex.json
+++ b/plugin/hooks/hooks.codex.json
@@ -59,10 +59,6 @@
           {
             "type": "command",
             "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/stop.mjs"
-          },
-          {
-            "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/session-end.mjs"
           }
         ]
       }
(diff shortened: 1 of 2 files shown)


rohitg00:
Merged + shipping in v0.9.20 — https://github.com/rohitg00/agentmemory

Thanks @Rex57 for both the original fix and the field-tested revert. Re-opened #493 to track the SessionEnd-shaped solution. #502 follow-up on graph extraction also in the queue.

PR #502, opened 2026-05-18
fix: run graph extraction on manual session end

Rex57:
## Summary
- Runs the stopped-session processing pipeline when /agentmemory/session/end is called manually.
- Keeps the session completion update, but wraps manual end in a keyed lock so repeated manual end calls do not trigger graph extraction multiple times.
- Adds a regression test proving manual session end reaches mem::graph-extract once when graph extraction is enabled.

## Why
The viewer's manual End Session action calls /agentmemory/session/end. Before this change, that endpoint only marked the session completed; it did not run vent::session::stopped, so graph extraction was skipped until users manually backfilled.

## Verification
- RED: 
px vitest run test/api-session-end.test.ts failed before the fix with xpected [] to have a length of 1 for mem::graph-extract.
- GREEN: 
px vitest run test/api-session-end.test.ts
- Focused: 
px vitest run test/api-session-end.test.ts test/graph.test.ts test/evict.test.ts test/codex-plugin.test.ts — 17 tests passed.
- Build: 
px tsdown
- Note: 
px tsc --noEmit still fails on existing repo-wide type config/issues such as missing Node globals/types; not introduced by this change.

### DIFF
--- a/src/triggers/api.ts
+++ b/src/triggers/api.ts
@@ -570,10 +570,19 @@ export function registerApiTriggers(
           body: { error: "sessionId is required and must be a non-empty string" },
         };
       }
-      await kv.update(KV.sessions, sessionId, [
-        { type: "set", path: "endedAt", value: new Date().toISOString() },
-        { type: "set", path: "status", value: "completed" },
-      ]);
+      await withKeyedLock(`session:end:${sessionId}`, async () => {
+        const session = await kv.get<Session>(KV.sessions, sessionId);
+        if (session?.status !== "completed") {
+          await sdk.trigger({
+            function_id: "event::session::stopped",
+            payload: { sessionId },
+          });
+        }
+        await kv.update(KV.sessions, sessionId, [
+          { type: "set", path: "endedAt", value: new Date().toISOString() },
+          { type: "set", path: "status", value: "completed" },
+        ]);
+      });
       return { status_code: 200, body: { success: true } };
     },
   );
--- a/src/triggers/events.ts
+++ b/src/triggers/events.ts
@@ -63,10 +63,13 @@ export function registerEventTriggers(sdk: ISdk, kv: StateKV): void {
         );
         const compressed = observations.filter((o) => o.title);
         if (compressed.length > 0) {
-          sdk.triggerVoid("mem::graph-extract", { observations: compressed });
+          await sdk.trigger({
+            function_id: "mem::graph-extract",
+            payload: { observations: compressed },
+          });
         }
       } catch (err) {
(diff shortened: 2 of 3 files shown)


rohitg00:
Hey @Rex57 — holding to scope this against the minimal-surface rule.

The PR adds a new trigger path (manual session end fires graph extraction). Existing paths already extract on auto session-end via `event::session::stopped` (added in v0.9.4). Adding manual end as a second trigger is feature-additive.

Two questions:

1. **Which user is blocked today?** A tracked issue with the specific failure mode (manual end leaves graph empty) would confirm this is a real gap, not theoretical.
2. **Does the auto path miss your case for a reason that's reproducible?** If `event::session::stopped` fires but doesn't extract, that's a bug fix and merges easily. If it doesn't fire on manual end, the right fix might be to make the event fire reliably on both paths instead of adding a second trigger.

I like that the lock + status guard prevents double-extraction (catches the same failure mode that bit your #495 Stop chain). That's careful. Just want to make sure the surface addition matches a real need first.

(Not blocking — please link the user issue and I'll merge.)

Rex57:
User-facing repro for this PR:

Codex plugin currently does not have a SessionEnd hook in `plugin/hooks/hooks.codex.json`; it only wires `Stop -> stop.mjs`. Because of that, Codex sessions can remain active in the viewer, so the user has to open the AgentMemory viewer and manually click End Session.

Current manual End Session endpoint only marks the session completed (`endedAt`, `status`) and does not fire `event::session::stopped`. Since graph extraction is attached to the stopped-session event path, manual End Session leaves graph extraction unrun and the graph stays stale/empty.

So this is not a theoretical second trigger: it is the only reachable session-end path for the Codex viewer workflow after the Stop-hook session-end approach was reverted.

allandelmare:
+1 — adding a Claude Code use case for the maintainer's question about who's blocked.

Real-world impact on Claude Code (not just Codex): with the Claude Code plugin's `SessionEnd` hook, `/agentmemory/session/end` is what gets called when a session truly ends. In our testing on v0.9.21, this endpoint only marked the session completed; `event::session::stopped` never fired, so graph extraction stayed empty across 676 imported sessions and weeks of live work — exactly the symptom this PR fixes.

We built a workaround for this — a LaunchAgent that runs hourly, fetches recent sessions, and POSTs each one's compressed observations to `/agentmemory/graph/extract` directly. It works (got us to ~75K nodes / ~6K edges across 676 sessions) but shouldn't be necessary. This PR would obsolete our workaround and benefit Claude Code users too, not just Codex.

PR #503, opened 2026-05-18
fix(embed): allow separate OPENAI_EMBEDDING_BASE_URL + OPENAI_EMBEDDING_API_KEY

efenex:
## Summary

The OpenAI-compat embedding provider read `OPENAI_BASE_URL` and `OPENAI_API_KEY` only, which the chat-LLM path also reads — coupling both calls to the same endpoint. Operators who want chat on a fast hosted provider and embeddings on a self-hosted GPU box (or vice versa) had no way to split them.

Add two embedding-scoped overrides with fallback to the existing vars:

```
OPENAI_EMBEDDING_BASE_URL → falls back to OPENAI_BASE_URL → default
OPENAI_EMBEDDING_API_KEY  → falls back to OPENAI_API_KEY    → required
```

Existing setups continue working without changes (both fall back to the old vars).

## Why

Personal motivating example:

- I run DeepSeek V4 Flash on Novita for chat (cheap, fast)
- I run Qwen3-Embedding-8B on a dedicated GPU box via vLLM for embeddings (free, unlimited throughput, batchable)

Previously these had to live on one endpoint. Now they don't:

```bash
OPENAI_BASE_URL=https://api.novita.ai/v3/openai
OPENAI_API_KEY=sk-novita-...
OPENAI_EMBEDDING_BASE_URL=https://embed.example.lan
OPENAI_EMBEDDING_API_KEY=local-no-auth
OPENAI_EMBEDDING_MODEL=Qwen3-Embedding-8B
OPENAI_EMBEDDING_DIMENSIONS=4096
```

Other common patterns this enables:

- **Local Ollama for embeddings, remote OpenAI for chat** — `OPENAI_EMBEDDING_BASE_URL=http://localhost:11434 OPENAI_EMBEDDING_API_KEY=ollama`
- **LM Studio for embeddings + a remote chat** — same shape
- **Direct OpenAI embeddings + a cheaper proxy for chat** — set `OPENAI_EMBEDDING_BASE_URL=https://api.openai.com` and put the proxy URL in `OPENAI_BASE_URL`

## Why the separate API key

Most local endpoints (Ollama / LM Studio / llama.cpp / vLLM) ignore Authorization entirely, but Node fetch still requires a non-empty Bearer token. Setting `OPENAI_EMBEDDING_API_KEY=anything-truthy` unblocks that case without revealing the real `OPENAI_API_KEY` to whatever's on localhost.

## Test plan

Reviewed against the 17 existing test cases in `test/embedding-provider.test.ts`; no regression. The 6 pre-existing failures on `main` are env-pollution tests that fail when `~/.agentmemory/.env` has API keys set — verified unrelated to this PR by re-running on `main`.

Verified live by pointing embeddings at a local vLLM instance while keeping chat on Novita — `curl http://localhost:3111/agentmemory/lessons` save triggered embed calls visible in vLLM's `/metrics` endpoint, while chat-LLM (summarize, compress) continued hitting Novita.

## Files

- `src/providers/embedding/openai.ts` — 36 insertions, 7 deletions (mostly docstring expansion)

## Summary by CodeRabbit

* **New Features**
  * Embedding requests can now use a separate API key and base URL, letting embeddings target a different endpoint than other API calls.
* **Bug Fixes**
  * Clearer error messaging when no embedding credentials are configured, indicating which credential sources were checked.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/503?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/providers/embedding/openai.ts
+++ b/src/providers/embedding/openai.ts
@@ -48,16 +48,30 @@ function resolveDimensions(model: string, override: string | undefined): number 
  * `api-key` header instead of `Authorization: Bearer`.
  *
  * Required env vars:
- *   OPENAI_API_KEY            — API key
+ *   OPENAI_API_KEY               — API key (fallback for OPENAI_EMBEDDING_API_KEY)
  *
  * Optional:
- *   OPENAI_BASE_URL           — base URL without path (default: https://api.openai.com).
- *                               Azure: https://<resource>.openai.azure.com/openai/deployments/<deployment>
- *   OPENAI_API_VERSION        — Azure api-version query param (default: 2024-08-01-preview)
- *   OPENAI_EMBEDDING_MODEL    — model name (default: text-embedding-3-small)
- *   OPENAI_EMBEDDING_DIMENSIONS — override reported dimensions (required for
- *                                 custom / self-hosted models not in the
- *                                 MODEL_DIMENSIONS table above)
+ *   OPENAI_BASE_URL              — base URL without path (default: https://api.openai.com).
+ *                                  Azure: https://<resource>.openai.azure.com/openai/deployments/<deployment>
+ *   OPENAI_EMBEDDING_BASE_URL    — embedding-specific base URL override (defaults
+ *                                  to OPENAI_BASE_URL). Lets operators run
+ *                                  embeddings on a separate endpoint from chat —
+ *                                  e.g. local Ollama / LM Studio / llama.cpp /
+ *                                  vLLM at http://localhost:1234 for unlimited
+ *                                  free embeddings, while keeping chat
+ *                                  completions on a rate-limited but high-quality
+ *                                  hosted provider. Azure detection runs on
+ *                                  whichever URL ends up selected.
+ *   OPENAI_EMBEDDING_API_KEY     — separate API key for the embedding endpoint
+ *                                  (defaults to OPENAI_API_KEY). Useful when the
+ *                                  embedding endpoint requires a different key
+ *                                  or no key at all (set to e.g. "local" for
+ *                                  endpoints that ignore Authorization).
+ *   OPENAI_API_VERSION           — Azure api-version query param (default: 2024-08-01-preview)
+ *   OPENAI_EMBEDDING_MODEL       — model name (default: text-embedding-3-small)
+ *   OPENAI_EMBEDDING_DIMENSIONS  — override reported dimensions (required for
+ *                                  custom / self-hosted models not in the
+ *                                  MODEL_DIMENSIONS table above)
  */
 export class OpenAIEmbeddingProvider implements EmbeddingProvider {
   readonly name = "openai";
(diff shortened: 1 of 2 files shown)


efenex:
Rebased onto current `main` (now contains #462's OpenAI transport refactor) and addressed the CodeRabbit nit in [45f8e19](https://github.com/rohitg00/agentmemory/commit/45f8e19):

- **Rebase:** the embedding-specific URL/key fallback now layers on top of the shared `_openai-shared` module. `this.baseUrl = normalizeBaseUrl(getEnvVar("OPENAI_EMBEDDING_BASE_URL") || getEnvVar("OPENAI_BASE_URL"))`, so Azure auto-detection via `detectAzure(this.baseUrl)` automatically applies when the selected URL is an `*.openai.azure.com` host. Operators get the embedding endpoint split *and* Azure routing for free.
- **Error message nit:** `if (!this.apiKey)` now throws `API key is required (via constructor, OPENAI_EMBEDDING_API_KEY, or OPENAI_API_KEY)` — no more misleading single-source mention.
- Updated the matching test to use a regex matching all three sources.
- Merged the docstring: kept the Azure / `OPENAI_API_VERSION` section from the refactor and the embedding-specific env-var section from this PR.

Tests pass locally with 17/17 in `embedding-provider.test.ts` (after isolating from my own `~/.agentmemory/.env`).

rohitg00:
Hey @efenex — sitting on this to confirm scope. The PR adds **two new env vars** (`OPENAI_EMBEDDING_BASE_URL` + `OPENAI_EMBEDDING_API_KEY`) on top of the existing `OPENAI_BASE_URL` / `OPENAI_API_KEY` pair. New env vars are net-additive surface and we're trying to keep the agentmemory config table minimal.

To gate the addition: which concrete user is blocked today by the single-pair shape? E.g. "LLM on DeepSeek + embeddings on local Ollama" is a real ask, and #503 does unblock it. But if there's no tracked issue documenting a real user hitting this, the override might fit better as a downstream wrapper than as in-tree config.

Could you (a) link to a user issue describing the blocker, or (b) describe the deployment shape that needs the split? Once that's clear I'll either merge as-is or close in favor of a downstream pattern.

(Holding here, not blocking — the diff itself is clean and uses the existing `normalizeBaseUrl` helper correctly.)

efenex:
@rohitg00 thanks for sitting on this — the gate question is fair.

**Concrete deployment that this unblocks:**

The exact split I'm running locally — agentmemory daemon driving a serverless OpenAI-compat chat completions endpoint (novita.ai's hosted Qwen3.6) while embeddings stay on a separate vLLM at `https://embed1-openai.ai.pza.be` serving Qwen3-Embedding-8B with no auth token. The chat side is rate-limited and paid; the embedding side is unmetered and free. Without #503, the two pairs (`OPENAI_BASE_URL`/`OPENAI_API_KEY`) are shared — you can't point them at different endpoints, you can't run embeddings keyless against a local server while chat needs a paid API key, and you certainly can't have one side on Azure and the other on standard OpenAI shape.

It surfaced when I wrote `scripts/migrate-to-serverless-embedding.sh` (in a separate operator-script PR that I didn't open) — the migration script wanted to flip `embedding` off the vLLM onto a serverless provider while leaving `OPENAI_BASE_URL` pointed at the local Qwen3.6 daemon. I worked around it with `NODE_TLS_REJECT_UNAUTHORIZED=0` hackery; that's exactly the kind of operator scar tissue that gets reduced by the explicit override.

**Pattern is recurring upstream too:**

- #467 (honor2030, open) — `OPENAI_EMBEDDING_PATH` for the same problem space, different layer (custom path on the SAME base URL). The two PRs are complementary; #503 splits the host, #467 splits the path.
- The mirror pattern exists for LLM-side too: some users want chat on Anthropic but embeddings on OpenAI without it being a full provider-swap.

**If "downstream wrapper" is the policy, I'm fine closing:**

The whole thing can be achieved with a 5-line shell wrapper that twiddles env vars before spawning agentmemory, which is what I'd do otherwise. Just wanted to flag that this exact split has come up in two open PRs from two contributors in a few weeks — feels like the kind of operator knob that wants to live in-tree.

Either way, the PR's diff is independent of that policy call. Tell me which way you'd like it and I'll move accordingly.

rohitg00 (approved):
Audited locally: build clean, full suite passes (1081 + the 1 added test). Backward compat preserved — both env vars fall back to chat-side defaults so existing setups don't break. Ready to merge.

PR #504, opened 2026-05-18
fix(rebuild): batch embed calls in rebuildIndex (25h → 3h on large corpora)

efenex:
## Summary

`rebuildIndex` called `await vectorIndexAddGuarded(...)` per memory and per observation — each is one HTTP round-trip embedding a single input. On a real bulk-imported corpus + any non-zero network latency, the per-item serialization dominates wallclock.

Add a batched helper `vectorIndexAddBatchGuarded()` that calls `provider.embedBatch()` once for a buffered group of items, and refactor `rebuildIndex` to accumulate + flush in batches of `REBUILD_EMBED_BATCH_SIZE` (default 32).

## The numbers

Measured against a self-hosted vLLM (Qwen3-Embedding-8B) endpoint:

| call shape | latency | per-item |
|---|---|---|
| single embed | 175 ms | 175 ms |
| batch of 32 | 737 ms | **23 ms** |

That's **~7.6× speedup** per item. For a 500k-observation corpus, projected rebuild time drops from **~25 hours** to **~3 hours**.

The same benefit applies to any batchable OpenAI-compat endpoint — vLLM, Triton, OpenAI's own `/v1/embeddings`, LM Studio, llama.cpp server, etc. All accept an `input` array; their providers already amortize network + GPU setup across the batch.

## Companion to #500

#500 made `rebuildIndex` non-blocking so the viewer + later boot steps run immediately. But the rebuild itself still took the same wallclock. This PR cuts the rebuild itself. After both, boot is fast AND the index hot-loads in a small multiple of the time it takes the provider to physically embed the corpus.

## Failure semantics

| failure | before | after |
|---|---|---|
| network/provider error on the call | per-item warn × N | single warn for the batch, N items marked failed |
| dimension mismatch on one record | per-item warn | per-item warn (item skipped, others in batch continue) |
| `rebuildIndex` return value | count of attempted items | unchanged |

The soft-fail surface (call always returns, embed errors don't abort the rebuild) is preserved item-for-item.

## Configuration

```
# default
REBUILD_EMBED_BATCH_SIZE=32

# small endpoints that prefer fewer items per call
REBUILD_EMBED_BATCH_SIZE=8

# fall back to per-item path (effectively pre-PR behavior)
REBUILD_EMBED_BATCH_SIZE=1
```

## Test plan

`npx vitest run test/search-index.test.ts test/vector-index*.test.ts test/remember-bm25-index.test.ts` — 39/39 pass unchanged.

No new test added for the batching path itself (would need mocking the provider's `embedBatch` and asserting buffered flush at boundaries — possible but not done in this PR; the existing tests cover that `rebuildIndex` produces the same final index, and the batched helper is a transparent optimization). Happy to add if reviewer prefers.

Verified live against a real corpus + vLLM endpoint by counting vLLM's `vllm:request_success_total` before/after a rebuild — request count drops by a factor of ~32 as expected.

## Files

- `src/functions/search.ts` — new `vectorIndexAddBatchGuarded`, refactor `rebuildIndex` to use it; +112 lines, -9 lines

## Summary by CodeRabbit

* **New Features**
  * Batched embedding processing to speed up index rebuilds.

* **Improvements**
  * Granular success/failure tracking that skips problematic items without failing entire batches.
  * Configurable rebuild batch size (default 32) for optimized memory/observation indexing.
  * Faster handling when no sessions/data to process, avoiding unnecessary work.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/504?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/functions/search.ts
+++ b/src/functions/search.ts
@@ -86,6 +86,90 @@ export async function vectorIndexAddGuarded(
   }
 }
 
+// Batched variant: calls EmbeddingProvider.embedBatch ONCE for the whole
+// batch, then writes each resulting vector. Use this for bulk paths
+// (rebuildIndex, future bulk-add APIs) where per-item serial awaits
+// dominate wallclock. A batch of N has roughly the latency of a single
+// embed (network + GPU setup amortized), so backfilling a 500k-obs
+// corpus drops from days to hours on a per-batch endpoint like vLLM.
+//
+// Per-item failure shape:
+//   - whole-batch network/provider error → all skipped, single warn line
+//   - per-item dimension mismatch → that item skipped, others continue
+export async function vectorIndexAddBatchGuarded(
+  items: Array<{
+    id: string
+    sessionId: string
+    text: string
+    context: { kind: "memory" | "observation" | "synthetic"; logId: string }
+  }>,
+): Promise<{ ok: number; fail: number }> {
+  const vi = vectorIndex
+  const ep = currentEmbeddingProvider
+  if (!vi || !ep || items.length === 0) return { ok: 0, fail: 0 }
+
+  let embeddings: Float32Array[]
+  try {
+    embeddings = await ep.embedBatch(items.map((i) => clipEmbedInput(i.text)))
+  } catch (err) {
+    logger.warn("vector-index add batch: embed failed — skipping batch", {
+      batchSize: items.length,
+      provider: ep.name,
+      error: err instanceof Error ? err.message : String(err),
+    })
+    return { ok: 0, fail: items.length }
+  }
+
+  if (embeddings.length !== items.length) {
+    logger.warn(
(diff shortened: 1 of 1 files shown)


efenex:
Thanks for the catch — pushed [81c387a](https://github.com/rohitg00/agentmemory/commit/81c387a) wrapping the per-item `vi.add(...)` in try/catch inside the batch loop, mirroring your suggested diff. A single failing write now logs + increments `fail` and the loop continues, matching the pre-batch soft-fail semantics. The existing dimension-mismatch branch is untouched.

rohitg00:
Merged + shipping in v0.9.21. Thanks @efenex — 25h→3h is huge for users with large corpora. The batch-flush + REBUILD_EMBED_BATCH_SIZE env knob shape is clean.

PR #516, opened 2026-05-18
fix(mcp): memory_recall hits the right endpoint and forwards format/token_budget (#507)

### LINKED ISSUE #507: bug: memory_recall always returns compact mode — wrong endpoint + format param not forwarded in standalone.mjs
## Bug

`memory_recall` with `format="full"` always returns compact results, ignoring the format parameter.

## Root cause (two issues in `dist/standalone.mjs`)

**1. Wrong endpoint for `memory_recall`**

`memory_recall` and `memory_smart_search` are routed to the same endpoint:

```js
// standalone.mjs handleProxy()
case "memory_recall":
case "memory_smart_search": return textResponse(await handle.call("/agentmemory/smart-search", { ... }))
```

`/agentmemory/smart-search` does **not** support the `format` parameter and always returns compact. The correct endpoint for `memory_recall` is `/agentmemory/search`, which does support `format`.

**2. `format` parameter not included in request body**

Even after fixing the endpoint, `format` is never parsed from args or forwarded to the server:

```js
// validateArgs() — format never parsed
v.query = query.trim();
v.limit = parseLimit(args["limit"]);
// format missing

// handleProxy() — format never sent
body: JSON.stringify({
  query: v.query,
  limit: v.limit
  // format missing
})
```

## Fix

```js
// validateArgs()
v.format = typeof args["format"] === "string" ? args["format"].trim().toLowerCase() : "full";
v.token_budget = args["token_budget"];

// handleProxy() — split the cases
case "memory_recall": return textResponse(await handle.call("/agentmemory/search", {
  method: "POST",
  body: JSON.stringify({
    query: v.query,
    limit: v.limit,
    format: v.format ?? "full",
    ...(v.token_budget != null ? { token_budget: v.token_budget } : {})
  })
}), true);
case "memory_smart_search": return textResponse(await handle.call("/agentmemory/smart-search", {
  method: "POST",
  body: JSON.stringify({
    query: v.query,
    limit: v.limit,
    format: v.format ?? "full",
    ...(v.token_budget != null ? { token_budget: v.token_budget } : {})
  })
}), true);
```

## Verification

After applying the fix, `memory_recall` with `format="full"` correctly returns `facts`, `narrative`, `concepts`, and `files` from the server.

## Environment

- `@agentmemory/agentmemory`: 0.9.18
- `@agentmemory/mcp`: 0.9.18
- Server: running via systemd, `AGENTMEMORY_URL` set correctly
- MCP proxy mode (not local fallback)

serhiizghama:
## Problem

`memory_recall` in the standalone MCP shim always returns compact results, ignoring the `format` and `token_budget` parameters that the tool's input schema advertises (`src/mcp/tools-registry.ts`).

Two root causes in `src/mcp/standalone.ts`:

1. **Wrong endpoint.** `memory_recall` and `memory_smart_search` share a single `case` that routes both to `POST /agentmemory/smart-search`. That endpoint always returns compact mode and does not honor `format`. The endpoint that supports `format` (and exposes the full `facts` / `narrative` / `concepts` / `files` shape) is `POST /agentmemory/search`.
2. **Parameters dropped.** `validate()` never parses `format` or `token_budget`, and `handleProxy()` never includes them in the request body, so even after the endpoint is corrected the server has nothing to act on.

Fixes #507.

## Solution

- Parse `format` (string, lower-cased) and `token_budget` (positive integer, accepts string or number) in `validate()` for both `memory_recall` and `memory_smart_search`.
- Split the proxy case: `memory_recall` calls `POST /agentmemory/search` with `format` defaulted to `"full"` (matching the schema's documented default) and forwards `token_budget` when set. `memory_smart_search` keeps its existing `POST /agentmemory/smart-search` endpoint but now also forwards `format` and `token_budget` when the caller supplies them.
- Local `InMemoryKV` fallback behavior is unchanged.

## Testing

`npm test` — 1036 / 1036 passing locally.

Two new tests in `test/mcp-standalone-proxy.test.ts`:

- `proxies memory_recall to POST /agentmemory/search and forwards format/token_budget (#507)` — asserts the request URL, the body shape (`{query, limit, format, token_budget}`), and that `smart-search` is not called.
- `memory_recall defaults format to 'full' when omitted (#507)` — pins the default so the wire call matches the schema's documented default.

## New Features
* Memory tools now accept `format` and `token_budget` optional parameters for customized query behavior
* Memory recall defaults to `"full"` format when format parameter is unspecified
* Token budget parameter validates and accepts both numeric values and numeric string inputs

## Tests
* Added test coverage for memory tool proxy request formatting and parameter handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/516?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/mcp/standalone.ts
+++ b/src/mcp/standalone.ts
@@ -89,6 +89,8 @@ interface Validated {
   files?: string[];
   query?: string;
   limit?: number;
+  format?: string;
+  tokenBudget?: number;
   memoryIds?: string[];
   reason?: string;
 }
@@ -118,6 +120,17 @@ function validate(toolName: string, args: Record<string, unknown>): Validated {
       }
       v.query = query.trim();
       v.limit = parseLimit(args["limit"]);
+      const fmt = args["format"];
+      if (typeof fmt === "string" && fmt.trim()) {
+        v.format = fmt.trim().toLowerCase();
+      }
+      const budget = args["token_budget"];
+      if (typeof budget === "number" && Number.isFinite(budget) && budget > 0) {
+        v.tokenBudget = Math.floor(budget);
+      } else if (typeof budget === "string" && budget.trim()) {
+        const n = Number(budget);
+        if (Number.isFinite(n) && n > 0) v.tokenBudget = Math.floor(n);
+      }
       return v;
     }
     case "memory_sessions": {
@@ -159,11 +172,26 @@ async function handleProxy(
       });
       return textResponse(result);
     }
-    case "memory_recall":
+    case "memory_recall": {
+      const body: Record<string, unknown> = {
+        query: v.query,
+        limit: v.limit,
+        format: v.format ?? "full",
+      };
+      if (v.tokenBudget != null) body["token_budget"] = v.tokenBudget;
+      const result = await handle.call("/agentmemory/search", {
+        method: "POST",
+        body: JSON.stringify(body),
+      });
+      return textResponse(result, true);
+    }
     case "memory_smart_search": {
+      const body: Record<string, unknown> = { query: v.query, limit: v.limit };
+      if (v.format != null) body["format"] = v.format;
+      if (v.tokenBudget != null) body["token_budget"] = v.tokenBudget;
       const result = await handle.call("/agentmemory/smart-search", {
         method: "POST",
-        body: JSON.stringify({ query: v.query, limit: v.limit }),
+        body: JSON.stringify(body),
       });
       return textResponse(result, true);
     }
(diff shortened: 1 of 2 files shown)


rohitg00:
Merged + shipping in next release.

Thanks @serhiizghama — closes #507 and #440 (memory_recall returning compact mode + dropping format/token_budget). Both root causes addressed in a single PR: routes through /agentmemory/search instead of /smart-search, forwards format + token_budget end-to-end, defaults format to 'full' matching the schema.

PR #517, opened 2026-05-18
fix(viewer): prevent IME composition interruption in search inputs

jonathanzhan1975:
## Problem

The viewer's five search inputs — graph (`#graph-search`), memories (`#mem-search`), and the lessons/actions/crystals filters — destroy and recreate the focused input element via `innerHTML` on every keystroke. This interrupts any active IME composition session, making non-Latin input (Chinese pinyin, Japanese kana-kanji, Korean hangul) effectively impossible: the candidate window keeps closing before the user can commit a character.

While investigating I also noticed that the viewer's CSP includes:

```
script-src-attr 'none'
```

(see `src/viewer/document.ts` / `buildViewerCsp`). This silently blocks the inline `oninput="..."` handlers on the lessons/actions/crystals panels and the `onchange="..."` on the actions status `<select>`. Under the strict CSP, those three search controls and the status filter have been **non-functional** — the inline handlers never run. The user just sees a search box that does nothing.

## Fix

This PR adds two small helpers and migrates all five search inputs to use them. No framework, no render-layer refactor, no behavior change for ASCII users.

### `bindImeSafeSearch(input, ms, onSearch)`

Guards on **both** an explicit `compositionstart`/`compositionend` flag **and** `event.isComposing`, because engine-specific event ordering differs between Chromium/Firefox/Safari and you can't rely on either signal alone (per MDN [InputEvent.isComposing](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/isComposing) and the [compositionend spec](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionend_event)).

`compositionend` triggers an immediate commit (so the user sees the filter result without a 200ms delay after every IME selection). A one-shot `justCommitted` flag suppresses the redundant trailing `input` event that browsers dispatch after `compositionend`, avoiding a duplicate render.

### `captureSearchFocus(ids)` / `restoreSearchFocus(focus)`

Records `activeElement.id`, `selectionStart`, `selectionEnd` immediately before each `innerHTML = html`, then restores focus and cursor position after re-binding listeners. The guard checks `activeElement.id` is in the managed list before saving, so we never steal focus from anywhere else.

Without this, even after fixing the IME interrupt, multi-word Chinese input still required clicking back into the search box after every committed word.

### Migrated controls

All five search inputs and the actions status `<select>` are migrated to `addEventListener` via the helpers above, with a unified 200ms debounce.

| Panel | Element | Before | After |
|---|---|---|---|
| Graph | `#graph-search` | `addEventListener('input', debounce(..., 150))` | `bindImeSafeSearch(..., 200, ...)` + focus restore |
| Memories | `#mem-search` | `addEventListener('input', debounce(..., 200))` | `bindImeSafeSearch(..., 200, ...)` + focus restore |
| Lessons | `.search-input` → `#lessons-search` | inline `oninput=` (CSP-blocked) | `bindImeSafeSearch(..., 200, ...)` + focus restore |
| Actions | `.search-input` → `#actions-search` | inline `oninput=` (CSP-blocked) | `bindImeSafeSearch(..., 200, ...)` + focus restore |
| Actions filter | `<select>` → `#actions-status-filter` | inline `onchange=` (CSP-blocked) | `addEventListener('change', ...)` |
| Crystals | `.search-input` → `#crystals-search` | inline `oninput=` (CSP-blocked) | `bindImeSafeSearch(..., 200, ...)` + focus restore |

The patch carries an `// IME_SAFE_SEARCH_V2` marker comment for quick `grep` verification.

## Verification

**Synthetic regression** (Node + jsdom + synthetic `CompositionEvent`/`InputEvent`): 8/8 pass

- ASCII typing debounces and commits once
- IME composition: no commit during composing
- IME composition: immediate commit on compositionend
- `justCommitted` suppresses the trailing input event (single commit)
- Post-IME ASCII typing resumes debounced commits
- Composition cancel returns to idle
- Fast typing coalesces into a single commit

**Manual browser** (Windows 11 + Chrome): 8/8 pass

- Chinese pinyin commit works (memories, graph, lessons, actions, crystals)
- Multi-word Chinese input retains focus across commits — no click-back required
- English search debounces as before
- Cursor position preserved (`abc` → cursor between `a` and `b` → type `X` → result `aXbc`)
- DevTools Console shows **0** `script-src-attr` CSP violations after the fix (previously had violations on every lessons/actions/crystals interaction)
- Actions `<select>` change handler binds correctly

## What this PR explicitly does not do

- No framework introduction (no React/Vue/lit, etc.)
- No keyed reconciliation, no virtual DOM
- No changes to the render functions' overall structure (still `innerHTML`-based)
- No CSP changes — the CSP was already correct; the inline handlers were the regression

## Risk surface

- Focus restoration runs `el.focus()` after each managed render. This is gated on `activeElement.id` matching a managed input, so it can't steal focus from elsewhere on the page.
- `setSelectionRange` is wrapped in `try/catch` to defend against future input types that might reject it (e.g., `type="number"` doesn't support it).
- The `setTimeout(..., 0)` to clear `justCommitted` runs as a microtask after the next event-loop tick, which is the same tick where the trailing post-compositionend `input` event fires — so the suppression window is tight but reliable across the browsers I tested.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search functionality for composition-based input methods (IME), preventing accidental searches during text composition
  * Enhanced focus and selection preservation in search inputs when navigating between tabs

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/517?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/viewer/index.html
+++ b/src/viewer/index.html
@@ -1127,6 +1127,39 @@
       };
     }
 
+    // IME_SAFE_SEARCH_V2
+    function bindImeSafeSearch(input, ms, onSearch) {
+      var composing = false;
+      var justCommitted = false;
+      var run = debounce(function(value) { onSearch(value); }, ms);
+      input.addEventListener('compositionstart', function() { composing = true; });
+      input.addEventListener('compositionend', function() {
+        composing = false;
+        justCommitted = true;
+        onSearch(input.value);
+        setTimeout(function() { justCommitted = false; }, 0);
+      });
+      input.addEventListener('input', function(e) {
+        if (composing || e.isComposing) return;
+        if (justCommitted) return;
+        run(input.value);
+      });
+    }
+    function captureSearchFocus(ids) {
+      var a = document.activeElement;
+      if (!a || ids.indexOf(a.id) < 0) return null;
+      return { id: a.id, start: a.selectionStart, end: a.selectionEnd };
+    }
+    function restoreSearchFocus(focus) {
+      if (!focus) return;
+      var el = document.getElementById(focus.id);
+      if (!el) return;
+      el.focus();
+      if (typeof el.setSelectionRange === 'function') {
+        try { el.setSelectionRange(focus.start, focus.end); } catch (e) {}
+      }
+    }
+
     async function api(path, opts) {
       try {
         var url = REST + '/agentmemory/' + path;
@@ -1629,6 +1662,7 @@
 
       html += '<button class="btn" style="margin-top:14px;width:100%;font-size:11px;padding:8px;letter-spacing:0.06em;transition:all 0.15s ease;" data-action="rebuild-graph">↻ Rebuild Graph</button>';
       html += '<div id="selected-node-panel"></div>';
+      var __focus = captureSearchFocus(['graph-search']);
       sb.innerHTML = html;
 
       sb.querySelectorAll('input[type="checkbox"]').forEach(function(cb) {
@@ -1640,11 +1674,9 @@
 
       var searchInput = document.getElementById('graph-search');
       if (searchInput) {
-        searchInput.addEventListener('input', debounce(function() {
-          graphSearchTerm = this.value.toLowerCase();
-          renderGraph();
-        }, 150));
+        bindImeSafeSearch(searchInput, 200, function(v){ graphSearchTerm = v.toLowerCase(); renderGraph(); });
       }
+      restoreSearchFocus(__focus);
     }
 
     function initGraph() {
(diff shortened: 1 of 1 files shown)


rohitg00:
Merged + shipping in next release.

Thanks @jonathanzhan1975 for the IME composition guard — clean fix for CJK users hitting mid-character interruption on the viewer search inputs.

PR #560, opened 2026-05-20
fix(cli): accurately display bound viewer port on splash screen

### LINKED ISSUE #521: CLI splash claims `Viewer http://localhost:3113` before viewer is actually listening (v0.9.20)
### Version
`@agentmemory/agentmemory` `v0.9.20`

### Environment
- macOS Apple Silicon (Darwin 25.4.0)
- Node.js (installed via Homebrew)
- agentmemory installed via `npm install -g @agentmemory/agentmemory`
- Launched in tmux foreground from `~/.agentmemory`:
  ```bash
  tmux new-session -d -s agentmemory -c "$HOME/.agentmemory" agentmemory
  ```

### What happens
On a non-trivial fraction of restarts, the boxed startup splash prints:

```
│  Viewer       http://localhost:3113                            │
```

…but `lsof -iTCP:3113 -sTCP:LISTEN` immediately afterwards shows **nothing listening**. At the same moment:

- REST API on `3111` is healthy (`/agentmemory/health` returns `{"status":"healthy"}`)
- streams on `3112` are healthy
- iii engine WS on `49134` is healthy

Waiting 5-30 more seconds, `3113` eventually does come up — or it never does and a subsequent `am-restart` resolves it.

### Observed log sequence (tmux)

```
[iii] Worker registered with ID: <uuid>
◇  agentmemory v0.9.20
│  REST API     http://localhost:3111
│  Viewer       http://localhost:3113          ← printed here
│  Streams      ws://localhost:3112
│  Engine       ws://localhost:49134
Try: agentmemory demo
[agentmemory] info Observation captured ...
```

### Reproduction
1. Start agentmemory in tmux.
2. Restart it repeatedly with a fresh process group.
3. On some restarts, splash prints `Viewer http://localhost:3113`.
4. Immediately run:
   ```bash
   lsof -iTCP:3113 -sTCP:LISTEN
   ```
5. Nothing is listening on `3113`, even though the splash said it was.
6. `curl http://localhost:3111/agentmemory/health` still returns healthy.

### Root cause hypothesis

Independent source-trace analysis (across `dist/index.mjs` and `dist/cli.mjs`) suggests two bugs combine to produce this:

**1. Worker — `startViewerServer()` does not await bind (`dist/index.mjs` around line 19978)**

- `server.listen(currentPort, "127.0.0.1")` is fire-and-forget at ~line 20044
- Success is only signalled via the later `listening` event (~line 20047)
- `EADDRINUSE` errors are handled internally and retried up to `MAX_VIEWER_PORT_RETRIES = 10` (~line 20052), but the retry loop never surfaces failure or the actual bound port back to the caller
- `main()` (~line 20436) just calls `startViewerServer(...)` and continues immediately

**2. CLI — splash prints from config, never probes the viewer (`dist/cli.mjs`)**

- `waitForAgentmemoryReady()` only waits on REST `/agentmemory/livez` (around lines 1578, 2050)
- `printReadyHint()` then prints `Viewer ${getViewerUrl()}` unconditionally (around lines 1546, 2069, 2078)
- `getViewerUrl()` derives the URL from config (`REST_PORT + 2`), not from the worker's actual bound port
- There IS a `viewerReachable()` helper at ~line 2311 but readiness gating doesn't call it

So the splash can claim `Viewer http://localhost:3113` even when:
- the viewer's `server.listen()` has not yet emitted `listening`, or
- the viewer hit `EADDRINUSE` on `3113` and the worker silently retried up to `3114-3123` (the CLI never learns the actual bound port)

### Suggested minimal fix

`dist/cli.mjs` near line 2053 (inside `waitForAgentmemoryReady`):

```diff
- if (await isAgentmemoryReady()) return true;
+ if (await isAgentmemoryReady() && await fetch(getViewerUrl(), { signal: AbortSignal.timeout(500) }).then((r) => r.ok).catch(() => false)) return true;
```

This stops the false-success message at minimal cost. Limitation: if the viewer fell back to `3114+`, the splash will silently fail to ever return true, because `getViewerUrl()` still doesn't know the real port.

### Better fix
Have the worker report the *actual* bound viewer port (including any fallback after `EADDRINUSE`) back to the CLI on `/agentmemory/health` (or via stdout sync), and have the CLI use that port for both the splash and the reachability probe.

### Workaround in the meantime
Trust the REST API (`3111`) and probe `3113` directly with `lsof` / `curl` instead of trusting the splash.

---

Diagnosed read-only against the installed `0.9.20` bundle on macOS; no source modifications. Happy to PR the minimal one-liner if you'd like.

Tanmay-008:
- Expose viewerPort and viewerSkipped state in /agentmemory/livez endpoint.
- Update CLI readiness check to poll until the viewer port is bound or explicitly skipped.
- Prevents misleading default port (3113) display on splash screen when the viewer falls back to another port.

**Fixes #521**

## Summary by CodeRabbit

* **New Features**
  * CLI now auto-discovers the viewer port from the running server and uses it to build viewer URLs.
  * Health and liveness endpoints include viewer server status, port, and a "viewer skipped" indicator.
  * Readiness and diagnostics now honor discovered viewer port and treat skipped-viewer signals as ready, improving reachability checks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/560?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -195,9 +195,36 @@ function getBaseUrl(): string {
   return `http://localhost:${getRestPort()}`;
 }
 
+let discoveredViewerPort: number | null = null;
+
+export async function discoverViewerPort(): Promise<void> {
+  if (discoveredViewerPort !== null) return;
+  try {
+    const res = await fetch(`${getBaseUrl()}/agentmemory/livez`, {
+      signal: AbortSignal.timeout(1000),
+    });
+    if (res.ok) {
+      const data = await res.json() as { viewerPort?: number | null };
+      if (typeof data.viewerPort === "number") {
+        discoveredViewerPort = data.viewerPort;
+      }
+    }
+  } catch {}
+}
+
 function getViewerUrl(): string {
   const envUrl = process.env["AGENTMEMORY_VIEWER_URL"];
   if (envUrl) return envUrl.replace(/\/+$/, "");
+  
+  if (discoveredViewerPort !== null) {
+    try {
+      const u = new URL(getBaseUrl());
+      return `${u.protocol}//${u.hostname}:${discoveredViewerPort}`;
+    } catch {
+      return `http://localhost:${discoveredViewerPort}`;
+    }
+  }
+  
   try {
     const u = new URL(getBaseUrl());
     const vPort =
@@ -257,7 +284,18 @@ async function isAgentmemoryReady(): Promise<boolean> {
     const res = await fetch(`${getBaseUrl()}/agentmemory/livez`, {
       signal: AbortSignal.timeout(2000),
     });
-    return res.ok;
+    if (!res.ok) return false;
+    try {
+      const data = await res.json() as { viewerPort?: number | null; viewerSkipped?: boolean };
+      if (typeof data.viewerPort === "number") {
+        discoveredViewerPort = data.viewerPort;
+        return true;
+      }
+      if (data.viewerSkipped) return true;
+      return false;
+    } catch {
+      return false;
+    }
   } catch {
     return false;
   }
@@ -1101,6 +1139,9 @@ async function runStatus() {
       apiFetch<any>(base, "config/flags"),
     ]);
 
+    if (typeof healthRes?.viewerPort === "number") {
+      discoveredViewerPort = healthRes.viewerPort;
+    }
     const h = healthRes?.health;
     const status = healthRes?.status || "unknown";
     const version = healthRes?.version || "?";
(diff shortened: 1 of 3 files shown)


rohitg00:
Thanks @Tanmay-008 — exposing viewerPort + viewerSkipped on /livez is a clean way to make the splash truthful. Merged. Closes #521.

PR #561, opened 2026-05-20
fix: read tool_response instead of tool_output in PostToolUse hook

### LINKED ISSUE #539: [BUG] PostToolUse hook reads data.tool_output but Claude Code sends tool_response — compresses fail XML schema (~47% loss)
**Version:** v0.9.20
**Integration:** Claude Code plugin (`@agentmemory/agentmemory@0.9.20`)

## TL;DR
`scripts/post-tool-use.mjs` line 27 reads `data.tool_output`, but Claude Code's PostToolUse hook payload uses the field name `tool_response`. As a result, `cleanOutput` is `undefined`, the engine receives an observation with no `toolOutput`, the compress prompt contains no `Output:` section, and the model returns an XML response too sparse to pass `CompressOutputSchema` (which requires `narrative >= 10` chars + `facts >= 1`).

Effect: `mem::compress` silently fails on most real Claude Code tool calls (Bash, Read, Edit, Glob, Grep, ...). Observations still land in storage, but they have no compressed summary, so `memory_recall` titles/types are degraded and downstream consolidation has no substrate.

## Repro

1. Install agentmemory plugin in Claude Code via marketplace.
2. Set up provider (any LLM key works — bug is independent of provider).
3. Issue any normal tool call (e.g., a Bash `ls`).
4. Inspect `/agentmemory/health` → `functionMetrics[functionId='mem::compress']` shows failures.

## Captured evidence

Adding `appendFileSync` of raw stdin to `post-tool-use.mjs` captured this real PostToolUse payload from Claude Code for a Bash call:

```json
{
  \"session_id\": \"ee090c38-...\",
  \"hook_event_name\": \"PostToolUse\",
  \"tool_name\": \"Bash\",
  \"tool_input\": {\"command\": \"echo hi\"},
  \"tool_response\": {
    \"stdout\": \"hi\n\",
    \"stderr\": \"\",
    \"interrupted\": false,
    \"isImage\": false
  }
}
```

Note **`tool_response`**, not `tool_output`. Same field name for Edit, Read, etc. (with different inner shapes).

Source of the bug:

\`\`\`js
// post-tool-use.mjs line 27
const { imageData, cleanOutput } = extractImageData(data.tool_output);  // undefined
\`\`\`

## Patch

\`\`\`diff
-	const { imageData, cleanOutput } = extractImageData(data.tool_output);
+	const rawOutput = data.tool_response ?? data.tool_output;
+	const { imageData, cleanOutput } = extractImageData(rawOutput);
\`\`\`

The `?? data.tool_output` keeps backward compatibility for any non-Claude-Code consumer that does send `tool_output`.

## Metrics impact (measured on my install, Windows + Claude Code 4.7)

| Window | Sample | Success | Rate |
|---|---|---|---|
| Pre-patch cumulative (mixed: free-tier 429 + flash-lite + bug) | 91 | 48 | 53% |
| **Post-patch only (Bash + Read + Glob + Grep + Edit)** | **7** | **7** | **100%** |

Small sample but pattern is unambiguous: every compress call after the patch returned a fully populated XML observation that passed `CompressOutputSchema`. Before patch, the same kinds of calls produced sparse XML rejected for missing `<facts>` or short `<narrative>`.

## Related

- The `try { ... } catch {}` at the fetch site in `post-tool-use.mjs` matches the same pattern flagged in #522 — failures during observation submission are swallowed silently, making this kind of payload-schema mismatch hard to surface. A `console.error` on the catch (gated by `AGENTMEMORY_DEBUG`) would have made this bug a 1-minute discovery instead of a 1-hour root-cause hunt.
- I have not surveyed `pre-tool-use.mjs` or `post-tool-failure.mjs` for the same field-name issue. PreToolUse has no response payload yet, so it's likely unaffected. PostToolUseFailure may have a similar `error` vs `tool_response.error` mismatch — worth a glance.

Happy to PR if useful.

faraz152:
## Root cause

`src/hooks/post-tool-use.ts` (and its compiled counterpart `plugin/scripts/post-tool-use.mjs`) read the tool output from `data.tool_output`:

```ts
const { imageData, cleanOutput } = extractImageData(data.tool_output);
```

But Claude Code's PostToolUse hook payload uses `tool_response`, not `tool_output`. This means `cleanOutput` is always `undefined`, the observe request reaches the server with no output value, and `mem::compress` fails schema validation on every real tool call — `narrative` and `facts` are empty so the XML schema check fails.

Captured real PostToolUse payload from Claude Code (from #539):

```json
{
  "hook_event_name": "PostToolUse",
  "tool_name": "Bash",
  "tool_input": { "command": "echo hi" },
  "tool_response": { "stdout": "hi\n", "stderr": "", "interrupted": false }
}
```

## Fix

Read `data.tool_response` with `data.tool_output` as a fallback so any older integration still using the legacy field name keeps working:

```ts
const { imageData, cleanOutput } = extractImageData(
  data.tool_response ?? data.tool_output,
);
```

Same one-liner applied to the compiled `plugin/scripts/post-tool-use.mjs`.

## Impact

Without this fix `mem::compress` silently fails on ~100% of real Claude Code tool calls, degrading `memory_recall` output and starving the consolidation pipeline of compressed observations.

Fixes #539

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool response processing to prefer the newer response field when present, falling back to legacy output otherwise.
  * Ensures image data and cleaned output are correctly extracted and included in downstream observations, improving compatibility and reliability across response formats.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/561?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/plugin/scripts/post-tool-use.mjs
+++ b/plugin/scripts/post-tool-use.mjs
@@ -23,7 +23,7 @@ async function main() {
 	}
 	if (isSdkChildContext(data)) return;
 	const sessionId = data.session_id || "unknown";
-	const { imageData, cleanOutput } = extractImageData(data.tool_output);
+	const { imageData, cleanOutput } = extractImageData(data.tool_response ?? data.tool_output);
 	try {
 		await fetch(`${REST_URL}/agentmemory/observe`, {
 			method: "POST",
--- a/src/hooks/post-tool-use.ts
+++ b/src/hooks/post-tool-use.ts
@@ -32,7 +32,11 @@ async function main() {
 
   const sessionId = (data.session_id as string) || "unknown";
 
-  const { imageData, cleanOutput } = extractImageData(data.tool_output);
+  // Claude Code's PostToolUse payload uses `tool_response`; fall back to
+  // `tool_output` for older integrations that may use the legacy field name.
+  const { imageData, cleanOutput } = extractImageData(
+    data.tool_response ?? data.tool_output,
+  );
 
   try {
     await fetch(`${REST_URL}/agentmemory/observe`, {
--- a/src/hooks/post-tool-use.ts
+++ b/src/hooks/post-tool-use.ts
@@ -32,8 +32,6 @@ async function main() {
 
   const sessionId = (data.session_id as string) || "unknown";
 
-  // Claude Code's PostToolUse payload uses `tool_response`; fall back to
-  // `tool_output` for older integrations that may use the legacy field name.
   const { imageData, cleanOutput } = extractImageData(
     data.tool_response ?? data.tool_output,
   );


rohitg00:
Thanks @faraz152 — clean root-cause writeup with a captured payload, exactly the kind of repro that makes review easy. Merged.

PR #538, opened 2026-05-19
feat(graph): backfill graph from stored memories

jonathanzhan1975:
## Summary

Adds a real graph backfill path for existing AgentMemory data, so users with historical memories/observations can populate the viewer graph instead of seeing `0 nodes / 0 edges` forever.

This addresses the missing `/agentmemory/graph/build` path surfaced by the viewer rebuild flow and complements the existing session-end extraction path by adding historical backfill.

Fixes #505 — the viewer "Rebuild Graph" button now has a real backing endpoint instead of 404'ing. Confirmed by @code2tan to still be broken on v0.9.20.

## What changed

- Add `mem::graph-build` to scan stored observations and latest memories, plan/apply in batches, skip already-processed source IDs, and default to safe dry-run.
- Add `POST /agentmemory/graph/build` with query/body validation.
- Add `agentmemory graph-build` CLI, default dry-run; `--apply` writes graph records.
- Make CLI apply resumable in bounded chunks to avoid long HTTP request timeouts.
- Fix `mem::graph-extract` edge merging so edges point to persisted node IDs after duplicate node merges.
- Update viewer graph rebuild UX with dry-run preview, confirmation, chunked apply, and resume-friendly failure text.
- Replace Unix-only build asset copy commands with a cross-platform Node script.
- Add focused graph tests and a local implementation report.

## Verification

- `npm test -- test/graph.test.ts` -> 11 tests passed
- `npm run build` -> passed on Windows
- `git diff --check` -> no whitespace errors
- Local deployed dry-run before apply:
  - memories scanned: 157
  - source items found: 164
  - eligible items: 164
  - batches planned: 17
  - graph before: 0 nodes / 0 edges
- Local apply result after resumable chunks:
  - graph after: 395 nodes / 46 edges
  - follow-up dry-run: eligible items 0, skipped existing 164

## Notes

The viewer's empty-graph auto path remains read-only: it only performs a dry-run preview. A write happens only after explicit rebuild confirmation or CLI `--apply`.

## Summary by CodeRabbit

* **New Features**
  * Add graph-build REST endpoint and CLI with dry-run/apply, batching, per-chunk progress, and detailed build stats; viewer shows dry-run plan and guided “Rebuild Graph” flow.
* **Bug Fixes**
  * Fixes to graph extraction/edge remapping and XML entity parsing for more reliable graph imports.
* **Chores**
[...]

PR #568, opened 2026-05-20
feat(graph): auto-trigger mem::graph-extract on mem::remember

efenex:
## Summary

`mem::remember` (the explicit memory-save path used by `memory_save` MCP calls, plugin Stop hooks, JSONL imports, etc.) does NOT auto-trigger `mem::graph-extract`. Observations *do* auto-trigger extraction when `GRAPH_EXTRACTION_ENABLED=true`, but memories don't — so the knowledge graph grows from observation traffic only, leaving curated memories invisible to graph queries.

This wires the same auto-trigger pattern on the memory write path: after a successful `mem::remember`, fire `mem::graph-extract` against the newly-saved memory id when `GRAPH_EXTRACTION_ENABLED=true`. Best-effort; failure is logged-and-continued so a slow/down LLM provider can't block the save.

## Impact

Closes the gap between observation-driven and memory-driven graph growth. After this:

- `mem::lineage` (with `includeGraph: true`) sees the same node set whether a concept was extracted from observation traffic or written as a memory.
- `mem::smart-search` graph-score component lights up for curated memories.
- `agentmemory graph-build` (#538, if/when it lands) becomes a backfill tool for old data, not the only path for new data.

## Related

- Complements #502 (graph extraction on manual session end) — same auto-trigger pattern, different event.
- Independent of #538 (graph backfill) — that handles historical data, this handles new writes.

## Test plan

- [x] Existing tests pass (\`npm test\`)
- [x] \`GRAPH_EXTRACTION_ENABLED=true mem::remember\` produces graph nodes within the configured concurrency window
- [x] \`GRAPH_EXTRACTION_ENABLED=false\` (default) no-ops the trigger

## Summary by CodeRabbit

* **New Features**
  * Memory saves can now optionally trigger background knowledge-graph extraction when the feature flag is enabled. Extraction runs asynchronously after a save so the save remains fast; extraction errors are logged as warnings and do not block or revert the original save.

### DIFF
--- a/src/functions/remember.ts
+++ b/src/functions/remember.ts
@@ -7,7 +7,7 @@ import { memoryToObservation } from "../state/memory-utils.js";
 import { deleteAccessLog } from "./access-tracker.js";
 import { recordAudit } from "./audit.js";
 import { getSearchIndex, vectorIndexAddGuarded, vectorIndexRemove, flushIndexSave } from "./search.js";
-import { getAgentId } from "../config.js";
+import { isGraphExtractionEnabled, getAgentId } from "../config.js";
 import { logger } from "../logger.js";
 
 export function registerRememberFunction(sdk: ISdk, kv: StateKV): void {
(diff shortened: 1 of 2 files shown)


rohitg00 (approved):
Audited locally: build clean, tests pass. Closes #600 (knowledge graph never auto-fires). Gated on existing `GRAPH_EXTRACTION_ENABLED` flag, fire-and-forget so it never blocks mem::remember, logger-only on failure. Ready to merge.

efenex:
Rebased on current `main`. `sdk.triggerVoid` was removed in #773, so the graph-extract enqueue now uses `sdk.trigger` + `TriggerAction.Void()`. Tests green; only the Vercel deploy-preview is red (it fails on every fork PR). Ready when you are.

efenex:
Rebased onto current `main` (e04ba88). The only conflict was the `search.js` / `config.js` import line, since upstream added `isMemoryIndexReady`. Typecheck is unchanged vs. main. Tests: 1710 pass. The single failure (`cli-lifecycle-safety` › native removal shutdown) fails identically on `main`. Ready when you are.

efenex:

PR #569, opened 2026-05-20
feat(provider): AGENTMEMORY_DISABLE_THINKING to recover structured output on Qwen3/GLM/Kimi

efenex:
## Summary

Adds an opt-in env var \`AGENTMEMORY_DISABLE_THINKING=true\` that forces thinking mode OFF on hybrid-reasoning models served behind OpenAI-compatible endpoints (Qwen3 family, GLM, Kimi, DeepSeek V4-Flash, etc.).

## Motivation

When the configured \`OPENAI_BASE_URL\` points at a hybrid-reasoning model (e.g. Qwen3.6-35B-A3B-FP8 via vLLM, or a serverless provider hosting Qwen3-32B), every chat completion burns input tokens on a \`<think>...</think>\` block before the actual answer. Worse, structured-output prompts (graph extraction, XML/JSON-mode summarization, \`mem::compress\`, \`mem::summarize\`'s XML schema) frequently truncate inside the thinking block — \`message.content\` comes back empty and \`message.reasoning\` carries the half-finished meandering, which the existing parsers can't recover into a valid summary/graph entity.

I hit this concretely while bulk-rebuilding the graph (~8k nodes / ~5k edges across observations + memories) against a local Qwen3.6 vLLM, where ~30% of chunks failed to parse due to mid-thinking truncation. Same pattern documented in another OSS project's \`llm_engine.py:6207-6260\` as the "\$7 Qwen3-32B incident".

## What it does

Two signals, belt-and-suspenders:
1. **Server-side**: send \`chat_template_kwargs.enable_thinking=false\` in the chat completions request body. Honored by vLLM, SGLang, llama.cpp's OpenAI-compatible server, and most thinking-aware hosting layers.
2. **Client-side fallback**: prefix \`/no_think\` to the system message. This is the convention Qwen3 and friends recognize even when the server doesn't pass through \`chat_template_kwargs\`.

Both fire only when \`AGENTMEMORY_DISABLE_THINKING=true\` is set. Default off; existing setups with non-thinking models are unaffected.

## Test plan

- [x] \`npm test\` passes (existing OpenAI provider tests still green; no new tests added — the change is a body-shape additive)
- [x] With \`AGENTMEMORY_DISABLE_THINKING=true\` and a Qwen3 endpoint, \`mem::compress\` returns parseable XML on the first attempt instead of intermittently truncating mid-think
- [x] With the env var unset, request body is unchanged (verified by inspection)

## Related

- Sister to #467 / #503 / #467 — operator-side knobs for OpenAI-compatible deployments with non-standard transports or backends.

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to disable “thinking” behavior for hybrid AI models, enabling streamlined responses and improved structured-output handling.
  * Changed the default AI model used when no model is configured.

* **Documentation**
  * Added the new setting to the environment configuration example.
  * Updated the environment variable reference to include the new toggle and expanded list of recognized variables.


rohitg00 (approved):
Audited locally: build clean, tests pass. Off by default (`AGENTMEMORY_DISABLE_THINKING=true` opt-in). chat_template_kwargs + /no_think belt-and-suspenders is the documented vLLM-side pattern for Qwen3 / GLM / Kimi hybrid-reasoning models. Ready to merge.

cjx12138:
This change would be very helpful for my use case. I use DeepSeek with agentmemory, and every compression call currently consumes a significant number of reasoning tokens. Since compression is a straightforward summarization task, I don’t need thinking mode at all. Being able to disable it would substantially reduce token usage, cost, and latency. I’d really appreciate it if this PR could be merged.

efenex:
Rebased onto current `main` (e04ba88) and regenerated `REFERENCE.md` with `npm run skills:gen`. Added an `.env.example` entry and trimmed an inline comment, so this needs a quick re-look.

On the overlap with `AGENTMEMORY_LLM_NOTHINK`: that knob only appends `/no_think` to the graph-extraction prompt. This one applies to every OpenAI-provider call, and it also sends `chat_template_kwargs.enable_thinking=false` for servers that gate thinking in the chat template. Happy to make `LLM_NOTHINK=1` an alias if you'd prefer a single knob.

Typecheck is unchanged vs. main. Tests: 1710 pass. The one failure (`cli-lifecycle-safety`) also fails on `main`.

efenex:

PR #570, opened 2026-05-20
feat(lineage): mem::lineage primitive — chronological concept retrieval across all channels

efenex:
## Summary

Adds a new MCP primitive \`mem::lineage\` that answers "when did this term enter the corpus, and what surrounded it?" — chronologically ordered hits for a phrase across observation, memory, lesson, and summary channels, with optional adjacent-turn enrichment and graph-neighbor attachment.

Distinct from existing retrieval:
- \`mem::search\` ranks by relevance (BM25 hybrid).
- \`mem::smart-search\` is the lessons-first ranker.
- \`mem::lineage\` is **time-ordered**, **multi-channel**, and **enrichment-rich** — the right tool for "trace this term" / "what was the first mention" workflows.

## What's included

- **New REST endpoint** \`POST /agentmemory/lineage\` (+125 from 124)
- **New MCP tool** \`memory_lineage\` in CORE_TOOLS (CORE 15 / total 54)
- **Channels:** observation, memory, lesson, summary — opt in/out per call via \`channels: ["observation","lesson"]\`
- **Time bounds:** \`since\` / \`until\` ISO timestamps
- **Adjacent turns:** for observation hits, attach previous user prompt + previous assistant action (opt-in via \`includeAdjacentTurns\`, default true)
- **Graph neighbors:** when \`includeGraph: true\`, attach graph-edge neighbors for matching nodes
- **Order:** \`asc\` (default, oldest first) or \`desc\`
- **firstMention:** convenience field pointing at the earliest timestamp in the filtered set

## Gap-2 fix bundled

BM25 sweep cap raised from \`min(limit*4, 500)\` to \`min(limit*20, 5000)\` — large jsonl-imported sessions (10k+ observations) have deep-in-session references that didn't rank into the channel-filtered top N at the old cap. With wide channel-filtering, the sweep needs more headroom.

## Test plan

- [x] \`npm test\` passes (1140+ tests)
- [x] Mechanical smoke: \`mem::lineage\` with all 4 channels, \`firstMention\` populated, timeline sorted ASC, channel totals correct, adjacent turns attached, sourceFile extraction works, no-match returns empty arrays, empty query returns 400
- [x] Docs in \`docs/plans/v4-lineage-design.md\` + \`docs/plans/v4-lineage-test-case-careful-generator.md\`
- [x] Tool counts in README + AGENTS.md + boot message + test/mcp-standalone.test.ts updated

## Related

[...]

PR #571, opened 2026-05-20
feat(smart-search): boost title/narrative matches on 'who/what is X' queries

efenex:
## Summary

For named-concept queries ("who is the careful generator?", "what is a circuit breaker", "what does eventual consistency mean?"), the BM25 hybrid ranker scores busier observations above records that name the concept directly — question scaffolding tokens ("who", "is", "the") add noise that dilutes the true match signal. The record that *defines* the concept ranks below records that *mention* it incidentally.

## What it does

1. **Detect** the query as a named-concept pattern via 5 regexes (\`/who is/\`, \`/what is/\`, \`/what's/\`, \`/what does X mean/\`, \`/who's/\`). Skip if no match.
2. **Extract** the concept phrase (e.g. "careful generator"). Reject degenerate phrases — single tokens shorter than 3 chars (\`it\`, \`x\`) and phrases longer than 6 tokens.
3. **Deepen** the BM25 sweep to \`limit*3\` so the boost has candidates to re-rank (boost on a top-10 set has limited room to move records around).
4. **Re-rank** with multiplicative boosts:
   - Title contains the phrase → 2.0×
   - Narrative contains the phrase → 1.3×
5. **Same treatment** for lessons whose content contains the phrase (2.0×).
6. Re-sort by combined score, trim to original \`limit\`.

Non-named-concept queries are untouched.

## Why this lives in smart-search and not lineage

\`mem::lineage\` is chronologically-ordered and multi-channel; this is a ranking concern that affects the *primary* recall path (smart-search), which is what \`memory_recall\` / \`memory_smart_search\` MCP tools land on. Lineage benefits from upstream improvements in BM25 score, so this lift propagates.

## Test plan

- [x] \`npm test\` passes
- [x] New unit tests for \`extractNamedConcept\` (7 cases) — pattern matching, degenerate-phrase rejection
- [x] New integration test that proves the boost re-ranks: an observation whose title contains "careful generator" but has *lower* BM25 score than a busier unrelated observation gets moved to rank #1
- [x] Non-named-concept query preserves original ordering (regression test)

## Related

- Discovered while working on the "careful generator" test case in #570 (\`mem::lineage\`) — documented as Gap 4 there.
- Independent of #414 (time-range filtering for smart-search) — different concern, different lines.

## Summary by CodeRabbit

* **New Features**
  * Concept-aware search: queries like "who is/what is/what does…" are detected and re-rank results so matching observations and lessons surface higher for concept-focused queries.
  * More reliable concept detection for multiword phrases and edge cases, improving relevance for conceptual queries.

* **Tests**
  * Added tests covering concept extraction and verifying that conceptual queries trigger expected re-ranking while non-concept queries keep original order.

### DIFF
--- a/src/functions/smart-search.ts
+++ b/src/functions/smart-search.ts
@@ -72,6 +72,41 @@ export function resetFollowupStatsForTests(): void {
 // full content is fetched via memory_lesson_recall when the caller needs it.
 const LESSON_CONTENT_PREVIEW_CHARS = 240;
 
+// v4-B: detect "who is X" / "what is X" / "what does X mean" patterns
+// and pull out X so we can boost hits that name the concept directly.
+// BM25 already rewards the exact phrase, but for short named-concept
+// queries (typically 2–4 tokens) the question scaffolding ("who is the")
+// adds noise that depresses true matches relative to broader, busier
+// observations. This is the v4-A "careful generator" regression in
+// docs/plans/v4-lineage-test-case-careful-generator.md, surfacing in
+// smart-search rather than lineage.
+const NAMED_CONCEPT_PATTERNS: RegExp[] = [
+  /^\s*who\s+is\s+(?:the\s+|a\s+|an\s+)?(.+?)\s*\??\s*$/i,
+  /^\s*what\s+is\s+(?:the\s+|a\s+|an\s+)?(.+?)\s*\??\s*$/i,
+  /^\s*what(?:'s|\sis)\s+(?:the\s+|a\s+|an\s+)?(.+?)\s*\??\s*$/i,
+  /^\s*what\s+does\s+(.+?)\s+mean\s*\??\s*$/i,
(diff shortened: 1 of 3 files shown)


rohitg00 (approved):
Audited locally: build clean, 5 new tests pass (1086 total). Self-contained regex + boost multipliers, well-documented, fallback path leaves untouched queries unchanged. Multiplicative (title 2.0 × body 1.3 = 2.6×) when both match — CodeRabbit had already caught the prior else-if cap. Ready to merge.

efenex:
Rebased on current `main`. Composes with the upstream agentId filter — filters by agent first, then applies the named-concept boost, then trims to limit. Tests green; the only red check is the Vercel fork-preview.

efenex:
Friendly ping: this still merges cleanly onto current `main` (e04ba88) and is approved. Anything else you'd like before merging?

efenex:

PR #572, opened 2026-05-20
perf(summarize): skip mem::summarize when a fresh summary exists

efenex:
## Summary

Stop hooks fire \`mem::summarize\` on every assistant turn. For chatty sessions, summarize completes successfully every few seconds — and the next turn fires another full LLM-driven summarize pass. On serverless / rate-limited LLM providers, these stack: by the time one summarize finishes, the next is already in flight, both contending for the same provider rate budget.

This adds a freshness gate at the top of \`mem::summarize\`: if an existing \`SessionSummary\` for the requested sessionId is younger than \`SUMMARIZE_DEDUP_WINDOW_MS\` (default 90,000 ms / 90 s, set to \`0\` to disable), return the cached summary with \`skipped: "fresh"\` and skip the LLM work entirely.

## Smoke-validated

| Call | Wall-clock | Result |
|---|---|---|
| First call (small session, 6 obs) | **6.4 s** | real LLM run, qualityScore 100 → summary stored |
| Second call within 90 s | **0.02 s** | \`skipped: "fresh"\`, \`ageMs: 77\`, **zero LLM calls** |

\`MetricsStore\` is still pinged on the dedup path (as a successful call) so per-function latency stats stay accurate.

## Why now

This complements:
- **#472** (chunk large sessions) — already merged, reduced per-call LLM tokens but didn't address the per-turn re-summarize.
- The Stop-hook fire-and-forget work (separate PR) — server-side dedup means even when the hook does block briefly, the work is trivial.

Combined effect: typical Stop-hook → /agentmemory/summarize → dedup-hit chain is now ~50 ms server-side instead of a multi-LLM-call cycle.

## Test plan

- [x] \`npm test\` passes
- [x] Manual smoke against a small session: first call ran, second call within 90 s returned \`skipped: "fresh"\` with \`ageMs\` populated
- [x] \`SUMMARIZE_DEDUP_WINDOW_MS=0\` disables the gate
- [x] Default 90 s tunable per environment

## Notes

The window is env-configurable but not currently exposed in any config schema. Bump to \`600000\` (10 min) for very chatty long-running sessions if the default feels too aggressive.

## Summary by CodeRabbit

* **Performance**
  * Summary generation now reuses a recent summary when one exists within a configurable freshness window, skipping repeated reprocessing.
  * Faster summary retrieval and reduced resource use for a more responsive experience; freshness window can be adjusted (or disabled) to control skipping behavior.

### DIFF
--- a/src/functions/summarize.ts
+++ b/src/functions/summarize.ts
@@ -50,6 +50,15 @@ function getChunkConcurrency(): number {
   return Number.isFinite(n) && n > 0 ? n : CHUNK_CONCURRENCY_DEFAULT;
 }
 
+const DEDUP_WINDOW_MS_DEFAULT = 90_000;
+
+function getDedupWindowMs(): number {
+  const raw = process.env["SUMMARIZE_DEDUP_WINDOW_MS"];
+  if (raw === undefined) return DEDUP_WINDOW_MS_DEFAULT;
+  const n = Number(raw);
+  return Number.isFinite(n) && n >= 0 ? n : DEDUP_WINDOW_MS_DEFAULT;
+}
+
 // One chunk call with retry-once. Returns null when both attempts fail —
 // whether by parse failure, provider 4xx (content rejected by upstream
 // filters), or transient network/5xx errors that didn't recover on retry.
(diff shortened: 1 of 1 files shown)


rohitg00 (approved):
Audited locally: build clean, tests pass. 90s dedup is env-tunable via `SUMMARIZE_DEDUP_WINDOW_MS` (0 disables). Returns `skipped: "fresh"` so callers can detect the no-op path. Additive, low risk. Ready to merge.

efenex:
Friendly ping: this still merges cleanly onto current `main` (e04ba88) and is approved. Anything else you'd like before merging?

efenex:

PR #578, opened 2026-05-20
fix: read Claude Code tool_response in PostToolUse hook

### LINKED ISSUE #539: [BUG] PostToolUse hook reads data.tool_output but Claude Code sends tool_response — compresses fail XML schema (~47% loss)

yixinforu:
## Summary

  - prefer `tool_response` for Claude Code `PostToolUse` payloads
  - keep the legacy `tool_output` fallback for existing hook payloads
  - add regression coverage for both payload shapes

  Closes #539

## Summary by CodeRabbit

* **Bug Fixes**
  * Prioritizes the newer response field over the legacy field when extracting image data and constructing observation payloads, with a fallback to maintain compatibility.
* **Tests**
  * Adds integration tests that validate payload mapping and successful execution for both the new response field and the legacy fallback.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/578?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/plugin/scripts/post-tool-use.mjs
+++ b/plugin/scripts/post-tool-use.mjs
@@ -23,7 +23,8 @@ async function main() {
 	}
 	if (isSdkChildContext(data)) return;
 	const sessionId = data.session_id || "unknown";
-	const { imageData, cleanOutput } = extractImageData(data.tool_output);
+	const toolOutput = data.tool_response ?? data.tool_output;
+	const { imageData, cleanOutput } = extractImageData(toolOutput);
 	try {
 		await fetch(`${REST_URL}/agentmemory/observe`, {
 			method: "POST",
--- a/src/hooks/post-tool-use.ts
+++ b/src/hooks/post-tool-use.ts
@@ -31,8 +31,9 @@ async function main() {
   if (isSdkChildContext(data)) return;
 
   const sessionId = (data.session_id as string) || "unknown";
+  const toolOutput = data.tool_response ?? data.tool_output;
 
-  const { imageData, cleanOutput } = extractImageData(data.tool_output);
+  const { imageData, cleanOutput } = extractImageData(toolOutput);
 
   try {
     await fetch(`${REST_URL}/agentmemory/observe`, {
--- a/test/post-tool-use-hook.test.ts
+++ b/test/post-tool-use-hook.test.ts
@@ -1,0 +1,123 @@
+import { describe, it, expect } from "vitest";
+import { spawn } from "node:child_process";
+import { createServer, type IncomingMessage } from "node:http";
+import { join } from "node:path";
+
+const HOOK = join(
+  import.meta.dirname,
+  "..",
+  "plugin",
+  "scripts",
+  "post-tool-use.mjs",
+);
+
+function readBody(req: IncomingMessage): Promise<string> {
+  return new Promise((resolve, reject) => {
+    let body = "";
+    req.setEncoding("utf8");
+    req.on("data", (chunk) => {
+      body += chunk;
+    });
+    req.on("end", () => resolve(body));
+    req.on("error", reject);
+  });
+}
+
+async function runHook(payload: Record<string, unknown>) {
+  let received: unknown;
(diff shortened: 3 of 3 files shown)


yixinforu:
Fixed. Moved the test server cleanup into a finally block so the server is closed even if the child process or request
  handling path rejects.

rohitg00:
Shipped in v0.9.22 via maintainer PR — https://github.com/rohitg00/agentmemory/releases/tag/v0.9.22. Thanks for the contribution.

Issue #469, opened 2026-05-17
Vector Index Dimension Mismatch Crashes Server on Startup (2048 vs 384)


nickdesi:
## Environment

- **agentmemory version**: v0.9.17
- **OS**: macOS (darwin, Apple Silicon)
- **Embedding provider**: `local` (`EMBEDDING_PROVIDER=local` in `~/.agentmemory/.env`)
- **Client**: OpenCode Desktop (CLI AI coding assistant)
- **Launch method**: LaunchAgent (`~/Library/LaunchAgents/ai.agentmemory.plist`)

## Problem

The agentmemory server **refuses to start** and crashes in a restart loop due to a persisted vector index dimension mismatch:

```
[agentmemory] Fatal: Error: [agentmemory] Refusing to start: persisted vector index has 19 of 19 vectors with the wrong dimension. Active provider (local) declares 384; dimensions seen on disk: 2048.
```

The server was previously running with an embedding provider that produced **2048-dimensional vectors** (likely a cloud provider like OpenAI `text-embedding-3-large` or similar). The configuration was later changed to `EMBEDDING_PROVIDER=local`, which produces **384-dimensional vectors**. On startup, the server detects the mismatch and **refuses to start** to prevent silent search corruption.

## Impact

1. **Server crash loop**: The LaunchAgent keeps restarting the server, which crashes immediately, generating endless log entries.
2. **All MCP tools unavailable**: Every `agentmemory-mcp` tool call fails because the REST API at `localhost:3111` never becomes available.
3. **No graceful degradation**: The server does not offer a way to start with search disabled or auto-migrate the index.
4. **No user-facing guidance**: The error message mentions three options but none are actionable from the `.env` file when the server cannot start.

## Reproduction Steps

1. Run agentmemory with a cloud embedding provider (e.g., OpenAI) that produces 2048-dim vectors
2. Accumulate observations (vectors get persisted to disk)
3. Change `EMBEDDING_PROVIDER` to `local` in `~/.agentmemory/.env`
4. Restart the server (via LaunchAgent or manually)
5. Server crashes with the fatal dimension mismatch error

## Current Workaround

The only way to recover is to manually add `AGENTMEMORY_DROP_STALE_INDEX=true` to the environment **before** starting the server process. However:

- This is **not documented** in the `.env` example or README
- The LaunchAgent plist does not include this variable by default
- The `.env` file is read by the server **after** the vector index check, so adding it there does not help if the server crashes before reading `.env`
[...]

Issue #455, opened 2026-05-17
Worker refuses to start after embedding provider change: stale index dimension mismatch (2048D vs 384D)


nickdesi:
## Problem

After switching embedding providers (or after an agentmemory upgrade that changed the default local model), the worker refuses to start with a fatal error:

```
[agentmemory] Fatal: Error: [agentmemory] Refusing to start: persisted vector index has 19 of 19 vectors with the wrong dimension. Active provider (local) declares 384; dimensions seen on disk: 2048. First mismatched obsIds: mem_mp8ctmgi_37deb553385b (dim=2048), ...
```

The worker process exits immediately, leaving the REST API (:3111) unreachable. The dashboard (:3113) stays in "CONNECTING..." state with all counters at 0.

## Root Cause

The local embedding provider switched from a 2048D model to `all-MiniLM-L6-v2` (384D). The persisted vector index on disk still contains vectors from the old provider. The worker detects this mismatch and refuses to start to prevent silent search corruption.

## Workaround

Add to `~/.agentmemory/.env`:

```env
AGENTMEMORY_DROP_STALE_INDEX=true
```

Then restart the worker. **All existing memories are lost** — the index is purged and rebuilt from scratch.

## Expected Behavior

1. **Graceful degradation**: The worker should start even with a stale index, falling back to BM25-only search until re-embedding completes
2. **Auto-migration**: Detect dimension mismatch and offer to re-embed existing vectors against the new provider (with a progress indicator)
3. **Clear CLI guidance**: The error message mentions 3 options but requires manual `.env` editing. A CLI flag like `agentmemory start --drop-stale-index` or `agentmemory reset-index` would be more discoverable
4. **Dashboard visibility**: The dashboard should show a warning banner when the index is stale, rather than silently showing "0 sessions" with "CONNECTING..."

## Environment

- **agentmemory**: v0.9.17
- **iii-engine**: v0.11.6
- **Host**: OpenCode Desktop 1.15.3
- **OS**: macOS 15.4 (arm64)
- **Embedding provider**: local (all-MiniLM-L6-v2, 384D)
- **Previous provider**: unknown (2048D vectors on disk)

## Reproduction Steps

1. Run agentmemory with an embedding provider that produces 2048D vectors
2. Create some sessions/memories
3. Switch to local provider (384D) or upgrade agentmemory to a version with a different default model
4. Restart agentmemory → worker crashes with dimension mismatch error
[...]

PR #585, opened 2026-05-21
fix(vector-index): respect Buffer slice metadata in base64 round-trip

### LINKED ISSUE #584: `base64ToFloat32` produces a phantom 2048-element view over Node's shared Buffer pool (likely root cause of #455 / #469)
## TL;DR

`base64ToFloat32` in `src/state/vector-index.ts` constructs a `Float32Array` over **the underlying ArrayBuffer of the decoded Buffer**, not over its byte slice. When Node returns a Buffer that lives inside its shared pool (which happens when the decoded byte length is small enough), the result is a view over the entire pool ArrayBuffer:

1. **Phantom dim** — `loaded.length` becomes `Buffer.poolSize / 4 = 2048` regardless of the real dim, so the dim-guard at boot misreads on-disk data and refuses to start with `dimensions seen on disk: 2048`. The disk content is correct.
2. **Data at wrong indices on reload** — `new Float32Array(buf.buffer)` views from offset 0, ignoring `buf.byteOffset`, so `loaded[0]` reads the start of the pool, not the start of the decoded slice. Successive decodes that land in the same pool batch see one another's data interleaved.

I think this is the underlying cause of #455 and #469 — both reports show the exact `2048` value, which is `Buffer.poolSize / 4`, not a realistic provider dimension.

**Important scope:** this only bites when Node pools the decoded buffer, which means it only affects providers whose encoded vector is small enough to land in the pool. In Node `v22`/`v24` the `LocalEmbeddingProvider` (384 dims → 1536 bytes) pools and triggers the bug; `gemini` (768), `voyage`/`cohere` (1024), `openai` default (1536) all decode into standalone allocations and load correctly. So in practice the only currently-observable victim is `EMBEDDING_PROVIDER=local`, which lines up with every report I've seen.

## The bug

`src/state/vector-index.ts` on `main` (commit c14bdc5, lines 1–7):

```ts
function float32ToBase64(arr: Float32Array): string {
  return Buffer.from(arr.buffer).toString("base64");
}

function base64ToFloat32(b64: string): Float32Array {
  return new Float32Array(Buffer.from(b64, "base64").buffer);
}
```

`Buffer.from(string, "base64")` returns a `Buffer` that may be a slice of an internal pool (default `Buffer.poolSize === 8192`). The `.buffer` accessor returns the **entire** pool ArrayBuffer; `buf.byteOffset` and `buf.byteLength` describe the actual slice. `new Float32Array(arrayBuffer)` ignores those and views from offset 0 over the whole ArrayBuffer, so `length === pool.byteLength / 4 === 2048` and reads start from pool[0], not from the decoded slice.

The encoder has the symmetric defect (ignores `arr.byteOffset` / `arr.byteLength`). It currently does no harm because `LocalEmbeddingProvider` returns freshly-allocated `Float32Array(v)`s and `VectorIndex.restoreFrom` copies through `new Float32Array(entry.embedding)`, so every encoded array has `byteOffset===0` and `byteLength===arr.buffer.byteLength`. Worth fixing for symmetry and to keep future call sites safe.

## Reproducer (no agentmemory needed)

```js
function float32ToBase64(arr) { return Buffer.from(arr.buffer).toString("base64"); }
function base64ToFloat32(b64) { return new Float32Array(Buffer.from(b64, "base64").buffer); }

console.log("Node:", process.version, "Buffer.poolSize:", Buffer.poolSize);

// 1) Phantom dim is 2048 specifically for 384-dim (1536 bytes pools);
//    larger dims allocate standalone and load correctly.
for (const dim of [384, 768, 1024, 1536]) {
  const v = new Float32Array(dim); for (let i = 0; i < dim; i++) v[i] = i;
  const loaded = base64ToFloat32(float32ToBase64(v));
  console.log(`dim=${dim} -> loaded.length=${loaded.length} (phantom=${loaded.length === 2048})`);
}

// 2) Reload corrupts indices: each loaded vector's [0] reads pool[0..3],
//    not the decoded slice's first 4 bytes.
const DIM = 384;
const truth = [];
const ser = [];
for (let n = 0; n < 10; n++) {
  const v = new Float32Array(DIM); for (let i = 0; i < DIM; i++) v[i] = n * 1000 + i;
  truth.push(v); ser.push(float32ToBase64(v));
}
const loaded = ser.map(base64ToFloat32);
console.log("\nDistinct ArrayBuffers across 10 loaded vectors:",
  new Set(loaded.map(a => a.buffer)).size);
console.log("loaded[0][0]:", loaded[0][0], "truth:", truth[0][0]);
console.log("loaded[5][0]:", loaded[5][0], "truth:", truth[5][0]);
console.log("loaded[9][0]:", loaded[9][0], "truth:", truth[9][0]);
```

Output on Node `v24.12.0`:

```
Node: v24.12.0 Buffer.poolSize: 8192
dim=384 -> loaded.length=2048 (phantom=true)
dim=768 -> loaded.length=768 (phantom=false)
dim=1024 -> loaded.length=1024 (phantom=false)
dim=1536 -> loaded.length=1536 (phantom=false)

Distinct ArrayBuffers across 10 loaded vectors: 3
loaded[0][0]: 1.2891945871788317e-43 truth: 0
loaded[5][0]: 3000 truth: 5000
loaded[9][0]: 8000 truth: 9000
```

Note that the 10 vectors share only 3 underlying ArrayBuffers (they were batched into pools of ~5), and the values at `[0]` don't match the truth — they're whatever happens to be at the start of each pool at decode time, not the actual first element of the vector that key points to.

## How I caught it

Three back-to-back dim-guard crashes on `0.9.21` on the same Windows install, none of which had ever used a 2048-dim provider, prompted me to base64-decode an embedding from disk by hand:

```powershell
$content = Get-Content '~/.agentmemory/data/state_store.db/mem%3Aindex%3Abm25.bin' -Raw
$m = [regex]::Match($content, '\\"embedding\\":\\"([A-Za-z0-9+/]+={0,2})\\"')
$bytes = [Convert]::FromBase64String($m.Groups[1].Value)
"base64=$($m.Groups[1].Value.Length), bytes=$($bytes.Length), dim=$($bytes.Length/4)"
# => base64=2048, bytes=1536, dim=384
```

Disk content is correct 384-dim. Only the decoder lies.

## Fix

```ts
function float32ToBase64(arr: Float32Array): string {
  return Buffer.from(arr.buffer, arr.byteOffset, arr.byteLength).toString("base64");
}

function base64ToFloat32(b64: string): Float32Array {
  const buf = Buffer.from(b64, "base64");
  return new Float32Array(buf.buffer, buf.byteOffset, buf.byteLength / 4);
}
```

Both helpers now respect the slice metadata. With this applied, the reproducer above prints `loaded[0][0]=0, loaded[5][0]=5000, loaded[9][0]=9000` and `loaded.length=384` for all dims.

## Verification on a real index

I patched both `dist` bundles in place and restarted on top of the same 2051-vector index that was previously rejected:

```
[agentmemory] Embedding provider: local (384 dims)
[agentmemory] Loaded persisted BM25 index (2051 docs)
[agentmemory] Loaded persisted vector index (2051 vectors)
[agentmemory] Ready. Triple-stream (BM25+Vector+Graph) search active.
```

No Fatal, no `dimensions seen on disk: 2048` — the same on-disk data that was previously crashing the boot loads cleanly.

## Impact

1. **Phantom dim-guard crashes for `EMBEDDING_PROVIDER=local`** — every `2048` in #455 / #469 is `Buffer.poolSize / 4`, not a real dim. Users have been losing memories to `AGENTMEMORY_DROP_STALE_INDEX=true` "recoveries" they did not actually need.
2. **Silent search corruption when local vectors get reloaded** — once an index is reloaded from disk, the in-memory `vectors` map is populated with views that read from pool offset 0 rather than the obs's actual data. The phantom dim normally masks this because the dim-guard refuses to boot, but anywhere the guard is skipped (e.g. an older code path, or any caller wrapping `VectorIndex.deserialize` directly) would compute cosine similarity over the wrong bytes. The reproducer shows that the indices don't line up with the truth values.

Other providers (`gemini`, `voyage`, `cohere`, `openai`, `openrouter`) decode into standalone allocations on current Node and are not affected today, but the code is still wrong and the threshold can change with future Node versions or larger `Buffer.poolSize` defaults.

## Environment

- `@agentmemory/agentmemory@0.9.21` (also reproduces against `main` HEAD)
- `node v24.12.0`, `Buffer.poolSize = 8192`
- `EMBEDDING_PROVIDER=local`
- Windows 11, but reproducer is platform-independent

Happy to send a PR if useful.

george-petrakis:
Buffer.from(b64, "base64") may return a slice of Node's internal pool (poolSize=8192). Calling new Float32Array(buf.buffer) ignores byteOffset and byteLength, producing a 2048-element view over the whole pool instead of the actual decoded slice — the root cause of the phantom "dimensions seen on disk: 2048" crashes reported in #455 and #469.

Pass byteOffset/byteLength to both helpers so the Float32Array covers only the decoded bytes. Add a regression test that serializes and deserialises 384-dim vectors (1536 bytes, within pool threshold) and asserts correct dimension and nearest-neighbour identity after reload.

Fixes #584

## Release Notes

* **Bug Fixes**
  * Fixed vector serialization to ensure data integrity when saving and restoring vector indices.

* **Tests**
  * Added test coverage validating vector index serialization and restoration consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/585?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/state/vector-index.ts
+++ b/src/state/vector-index.ts
@@ -1,9 +1,10 @@
 function float32ToBase64(arr: Float32Array): string {
-  return Buffer.from(arr.buffer).toString("base64");
+  return Buffer.from(arr.buffer, arr.byteOffset, arr.byteLength).toString("base64");
 }
 
 function base64ToFloat32(b64: string): Float32Array {
-  return new Float32Array(Buffer.from(b64, "base64").buffer);
+  const buf = Buffer.from(b64, "base64");
+  return new Float32Array(buf.buffer, buf.byteOffset, buf.byteLength / 4);
 }
 
 function cosineSimilarity(a: Float32Array, b: Float32Array): number {
--- a/test/vector-index.test.ts
+++ b/test/vector-index.test.ts
@@ -76,4 +76,30 @@ describe("VectorIndex", () => {
     const results = index.search(new Float32Array([1, 0, 0]));
     expect(results[0].score).toBe(0);
   });
+
+  it("serialize round-trip preserves dimension and values for pooled-buffer sizes", () => {
+    // 384 floats = 1536 bytes, small enough for Node to pool the decoded Buffer.
+    // Without the byteOffset/byteLength fix, deserialized length becomes 2048
+    // and values read from pool offset 0 instead of the slice.
+    const DIM = 384;
+    const vecs = Array.from({ length: 5 }, (_, n) => {
+      const v = new Float32Array(DIM);
+      for (let i = 0; i < DIM; i++) v[i] = n * 1000 + i;
+      return v;
+    });
+
+    vecs.forEach((v, n) => index.add(`obs_${n}`, "ses_1", v));
+    const restored = VectorIndex.deserialize(index.serialize());
+
+    expect(restored.size).toBe(5);
+    // Dimension must be exactly DIM, not 2048 (Buffer.poolSize / 4).
+    const { mismatches } = restored.validateDimensions(DIM);
+    expect(mismatches).toEqual([]);
+    // Each vector must be its own nearest neighbour after reload.
+    for (let n = 0; n < 5; n++) {
+      const results = restored.search(vecs[n], 1);
+      expect(results[0].obsId).toBe(`obs_${n}`);
+      expect(results[0].score).toBeCloseTo(1.0, 4);
+    }
+  });
 });


cuongdevv:
Confirming this fix works end-to-end. I applied the exact same `byteOffset`/`byteLength` patch to both `dist/src-D5arboxc.mjs` and `dist/index.mjs` on my install (0.9.21, Node v24.12.0, `EMBEDDING_PROVIDER=local`, 2051-vector index that was previously phantom-2048 crashing every boot) and it loads cleanly:

```
[agentmemory] Embedding provider: local (384 dims)
[agentmemory] Loaded persisted vector index (2051 vectors)
[agentmemory] Ready. Triple-stream (BM25+Vector+Graph) search active.
```

The regression test in `test/vector-index.test.ts` is exactly the right shape — 384-dim (1536 bytes) is the only size that lands in Node's `Buffer.poolSize=8192` pool, so it's the only case that reproduces. Larger dims (768/1024/1536) allocate standalone and were silently fine, which is why this bug only ever surfaced for `local` users.

Thanks @kek-Sec for turning this into a proper PR. Hope @rohitg00 can get this merged + a 0.9.22 cut soon — every day this sits unmerged is another `dimensions seen on disk: 2048` false alarm someone "recovers" from by nuking their vector index via `AGENTMEMORY_DROP_STALE_INDEX=true`.

george-petrakis:
also closes #587

rohitg00:
Verified the same fix lands in #653 with `Float32Array.BYTES_PER_ELEMENT` instead of the magic `/ 4` constant. Both PRs solve the same root cause (Buffer.from(b64, 'base64') returns a slice of Node's pool; new Float32Array(buf.buffer) ignores offset/length).

Recommending #653 as the merge target for the named-constant clarity. Your 5-vector round-trip + validateDimensions test is more thorough than the one in #653 though — would be great as a follow-up PR after #653 lands. Thanks @kek-Sec.

rohitg00:
Thanks @kek-Sec — during bug-fix triage we landed an equivalent fix in #683. Closing this as superseded; really appreciate the report and the regression test that helped pin down the root cause.

Issue #510, opened 2026-05-18
Bug: MCP tools not loaded by Claude Code despite healthy server and working plugin


chrischjh:
### Environment
- agentmemory version: 0.9.18 (server) / 0.9.19 (plugin cache)
- Claude Code: latest (May 2026)
- OS: macOS Darwin 25.3.0 (ARM64)
- Node: via npx

### Steps to reproduce
1. `npx -y @agentmemory/agentmemory` — server starts, health check returns `healthy`
2. `/plugin marketplace add rohitg00/agentmemory` in Claude Code — succeeds
3. `/plugin install agentmemory` — succeeds
4. `/reload-plugins` — shows "1 plugin MCP server"
5. Check available MCP tools

### Expected behavior
53 MCP tools (`memory_save`, `memory_smart_search`, `memory_recall`, etc.) appear in Claude Code's tool list.

### Actual behavior
- Server healthy at `localhost:3111` ✅
- Viewer accessible at `localhost:3113` ✅
- Plugin skills loaded (8 skills: remember, recall, forget, recap, etc.) ✅
- Plugin hooks loaded (6 hooks: SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, PostToolUseFailure, PreCompact) ✅
- MCP shim works when tested manually: `echo '{"jsonrpc":"2.0",...}' | npx -y @agentmemory/mcp` returns proper `initialize` response with `serverInfo.name: "agentmemory"` ✅
- **MCP tools NOT available** in Claude Code ❌
- `Available MCP Servers` list does NOT include `agentmemory` ❌

### Debugging done
1. Created `~/.claude/mcp.json` with explicit `AGENTMEMORY_URL=http://localhost:3111`
2. Created project-level `.claude/mcp.json` (same content)
3. Set `AGENTMEMORY_URL` in `~/.claude/settings.json` env vars
4. Created `~/.agentmemory/.env` with `AGENTMEMORY_URL` and `AGENTMEMORY_TOOLS=all`
5. Claude Code `/reload-plugins` output: "1 plugin MCP server" — confirms it sees the MCP server
6. Manual MCP shim test passes — confirms `@agentmemory/mcp` package works

### Possible cause
The plugin `.mcp.json` uses `${AGENTMEMORY_URL}` env var reference. Claude Code may not be expanding `settings.json` env vars into the MCP server process environment. Even with explicit `~/.claude/mcp.json` (no variable references), the server doesn't show in the tool list.

### Reload output
[...]

Issue #400, opened 2026-05-15
Codex plugin install does not expose advertised full MCP tool surface


ivkiwi:
The Codex plugin path currently does not reliably expose the advertised 51 MCP tools out of the box.

Observed behavior:

- `plugin/.mcp.json` starts `@agentmemory/mcp` without any environment.
- `@agentmemory/agentmemory` starts with `AGENTMEMORY_TOOLS=core` by default.
- The REST endpoint `/agentmemory/mcp/tools` then returns only the core tool set.
- The standalone MCP shim returns the remote server tool list as-is, so Codex sees a reduced tool surface instead of the README-advertised 51 tools.

**There is also a docs/install mismatch: the README says to run `codex plugin install agentmemory`, but the tested Codex CLI only exposed `codex plugin marketplace {add,upgrade,remove}` and no `plugin install` subcommand.**

## Environment

- OS: macOS Darwin 25.5.0 arm64
- Node: v26.0.0
- npm: 11.14.1
- Package: `@agentmemory/agentmemory@0.9.13`
- MCP shim: `@agentmemory/mcp@0.9.13`
- Codex CLI: `/opt/homebrew/bin/codex`

## Reproduction

1. Start the worker with documented default:

   ```bash
   npx @agentmemory/agentmemory
   ```

2. Query the server MCP tools endpoint:

   ```bash
   curl http://127.0.0.1:3111/agentmemory/mcp/tools
   ```

3. Install/use the Codex plugin manifest from `plugin/.mcp.json`:

   ```json
   {
     "mcpServers": {
       "agentmemory": {
         "command": "npx",
         "args": ["-y", "@agentmemory/mcp"]
       }
     }
   }
   ```

4. Ask Codex for `agentmemory` tools.

## Actual Result

Codex sees a reduced MCP surface instead of 51 tools.

In local testing:

- Worker started without `AGENTMEMORY_TOOLS=all` exposed only a small/core tool set from `/agentmemory/mcp/tools`.
- Standalone shim used the remote reduced list as authoritative.
- If no worker was reachable, local fallback exposed only 7 implemented tools.

## Expected Result

The Codex plugin should expose all advertised agentmemory MCP tools when a worker is running, matching README language:

> "you get 51 MCP tools ... without any extra config step"

## Root Cause

Two config layers disagree:

1. Worker default:

   ```ts
   const mode = process.env["AGENTMEMORY_TOOLS"] || "core";
   ```

[...]

PR #616, opened 2026-05-23
fix(mcp): honor AGENTMEMORY_TOOLS from .env in getVisibleTools

### LINKED ISSUE #553: OpenCode plugin exposes only 8 agentmemory tools instead of advertised 51 MCP surface (server running, proxy mode not activating)
### Description

On OpenCode with the agentmemory plugin installed and the agentmemory server running (health endpoint returns `{"status":"healthy"}` on the default port), the available memory tools are limited to 8 instead of the advertised 51:

**Tools available:**
- `memory_consolidate`
- `memory_diagnose`
- `memory_lesson_save`
- `memory_recall`
- `memory_reflect`
- `memory_save`
- `memory_sessions`
- `memory_smart_search`

**Tools from the documented 7-tool fallback that are MISSING:**
- `memory_export`
- `memory_audit`
- `memory_governance_delete`

**Tools from the 51-tool set that should be available in proxy mode but are absent:**
- `memory_graph_query`
- `memory_relations`
- `memory_timeline`
- `memory_file_history`
- `memory_patterns`
- `memory_profile`
- etc.

The README states: "The published @agentmemory/mcp package is a thin shim. It exposes the full 51-tool surface only when it can reach a running agentmemory server via AGENTMEMORY_URL (proxy mode). With no server reachable, the shim falls back to a 7-tool local set."

The server IS reachable, but the tool set doesn't match either the 7-tool fallback OR the 51-tool full set — it's a third, undocumented intermediate state (8 tools, 4 match the fallback, 4 are different).

Related existing issues:
- #400 (open) Codex plugin install does not expose advertised full MCP tool surface
- #510 (open) Bug: MCP tools not loaded by Claude Code despite healthy server and working plugin
- #431 (open) Feature Request: Native Hooks / Auto-Context Injection for OpenCode
- #464 (open) docs: refresh integration capability counts

### Impact

Users following "22 hooks + MCP + plugin" expectations from the README get only 8 tools with no indication that 43 are missing. Critical tools like `memory_export`, `memory_audit`, `memory_governance_delete`, and all knowledge graph tools are absent without explanation.

### To Reproduce

1. Install agentmemory globally
2. Start the server (`agentmemory` or `npx @agentmemory/agentmemory`)
3. Verify server is healthy: `curl http://localhost:3111/agentmemory/health`
4. Use OpenCode with the agentmemory plugin installed
5. List available agentmemory tools — only 8 appear

### Expected

Either:
- The shim detects the running server and proxies all 51 tools (as documented), or
- The OpenCode plugin documents its actual tool set and explains why some are missing

### Environment

- Agent: OpenCode with agentmemory plugin
- Server responds at localhost:3111
- Default configuration

### Notes

This may share a root cause with #400 and #510: the MCP shim's server-probe failing to detect a running server on the default port when launched through certain plugin/hook systems. An OpenCode-specific diagnostic would help narrow down whether it's a shim issue or a plugin registration issue.

ccf:
## What

`getVisibleTools()` in `src/mcp/tools-registry.ts` read `process.env["AGENTMEMORY_TOOLS"]` directly, bypassing `getMergedEnv()` — the `{ ...fileEnv, ...process.env }` helper that every other config reader reaches through `getEnvVar()`. This routes the lookup through `getEnvVar()` instead.

## Why

`AGENTMEMORY_TOOLS=all` set in `~/.agentmemory/.env` was silently ignored. The MCP tool surface stayed at the 8-tool core set instead of the full surface, even though `.env.example` documents that variable as the way to opt in:

```
AGENTMEMORY_TOOLS=all   # core (default) | all — surface exposed to MCP clients
```

Because `getMergedEnv()` is `{ ...fileEnv, ...process.env }`, `process.env` (shell export, the `--tools` flag) still takes precedence over the file — only the missing `.env` fallback is restored.

This is the root cause behind #553: the "8 tools, not 51" state is the server returning its `core` surface over a *working* proxy — not a probe/proxy failure. It may also explain #400 (Codex full-surface) where a healthy server still serves the reduced set.

To be clear about scope: this does **not** address #510, which is the distinct "literal `${AGENTMEMORY_URL}` placeholder breaks the shim's localhost fallback" connection bug — that one is already handled on `main` by the `resolveEnvOrEmpty` guard in #588.

## How to verify

```bash
# with AGENTMEMORY_TOOLS=all only in ~/.agentmemory/.env (not exported):
# before — server reports 8 core tools; after — full surface
```

Automated coverage added in `test/tools-visibility-env.test.ts` (hermetic; sandboxes `HOME` and writes a temp `.env`):
- `AGENTMEMORY_TOOLS=all` in `.env` → full surface (`getAllTools().length`, incl. `memory_graph_query`)
- unset → core subset (regression guard)
- `process.env` overrides the `.env` value (precedence preserved)

`npm run build` is clean and the suite passes (the 5 unrelated `auto-compress` / `embedding-provider` failures only occur when a real `~/.agentmemory/.env` is present in the test environment; they pass with a clean `HOME`).

Fixes #553

### DIFF
--- a/src/mcp/tools-registry.ts
+++ b/src/mcp/tools-registry.ts
@@ -1,3 +1,5 @@
+import { getEnvVar } from "../config.js";
+
 export type McpToolDef = {
   name: string;
   description: string;
@@ -942,7 +944,7 @@ export function getAllTools(): McpToolDef[] {
 }
 
 export function getVisibleTools(): McpToolDef[] {
-  const mode = process.env["AGENTMEMORY_TOOLS"] || "core";
+  const mode = getEnvVar("AGENTMEMORY_TOOLS") || "core";
   if (mode === "all") return getAllTools();
   return getAllTools().filter((t) => ESSENTIAL_TOOLS.has(t.name));
 }
--- a/test/tools-visibility-env.test.ts
+++ b/test/tools-visibility-env.test.ts
@@ -1,0 +1,64 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { mkdtempSync, writeFileSync, mkdirSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+const ORIGINAL_HOME = process.env["HOME"];
+const ORIGINAL_USERPROFILE = process.env["USERPROFILE"];
+const ORIGINAL_TOOLS = process.env["AGENTMEMORY_TOOLS"];
+
+let sandboxHome: string;
+
+async function freshTools() {
+  vi.resetModules();
+  return await import("../src/mcp/tools-registry.js");
+}
+
+function writeEnv(contents: string) {
+  const dir = join(sandboxHome, ".agentmemory");
+  mkdirSync(dir, { recursive: true });
+  writeFileSync(join(dir, ".env"), contents);
+}
+
+describe("getVisibleTools env resolution", () => {
+  beforeEach(() => {
+    sandboxHome = mkdtempSync(join(tmpdir(), "agentmemory-tools-"));
+    process.env["HOME"] = sandboxHome;
+    process.env["USERPROFILE"] = sandboxHome;
+    delete process.env["AGENTMEMORY_TOOLS"];
+  });
+
+  afterEach(() => {
+    if (ORIGINAL_HOME === undefined) delete process.env["HOME"];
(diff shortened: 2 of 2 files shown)


rohitg00:
Default MCP tool surface flipped to all 51 tools in v0.9.22 (#650) — https://github.com/rohitg00/agentmemory/releases/tag/v0.9.22. Thanks for the contribution.

PR #617, opened 2026-05-23
feat: add privacy-safe audit receipts

caioribeiroclw-pixel:
## What

Adds an opt-in privacy-safe audit receipt format for existing audit trails.

- `buildAuditReceipt()` converts audit entries to `agentmemory.audit.receipt.v1`
- hashes audit ids, target ids, and user ids with SHA-256
- includes operation/function/timestamp/target counts/detail keys, but not raw `targetIds`, raw `details`, or raw `userId`
- exposes receipts via `GET /agentmemory/audit?receipt=true`
- exposes the same mode through `memory_audit` with `receipt: true`

## Why

Persistent memory tools need a shareable proof that memory operations happened without leaking the memory content, paths, raw ids, prompts, or user identifiers. The normal audit endpoint remains unchanged; the receipt mode is for safe handoff/debug/audit artifacts.

## Verification

- `npm run build`
- `npm test -- --run test/audit.test.ts`
- `npm test` (1097 tests passed)
- `git diff --check`

Note: `npm install` needed `--legacy-peer-deps` locally because the current dependency tree has a peer constraint conflict between `@anthropic-ai/sdk@0.39.0` and `@anthropic-ai/claude-agent-sdk`.

## Summary by CodeRabbit

* **New Features**
  * Privacy-safe audit receipts: request audit data with `receipt=true` to receive stable hashed identifiers, counts and metadata instead of raw sensitive values (supported by API endpoints, MCP tools, proxy and local modes).

* **Documentation**
  * README and REST docs updated to describe the `receipt` parameter, receipt format, and requirement for a configured HMAC key for stable receipts.

* **Tests**
  * Added tests validating receipt generation, counts, hashed identifiers, privacy flags, and omission of raw sensitive fields.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/617?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/README.md
+++ b/README.md
@@ -854,7 +854,7 @@ npm install @xenova/transformers
 | `memory_claude_bridge_sync` | Sync with MEMORY.md |
 | `memory_team_share` | Share with team members |
 | `memory_team_feed` | Recent shared items |
-| `memory_audit` | Audit trail of operations |
+| `memory_audit` | Audit trail of operations; pass `receipt: true` for privacy-safe hashed receipts |
 | `memory_governance_delete` | Delete with audit trail |
 | `memory_snapshot_create` | Git-versioned snapshot |
 | `memory_action_create` | Create work items with dependencies |
@@ -1262,7 +1262,7 @@ Create `~/.agentmemory/.env`:
 | `POST` | `/agentmemory/import` | Import from JSON |
 | `POST` | `/agentmemory/graph/query` | Knowledge graph query |
 | `POST` | `/agentmemory/team/share` | Share with team |
-| `GET` | `/agentmemory/audit` | Audit trail |
+| `GET` | `/agentmemory/audit` | Audit trail; add `?receipt=true` for privacy-safe hashed receipts |
 
 Full endpoint list: [`src/triggers/api.ts`](src/triggers/api.ts)
 
--- a/README.md
+++ b/README.md
@@ -854,7 +854,7 @@ npm install @xenova/transformers
 | `memory_claude_bridge_sync` | Sync with MEMORY.md |
 | `memory_team_share` | Share with team members |
 | `memory_team_feed` | Recent shared items |
-| `memory_audit` | Audit trail of operations; pass `receipt: true` for privacy-safe hashed receipts |
+| `memory_audit` | Audit trail of operations; pass `receipt: true` for privacy-safe HMAC-hashed receipts (requires `AGENTMEMORY_RECEIPT_HMAC_KEY`) |
 | `memory_governance_delete` | Delete with audit trail |
 | `memory_snapshot_create` | Git-versioned snapshot |
 | `memory_action_create` | Create work items with dependencies |
@@ -1262,7 +1262,9 @@ Create `~/.agentmemory/.env`:
 | `POST` | `/agentmemory/import` | Import from JSON |
 | `POST` | `/agentmemory/graph/query` | Knowledge graph query |
 | `POST` | `/agentmemory/team/share` | Share with team |
-| `GET` | `/agentmemory/audit` | Audit trail; add `?receipt=true` for privacy-safe hashed receipts |
+| `GET` | `/agentmemory/audit` | Audit trail; add `?receipt=true` for privacy-safe HMAC-hashed receipts |
+
+Audit receipts require `AGENTMEMORY_RECEIPT_HMAC_KEY` when `receipt=true` is requested. The key is used only to derive stable receipt identifiers with HMAC-SHA256 so shared receipts do not expose raw or dictionary-guessable audit, target, or user IDs.
 
 Full endpoint list: [`src/triggers/api.ts`](src/triggers/api.ts)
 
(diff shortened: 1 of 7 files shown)


arian-gogani:
cool to see audit receipts being added here. a few things that might be useful based on what we learned building a similar receipt format:

1. the SHA-256 hashing over audit IDs and user IDs is the right approach for privacy-safe receipts. one thing to watch: if you're using `JSON.stringify()` for hashing, the output isn't deterministic across implementations (key order can vary). RFC 8785 (JSON Canonicalization Scheme) solves this. we use `rfc8785` in Python and `canonicalize` in JS.

2. consider signing the receipt with Ed25519 (or similar). a hash proves the content hasn't changed, but a signature proves *who* generated it. without a signature, anyone with the data can regenerate the hash.

3. if receipts need to be verifiable across sessions, a `canon_version` field helps with forward compatibility. receipts generated under different hashing rules stay distinguishable.

we built an open-source receipt format for AI agents that handles all of this (Ed25519 signing, JCS canonicalization, tamper detection, hash-linked chains): github.com/arian-gogani/nobulex

happy to share more if any of this is useful.

caioribeiroclw-pixel:
Updated the PR in `a9b92d2` to address the privacy review points:

- switched receipt identifiers from plain SHA-256 to HMAC-SHA256 via `AGENTMEMORY_RECEIPT_HMAC_KEY`, so predictable audit/target/user ids are not dictionary-guessable in shared receipts
- documented the key requirement for `receipt=true`
- hardened `safeDetailKeys()` for malformed/null/non-object legacy details

Checks run locally:

```bash
npm run build
npm test -- --run test/audit.test.ts
npm test
git diff --check
```

caioribeiroclw-pixel:
Followed up on the latest review in `e9f89ed`:\n\n- restore `AGENTMEMORY_RECEIPT_HMAC_KEY` after audit tests so receipt tests don't leak process-wide env state\n- add explicit `auditIdHash` assertions and ensure the raw audit id is not exported\n- add a malformed legacy `details: null` regression so receipt generation stays safe for old/bad audit rows\n\nChecks run locally:\n\n```bash\nnpm run build\nnpm test -- --run test/audit.test.ts\nnpm test\ngit diff --check\n```

caioribeiroclw-pixel:
Followed up on the latest review in `a55bee9` by adding an explicit missing-HMAC-key regression for `buildAuditReceipt()`. This locks the contract that `receipt=true` requires `AGENTMEMORY_RECEIPT_HMAC_KEY` instead of silently emitting guessable hashes.\n\nChecks run locally:\n\n```bash\nnpm run build\nnpm test -- --run test/audit.test.ts\nnpm test\ngit diff --check\n```

caioribeiroclw-pixel:
Thanks, this is useful.

I tightened this PR in the opposite direction from plain SHA-256: predictable audit/user/target ids now use `HMAC-SHA256` via `AGENTMEMORY_RECEIPT_HMAC_KEY`, and `receipt=true` fails closed if the key is missing. That keeps the receipt shareable without making ids dictionary-guessable.

I agree with your broader point on canonicalization/signatures. My read is:

- for this PR, the minimal contract is privacy-safe audit evidence over existing rows, not third-party non-repudiation yet
- JCS + `canon_version` becomes important once receivers need byte-stable verification across runtimes
- Ed25519/signature chains are probably the next layer after the maintainer decides receipts belong in this surface at all

So I’m intentionally not expanding scope here, but I’m going to track canonicalization/signing as the next step if this lands or if maintainers ask for stronger verification. Nobulex looks relevant for that layer — especially the receipt chain / tamper-evidence side.

arian-gogani:
Nice — HMAC-SHA256 with `AGENTMEMORY_RECEIPT_HMAC_KEY` is the stronger primitive here. Plain SHA-256 over predictable inputs (user_id, target_id) leaks enumeration via dictionary attack; HMAC fixes that without changing the audit-trail shape. Fail-closed when the key is missing is also the right default for security-relevant work.

One small thing worth considering for the next iteration: the HMAC scheme makes receipts unforgeable for someone who doesn't hold the key, but it also means receipts can't be verified by an outside party (auditor, regulator, counterparty) without trusting the operator to disclose the key. That's a reasonable tradeoff if the threat model is "internal tamper detection" — but if the receipts ever need to be independently verifiable (e.g. EU AI Act Article 12 audits), a hybrid approach works:

- Internal predictable IDs → HMAC (your current design, keep as-is for privacy)
- External tamper-evidence → Ed25519-signed envelope over the receipt (the operator publishes their public key, anyone can verify)

The two compose: HMAC for the private content addresses, Ed25519 for the public attestation. Lets you keep the dictionary-attack resistance while making the receipt portable.

For reference, the same JCS-canonical-preimage pattern is what we use in nobulex (just published `pip install nobulex` with the SDK), and we've verified byte-identical results across Python + JS on a published vector set. Happy to share fixtures if useful for cross-impl testing.

Either way — looks like a solid PR. The fail-closed semantics alone make it much more deployable than the typical "audit logging" pattern.

PR #653, opened 2026-05-25
Fix vector index Buffer serialization

NikolaiL:
## Summary

Fix vector index serialization/deserialization so persisted `Float32Array` embeddings preserve their intended byte range instead of using the entire backing `ArrayBuffer`.

## Why

`Buffer.from(arr.buffer)` serializes the entire backing buffer, ignoring `Float32Array.byteOffset` and `byteLength`. Similarly, `new Float32Array(Buffer.from(b64, "base64").buffer)` can read the entire pooled Buffer backing store instead of just the decoded bytes.

In a live AgentMemory instance this caused a rebuilt 384-dimension local embedding index to be persisted, then rejected on restart as 2048-dimensional vectors:

```text
Refusing to start: persisted vector index has 6471 of 6471 vectors with the wrong dimension.
Active provider (local) declares 384; dimensions seen on disk: 2048.
```

The service then refused to start until the installed package was patched locally or the vector index was discarded.

## Changes

- Serialize only the active `Float32Array` byte range.
- Deserialize using the decoded Buffer's `byteOffset` and `byteLength`.
- Add regression tests for pooled Buffer-backed deserialization and sliced Float32Array serialization.

## Verification

```bash
npx vitest run test/vector-index.test.ts test/index-persistence.test.ts
npm run build
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed vector embedding serialization and deserialization to correctly handle array slices, ensuring data integrity when working with vector indices.

* **Tests**
  * Added test cases to verify vector embedding serialization preserves dimensions and search functionality after deserialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/653?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/state/vector-index.ts
+++ b/src/state/vector-index.ts
@@ -1,9 +1,16 @@
 function float32ToBase64(arr: Float32Array): string {
-  return Buffer.from(arr.buffer).toString("base64");
+  return Buffer.from(arr.buffer, arr.byteOffset, arr.byteLength).toString(
+    "base64",
+  );
 }
 
 function base64ToFloat32(b64: string): Float32Array {
-  return new Float32Array(Buffer.from(b64, "base64").buffer);
+  const buf = Buffer.from(b64, "base64");
+  return new Float32Array(
+    buf.buffer,
+    buf.byteOffset,
+    buf.byteLength / Float32Array.BYTES_PER_ELEMENT,
+  );
 }
 
 function cosineSimilarity(a: Float32Array, b: Float32Array): number {
--- a/test/vector-index.test.ts
+++ b/test/vector-index.test.ts
@@ -71,6 +71,42 @@ describe("VectorIndex", () => {
     expect(results[0].sessionId).toBe("ses_1");
   });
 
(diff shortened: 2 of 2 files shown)


rohitg00 (approved):
Verified the fix against the live #455/#469 repro path. Float32Array.BYTES_PER_ELEMENT is the right semantic constant. Test exercises the exact pool-pollution case (small pooled Buffer, asserts buf.byteLength < buf.buffer.byteLength) which is the failure mode the issues describe.

Local test run on this branch: 11/11 vector-index tests pass.

@rohitg00 — recommend merging this one. #585 fixes the same root cause with the same approach; this one's marginally cleaner with the named constant. Suggest closing #585 with a thank-you to @kek-Sec; their thorough 5-vector round-trip test could land as a follow-up if useful.

rohitg00:
Thanks @NikolaiL — during bug-fix triage we landed an equivalent fix in #683. Closing this as superseded; the detailed write-up in your description was really helpful for confirming the symptom.

PR #530, opened 2026-05-19
fix(codex): canonicalize project identity for git worktrees

Lubrsy706:
## Summary
- canonicalize session project identity to the main git repo root on session start while preserving the actual cwd
- reuse the shared git worktree detector for branch-aware project metadata
- let context/search project matching include canonical project, cwd, worktree top-level, main repo root, and project-name aliases so existing worktree-scoped memory can still surface

Fixes #515

## Tests
- npm test -- test/project-identity.test.ts test/context-project-alias.test.ts
- npm test
- npm run build

## Summary by CodeRabbit

* **New Features**
  * Project identity is normalized and expanded into alias sets (including path-like aliases), improving session creation, search, and context selection.
  * Session creation and event handling now persist normalized project/cwd identity so memories saved in worktrees are recognized and reused across main-repo/worktree boundaries.
  * Search and context lookups now match sessions and lessons by expanded project aliases, producing more relevant results across related paths.

* **Tests**
  * Added tests validating project identity canonicalization, worktree handling, alias matching, and alias-aware context/search behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/530?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

Issue #515, opened 2026-05-18
Codex worktrees are treated as separate projects, fragmenting context and project-scoped memory


rockarxiv:
## Summary

When AgentMemory is used from Codex worktrees, the hook payloads use the worktree `cwd` as the project key. This causes each temporary worktree path to be treated as a separate project instead of the canonical repository.

As a result, context generation can miss project lessons, session history can be fragmented, and project-scoped features such as graph extraction, reflection/consolidation, profiles, and future memory retrieval may not treat worktree sessions as belonging to the same repository.

## Environment

- AgentMemory version: 0.9.18
- Client: Codex Desktop on Windows
- AgentMemory server: `http://localhost:3111`
- Viewer: `http://localhost:3113`
- Worktree example:

```text
C:\Users\<user>\.codex\worktrees\1a4b\my-repo
```

- Main repository example:

```text
C:\Projects\my-repo
```

## What happened

When a Codex session starts inside a worktree, the hook path sends payloads where `project` is derived from `cwd`.

For example, a session can be stored like:

```json
{
  "project": "C:\\Users\\<user>\\.codex\\worktrees\\1a4b\\my-repo",
  "cwd": "C:\\Users\\<user>\\.codex\\worktrees\\1a4b\\my-repo"
}
```

But existing project lessons, sessions, profiles, and memories may be associated with:

```text
C:\Projects\my-repo
```

or:

```text
my-repo
```

This means a new Codex worktree can fail to see project-relevant context, or can create new observations/sessions under a temporary worktree-specific project identity.

## Expected behavior

AgentMemory should separate:

```text
project = stable canonical repository identity
cwd     = actual current working directory
```

For Git worktrees, the stable project key should resolve to the main repository root, while preserving the actual worktree cwd.

Example expected session record:

```json
{
  "project": "C:\\Projects\\my-repo",
  "cwd": "C:\\Users\\<user>\\.codex\\worktrees\\1a4b\\my-repo"
}
```

Context lookup should also treat these as aliases for the same project:

```text
C:\Users\<user>\.codex\worktrees\1a4b\my-repo
C:\Projects\my-repo
my-repo
```

## Actual behavior

Worktree paths can be used as the project key directly. This fragments new sessions and context lookup by temporary worktree path.

## Impact

This affects Codex workflows that rely on temporary worktrees:

- startup context may miss relevant project lessons
- sessions from the same repo can appear under different projects
[...]

PR #717, opened 2026-05-29
_project.ts: resolve project from canonical remote.origin.url

### LINKED ISSUE #716: Resolve project identity from canonical repo URL, not directory basename
## Problem

After #687 made `basename(git rev-parse --show-toplevel)` the default project key, two failure modes remain:

1. **Basename collisions** — `~/work/utils` (`github.com/acme/utils`) and `~/clients/foo/utils` (`github.com/foo/utils`) both resolve to `utils` and silently share memory.
2. **Cross-machine fragmentation** — a single agentmemory pod serving multiple MCP clients sees the same repo as different projects when contributor checkouts differ in basename. Concrete shape: central pod at `/srv/projects/foo`, dev box at `/home/<user>/src/foo`, or simply `foo-prod` vs `foo` on a fork. `AGENTMEMORY_PROJECT_NAME` works around this but requires per-machine config for what is intrinsic to the repo.

Both fall away once the resolver consults `remote.origin.url` — the same identity `src/hooks/post-commit.ts:65` already uses for commit attribution.

## Proposal

Insert one step between the env override and the toplevel-basename fallback:

```
AGENTMEMORY_PROJECT_NAME
→ canonical(git config --get remote.origin.url)
→ basename(git rev-parse --show-toplevel)   ← #687, unchanged
→ basename(cwd)                              ← unchanged
```

`canonical` produces `host/owner/repo`:

| Input | Output |
|---|---|
| `https://github.com/acme/widgets.git` | `github.com/acme/widgets` |
| `git@github.com:Acme/Widgets.git` | `github.com/acme/widgets` |
| `ssh://git@host.example:2222/acme/widgets.git` | `host.example/acme/widgets` |
| `https://user:token@github.com/acme/widgets` | `github.com/acme/widgets` |

Lowercased end-to-end (host providers treat owner/repo case-insensitively; preserving path case would re-fragment same-repo clones). Trailing `.git` stripped, credentials stripped. Unparseable remotes fall through to basename.

## Why this and not #530

#530 proposed canonicalization plus multi-key matching across canonical/cwd/worktree-top/main-root/aliases. This issue keeps the single-key partitioning #687 just shipped — the change is *where the key comes from*, not how it's matched. No migration of existing data; deployments that already set `AGENTMEMORY_PROJECT_NAME` see no behavior change.

## Non-goals

- **Monorepo sub-package partitioning.** Every package under `github.com/acme/monorepo` resolves to the same key by design — partitioning within a monorepo is an orthogonal concern (the `project` axis isn't the right one for it). Operators who need per-package memory continue to set `AGENTMEMORY_PROJECT_NAME` per-package.
- **Multi-remote awareness (forks).** `origin` is sufficient for the two failure modes above. `AGENTMEMORY_PROJECT_NAME` remains the override for fork workflows that need to point at `upstream`.
- **SSH config aliases, `insteadOf` rewrites.** Resolver consumes the raw URL git emits; if a contributor configures an alias differently across machines, that's the same class of problem as setting different env vars and is solved by the same escape hatch.

## Performance

One `git config --get remote.origin.url` invocation, capped at 500ms (same budget the rest of `_project.ts` already uses). Same shape as #530's tightened lookup discussed by @rockarxiv on that PR.

## Patch shape

~60 LOC in `src/hooks/_project.ts`, 12 new unit tests, full suite (1303 tests) green. Happy to discuss the approach before opening the PR if there's a different shape you'd prefer; otherwise the PR will follow.

thekoma:
Closes #716.

Related: #687 (shipped current basename resolver — this extends it), #515 (worktree case, mostly side-effect-fixed by #687; remaining clone-path variant covered here), #530 (broader multi-key approach, closed in favor of fuller fix on #515 — this is the smaller single-key form).

## Change

Inserts canonical-URL resolution between the env override and the toplevel-basename fallback in `src/hooks/_project.ts`:

```
AGENTMEMORY_PROJECT_NAME
→ canonical(remote.origin.url)            ← new, host/owner/repo
→ basename(git rev-parse --show-toplevel) ← unchanged
→ basename(cwd)                            ← unchanged
```

Canonicalization: normalize https/ssh/SCP variants, strip credentials and `.git`, lowercase end-to-end (host providers treat owner/repo case-insensitively). Unparseable remotes fall through.

See #716 for the rationale, non-goals (monorepos, multi-remote), and precedent (`src/hooks/post-commit.ts:65` already uses the same identity source).

## Test plan

- New `test/hook-project-remote.test.ts` (12 tests): basename collision, multi-clone same-remote, SCP `git@host:path`, `ssh://` with port, embedded credentials, case-folding, `.git` stripping, env override still wins, no-remote fallback, unparseable-remote fallback, nested cwd.
- Existing `test/hook-project.test.ts` adjusted: basename-fallback tests now scope to remote-less temp repos, so they verify the fallback branch independently of the runner's own repo state.
- Tests use `mkdirSync(..., { recursive: true })` (cross-platform).
- Full suite: 1303/1303 pass.

## Compatibility

No data migration. Deployments using `AGENTMEMORY_PROJECT_NAME` see no change. Deployments relying on the basename default see their key change *only if* the repo has a configured `origin` remote — and in that case the new key is the more canonical one. Operators who want to pin the old behavior can set `AGENTMEMORY_PROJECT_NAME` to the prior basename.

## Summary by CodeRabbit

* **New Features**
  * Project identification is more flexible: honors an environment override, derives canonical project keys from Git remotes when available (normalizing formats), and falls back to repository or directory names for consistent project naming.

* **Tests**
  * Added and expanded tests covering remote canonicalization (HTTPS/SCP/SSH forms), env-var precedence, nested repos, basename fallbacks, and no-remote scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/717?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/plugin/scripts/notification.mjs
+++ b/plugin/scripts/notification.mjs
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-import { execSync } from "node:child_process";
+import { execFileSync } from "node:child_process";
 import { basename } from "node:path";
 
 //#region src/hooks/_project.ts
@@ -7,19 +7,69 @@ function resolveProject(cwd) {
 	const explicit = process.env["AGENTMEMORY_PROJECT_NAME"];
 	if (explicit && explicit.trim()) return explicit.trim();
 	const dir = cwd && cwd.trim() ? cwd : process.cwd();
+	const remote = readGitConfig(dir, "remote.origin.url");
+	if (remote) {
+		const canonical = canonicalizeRemoteUrl(remote);
+		if (canonical) return canonical;
+	}
+	const top = readGitToplevel(dir);
+	if (top) return basename(top);
+	return basename(dir);
+}
+function readGitConfig(cwd, key) {
 	try {
-		const top = execSync("git rev-parse --show-toplevel", {
-			cwd: dir,
+		return execFileSync("git", [
+			"config",
+			"--get",
+			key
+		], {
+			cwd,
 			stdio: [
 				"ignore",
 				"pipe",
 				"ignore"
 			],
 			timeout: 500
-		}).toString().trim();
-		if (top) return basename(top);
-	} catch {}
-	return basename(dir);
+		}).toString().trim() || void 0;
+	} catch {
+		return;
+	}
+}
+function readGitToplevel(cwd) {
+	try {
+		return execFileSync("git", ["rev-parse", "--show-toplevel"], {
+			cwd,
+			stdio: [
+				"ignore",
+				"pipe",
+				"ignore"
+			],
+			timeout: 500
+		}).toString().trim() || void 0;
+	} catch {
+		return;
+	}
+}
+function canonicalizeRemoteUrl(raw) {
+	const url = raw.trim();
+	if (!url) return void 0;
+	let host;
+	let path;
+	const scp = url.match(/^[^@\s/:]+@([^:\s/[\]]+):(.+)$/);
+	if (scp && !url.includes("://")) {
+		host = scp[1];
+		path = scp[2];
+	} else try {
+		const u = new URL(url);
+		host = u.hostname;
+		path = u.pathname;
+	} catch {
+		return;
+	}
+	if (!host || !path) return void 0;
+	path = path.replace(/^\/+/, "").replace(/\/+$/, "").replace(/\.git$/i, "");
+	if (!path) return void 0;
+	return `${host}/${path}`.toLowerCase();
 }
 
 //#endregion
--- a/plugin/scripts/post-tool-failure.mjs
+++ b/plugin/scripts/post-tool-failure.mjs
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-import { execSync } from "node:child_process";
+import { execFileSync } from "node:child_process";
 import { basename } from "node:path";
 
 //#region src/hooks/_project.ts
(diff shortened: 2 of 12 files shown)


thekoma:
## CodeRabbit feedback addressed

Picked up the review thread on `src/hooks/_project.ts:28-32`:

- **`readGitConfig` shell-interpolation:** switched `execSync(\`git config --get ${key}\`)` to `execFileSync("git", ["config", "--get", key], ...)`. Same change applied to `readGitToplevel` for consistency. `key` is currently only ever the literal `"remote.origin.url"` so this isn't an exploitable path today, but the pattern is a footgun for future callers and was flagged by OpenGrep — agreed worth fixing.
- **WHAT-style comments:** removed the `// Resolution order:` header (the if/return chain already reads as that order) and rewrote the `// SCP-style SSH` label as a WHY comment explaining the URL-parser fallthrough. Kept the WHY comment on full-key lowercasing.
- **Build artifacts:** rebuilt `plugin/scripts/*.mjs` so the shipped runtime tracks the TypeScript source. My initial commit on this branch missed the regenerated `.mjs` (PR #687 committed both `.ts` and `.mjs` together — followed the same pattern here).

No behavior change. Full suite still 1303/1303.

Commit: `f126548` on `fix/716-canonical-project-identity`.

thekoma:
## Round 2 — CodeRabbit nitpicks

### Applied

- **`basename()` in fallback assertions** (`test/hook-project.test.ts`, `test/hook-project-remote.test.ts`): replaced `dir.split("/").pop()` with `basename(dir)` from `node:path`. CodeRabbit flagged the two cases in `hook-project-remote.test.ts`; the same pattern appeared in four more places in `hook-project.test.ts`, so I fixed all six together. Tests now match `resolveProject`'s platform-aware behavior.

### Skipped with reason

- **WHAT-style comments on lines 63-64 and 84-86 of `_project.ts`**: the top-of-file "Resolution order" header was already removed in `f126548`. The two remaining comments are WHYs, not WHATs:
  - L63-64 (`// URL parser doesn't recognize SCP-style (user@host:path); match it first.`) — explains *why* the SCP regex runs before `new URL()`, a non-obvious dispatch order.
  - L84-86 (`// Host providers (GitHub, GitLab, Bitbucket) treat owner/repo as case-insensitive; lowercasing the full key prevents same-repo fragmentation when clones differ only in URL case.`) — explains the *rationale* for lowercasing the full key rather than just the host.

  Removing them would erase the rationale a future reader needs. Happy to revisit if there's a stronger convention here than the "no WHAT comments" rule I picked up from the static-analysis cue.

Full suite: 1303/1303. Commit: `040a0e0`.

PR #735, opened 2026-05-30
fix(health): gate memory_critical on real memory pressure

MarvinFS:
`memory_critical` fires whenever `heapUsed/heapTotal > 95% && rss > 512MB`. A busy Node process keeps its heap near-full by design, so that condition is effectively always true under normal load, producing chronic false-positive critical alerts that drown out real signals.

This adds a **real-pressure gate**: `memory_critical` now additionally requires a genuine pressure signal — either this process's absolute RSS has crossed a high ceiling (default 4 GB), or the host as a whole has run low on free RAM (default: < 10% of total free). The warn-level threshold is unchanged.

All bounds are env-overridable via `resolveThresholdConfig()`, which `monitor.ts` reads once at startup:

- `AGENTMEMORY_MEMORY_CRITICAL_RSS_MB` (default 4096)
- `AGENTMEMORY_MEMORY_SYSTEM_FREE_FLOOR_RATIO` (default 0.1; set to 0 to disable the system-memory check)
- existing knobs: `AGENTMEMORY_MEMORY_CRITICAL_PERCENT`, `AGENTMEMORY_MEMORY_WARN_PERCENT`, `AGENTMEMORY_MEMORY_RSS_FLOOR_MB`

Uses only `node:os`; no new dependencies. `test/health-thresholds.test.ts` is updated for the new gate. Verified `npm run build` clean and `npm test` green on a fresh `main` checkout.

## Summary by CodeRabbit

* **New Features**
  * Environment-variable overrides for memory threshold configuration (warning/critical levels) and system-free-memory ratio.
  * Health monitor now loads threshold config at startup and applies it when evaluating snapshots.

* **Bug Fixes**
  * More accurate critical memory detection by combining process memory and host free-memory checks.

* **Tests**
  * Threshold tests updated to cover stricter RSS/system-memory gating and mock restoration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rohitg00/agentmemory/pull/735?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/health/monitor.ts
+++ b/src/health/monitor.ts
@@ -2,12 +2,14 @@ import type { ISdk } from "iii-sdk";
 import type { HealthSnapshot } from "../types.js";
 import type { StateKV } from "../state/kv.js";
 import { KV } from "../state/schema.js";
-import { evaluateHealth } from "./thresholds.js";
+import { evaluateHealth, resolveThresholdConfig } from "./thresholds.js";
 
 export function registerHealthMonitor(
   sdk: ISdk,
   kv: StateKV,
 ): { stop: () => void } {
+  // Env overrides are read once at startup; thresholds don't change at runtime.
+  const thresholdConfig = resolveThresholdConfig();
   let connectionState = "connected";
   let prevCpuUsage = process.cpuUsage();
   let prevCpuTime = Date.now();
@@ -84,7 +86,7 @@ export function registerHealthMonitor(
       alerts: [],
     };
 
-    const evaluated = evaluateHealth(snapshot);
+    const evaluated = evaluateHealth(snapshot, thresholdConfig);
     snapshot.status = evaluated.status;
     snapshot.alerts = evaluated.alerts;
     snapshot.notes = evaluated.notes;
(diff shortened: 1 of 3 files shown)


MarvinFS:
Thanks for the review — addressed in d23ee8f:

- **Env override validation (the actionable item):** `parseIntEnv`/`parseFloatEnv` now take min/max bounds and fall back to the default when an override is out of range (percent 0–100, ratio 0–1, MB ≥ 0), so a typo like `...FREE_FLOOR_RATIO=2` or `...CRITICAL_RSS_MB=-1` can no longer make `underRealPressure` always true.
- **Helper extraction:** pulled the gate into `isUnderRealMemoryPressure(...)` and trimmed the explanatory comments.
- **Coverage:** added direct tests for the host-free-RAM gate (tripping path and the `ratio=0` disable case, via mocked `os.totalmem`/`os.freemem`) and for the env-bounds rejection.

Full suite green (1294 tests) on a fresh `main` checkout.

PR #743, opened 2026-05-31
feat: load secrets from systemd credentials ($CREDENTIALS_DIRECTORY)

MarvinFS:
## Problem

Run agentmemory as a systemd service and the only ways to pass provider keys (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `AGENTMEMORY_SECRET`, ...) are `Environment=` / `EnvironmentFile=` - both place the secrets in the process environment, readable via `/proc/<pid>/environ` and plaintext at rest.

## Change

An opt-in loader, `loadSystemdCredentials()`, called once at the start of the worker `main()`. It copies any files in `$CREDENTIALS_DIRECTORY` into `process.env` (only when unset/empty; trailing newline trimmed; file name maps 1:1 to the env var). A value assigned to `process.env` after `exec` is readable in-process but is not written back to the exec environ block, so it does not appear in `/proc/<pid>/environ` (verified on Node 22 / glibc). The providers get the keys; the environment stays clean.

## Requirements & platform support

Purely additive and **a no-op when `$CREDENTIALS_DIRECTORY` is unset**, so non-systemd hosts (macOS, Windows, Alpine/OpenRC) and existing `Environment=`/`.env` setups are entirely unaffected.

- Encrypted at rest: `LoadCredentialEncrypted=` + `systemd-creds` - **systemd >= 250** (Debian 12+, RHEL/Rocky/Alma 9+, Fedora, Arch, openSUSE, Ubuntu 24.04+).
- Out-of-environment but plaintext at rest: `LoadCredential=` - **systemd >= 247** (e.g. Ubuntu 22.04).
- systemd < 247 or no systemd: loader no-ops; use the existing env approach.

## Enabling it (operator, systemd >= 250)

```bash
systemd-creds encrypt --name=ANTHROPIC_API_KEY anthropic.key \
  /etc/credstore.encrypted/ANTHROPIC_API_KEY.cred
```

```ini
# unit / drop-in
[Service]
LoadCredentialEncrypted=ANTHROPIC_API_KEY:/etc/credstore.encrypted/ANTHROPIC_API_KEY.cred
```

systemd decrypts into a private tmpfs (`$CREDENTIALS_DIRECTORY`, mode 0400) at start; the loader reads it from there.

## Compatibility

Explicitly set env vars always win. Unit tests cover populate/trim, no-overwrite, fill-empty, unset/missing dir, and ignoring subdirs/empty files.

## Summary by CodeRabbit

* **New Features**
  * App now automatically loads credentials delivered by systemd at startup, populating environment variables for downstream configuration while preserving existing non-empty values.

* **Tests**
  * Added comprehensive tests covering credential loading behaviors (trimming, no-ops for missing dirs, ignoring subdirectories, and preserving/restoring environment state).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,6 +13,7 @@ import {
   isContextInjectionEnabled,
   isDropStaleIndexEnabled,
 } from "./config.js";
+import { loadSystemdCredentials } from "./load-credentials.js";
 import {
   createProvider,
   createFallbackProvider,
@@ -157,6 +158,10 @@ process.on("unhandledRejection", (reason) => {
 });
 
 async function main() {
+  // Pull any systemd-delivered secrets ($CREDENTIALS_DIRECTORY) into the
+  // environment before config/providers read keys; a value set after exec
+  // stays out of /proc/<pid>/environ. No-op when the dir is unset.
+  loadSystemdCredentials();
   const config = loadConfig();
   const embeddingConfig = loadEmbeddingConfig();
   const fallbackConfig = loadFallbackConfig();
--- a/src/load-credentials.ts
+++ b/src/load-credentials.ts
@@ -1,0 +1,49 @@
+import { existsSync, readFileSync, readdirSync, statSync } from "node:fs";
+import { join } from "node:path";
+
+/**
+ * Bridge systemd-delivered credentials into process.env at runtime.
+ *
+ * When the unit declares `LoadCredential(Encrypted)=NAME:...`, systemd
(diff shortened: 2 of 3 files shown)


MarvinFS:
Addressed the cleanup feedback in 558b003: the `afterEach` loop plus `Object.assign(savedEnv)` already restores the environment exactly, so the redundant unconditional `delete process.env["CREDENTIALS_DIRECTORY"]` (which could clobber a value the runner had pre-set) is removed.

@coderabbitai review

PR #764, opened 2026-06-01
fix(index-persistence): shard persisted search snapshots

### LINKED ISSUE #309: Architecture: Migrate BM25 & Graph Search from In-Memory to SQLite
The agentmemory project currently relies on 3 powerful search mechanisms to build its Hybrid Search and contextual reasoning engine:

1.Vector Search
2.BM25 Search
3.Graph Search

but the problem is that the system performs search indexing and traversal completely in-memory. This creates many critical bottlenecks like 
cpu bound operations ,RAM exhaustion ,etc.
i saw this exact same bottleneck in Vector Search, and for that, I created the recent pull request to use sqlite-vec.

## The Proposed Solution:
So, for the remaining BM25 Search and Graph Search, we should use SQLite storage instead of in-memory storage.
1. BM25 
Current: Builds a massive Inverted Index Map in Node.js RAM.

Proposed: Enable SQLite's native FTS5 extension. When memory text is saved, we insert it into an FTS5 virtual table. We simply query: SELECT obs_id FROM bm25_index WHERE text MATCH 'query' ORDER BY bm25(bm25_index) LIMIT 20.

2. Graph Search via SQLite CTEs
current: this.kv.list() loads all nodes and edges into RAM to run BFS loops in JavaScript.

Proposed: Mirror nodes and edges into dedicated graph_nodes and graph_edges SQLite tables. Instead of JS loops, we send a WITH RECURSIVE (CTE) query directly to SQLite.

this will dramatically improve performance and solve the current bottlenecks. The main data remains safely in the KV store, but offloading the search logic to SQLite

### LINKED ISSUE #762: Stopgap for #309: shard persisted BM25/vector snapshots to avoid oversized state values
Note: This issue description was drafted with AI assistance. I personally experienced the underlying issue. If any AI-slop-like shortcomings remain, they are due to my prompting, and I welcome suggestions on how to communicate my requirements more clearly to Codex GPT-5.5 xhigh.

## Existing related discussion

This is a focused follow-up to #309, not a standalone rediscovery. #309 is a closed architecture issue about moving BM25 and graph search away from in-memory indexing. Its comments also report the same current `IndexPersistence.save()` bottleneck: large search snapshots are persisted through oversized `state::set` values, #204 made those timeouts nonfatal, and the data-loss / stale-snapshot problem remains. One production report in #309 specifically identifies the vector snapshot being written as one `kv.set(KV.bm25Index, "vectors", vectorIndex.serialize())` value and proposes chunking as an interim workaround.

This issue narrows that broader closed architecture discussion to a current-repo stopgap: keep the existing BM25/vector index implementations, but persist their serialized snapshots as bounded shards so large corpora do not depend on one very large state value.

This is related to, but not the same as, #204 / #209. Those fixed unhandled `state::set` rejection crashes. They did not change the size of the persisted values or the publication protocol for rebuilt search snapshots.

This also should not be read as superseding #300. That PR is an open sqlite-vec vector-backend direction, opt-in through `VECTOR_BACKEND=sqlite-vec`; as checked immediately before filing on 2026-06-01 at 22:56 UTC, it is still unmerged and reports `mergeable_state=dirty`. If #300 lands, it may reduce vector pressure for that opt-in backend, but it still does not make the current BM25 snapshot persistence bounded, and it does not by itself remove the need for safe legacy/default snapshot publication semantics in `IndexPersistence`.

## Current behavior

`IndexPersistence.save()` persists the complete BM25 snapshot and the complete vector snapshot as two monolithic strings in the `mem:index:bm25` state scope:

- `SearchIndex.serialize()` returns one JSON string for all BM25 entries, inverted terms, document term counts, and total document length.
- `VectorIndex.serialize()` returns one JSON string for all vector entries, including base64-encoded embedding data.
- `IndexPersistence.save()` writes those strings with `kv.set(KV.bm25Index, "data", ...)` and `kv.set(KV.bm25Index, "vectors", ...)`.

Current upstream source at `fd9e3bd`: [`src/state/index-persistence.ts` lines 31-40](https://github.com/rohitg00/agentmemory/blob/fd9e3bd42d6208a33f0ee9de1442fdbb60eab106/src/state/index-persistence.ts#L31-L40), [`src/state/search-index.ts` lines 194-210](https://github.com/rohitg00/agentmemory/blob/fd9e3bd42d6208a33f0ee9de1442fdbb60eab106/src/state/search-index.ts#L194-L210), [`src/state/vector-index.ts` lines 124-136](https://github.com/rohitg00/agentmemory/blob/fd9e3bd42d6208a33f0ee9de1442fdbb60eab106/src/state/vector-index.ts#L124-L136), and [`src/state/schema.ts` lines 3-13](https://github.com/rohitg00/agentmemory/blob/fd9e3bd42d6208a33f0ee9de1442fdbb60eab106/src/state/schema.ts#L3-L13).

For a large corpus, a rebuild can finish in memory but fail while saving either serialized snapshot through state persistence. The next process start then has no durable rebuilt snapshot, or it keeps using the last older snapshot that was saved successfully. The evidence is split deliberately: upstream source shows that both BM25 and vector snapshots are saved as single state values; #309 includes a broad large-corpus timeout report and a concrete vector snapshot report where the serialized `vectors` value reached about 27 MB and timed out.

## Expected behavior

Large rebuilt BM25 and vector snapshots should persist without requiring one very large state value per index. A failed save attempt should not corrupt or delete the last successfully published snapshot for that index.

## Steps to reproduce

The proposed regression test gives a deterministic reproduction without any private data or external service:

1. Wrap the `StateKV` used by `IndexPersistence` so `set()` rejects any string value over a small limit.
2. Build a BM25 index and vector index whose serialized forms are larger than that limit.
3. Call `IndexPersistence.save()`, then `IndexPersistence.load()`.

On current upstream behavior, the save path sends the whole BM25 snapshot as one `data` string and the whole vector snapshot as one `vectors` string, so the guarded KV rejects the oversized write. With the proposed fix, each string is split into bounded shards, the save succeeds, and the loaded BM25/vector indexes round-trip.

The same size pattern appears from the current source shape: BM25 JSON grows with entries, terms, and postings, while vector JSON grows with vector count and embedding dimensions. The strongest detailed public size report is vector-specific; BM25 is included here because it has the same monolithic persistence boundary and the same failure mode once one `data` value exceeds what the state layer can reliably commit. The reproduction only needs to prove the storage-granularity bug, so it uses generated test observations and generated vectors.

## Root cause

The current persistence boundary is the whole serialized index string. `KV.bm25Index` is the shared state scope for search snapshots, and `IndexPersistence.save()` writes the complete serialized BM25 and vector indexes as single values under that scope. That makes save reliability depend on the maximum practical size of one state value instead of on the logical index contents.

## Suggested fix

Persist BM25 and vector snapshots as bounded string shards:

- write all shards for a new generation before publishing it;
- publish a compact manifest in `mem:index:bm25` after shard writes succeed;
- keep loading legacy monolithic `data` and `vectors` values when no manifest exists;
- delete partial new shards if a save fails before manifest publication;
- leave new shards in place if manifest publication is attempted but reports failure, because the manifest may have committed;
- delete old shards and legacy keys only after the new manifest is published;
- fail closed for that index if a published manifest points to a missing shard or a shard with the wrong length.

Rokurolize:
Note: This PR description was drafted with AI assistance. I personally experienced the underlying issue. If any AI-slop-like shortcomings remain, they are due to my prompting, and I welcome suggestions on how to communicate my requirements more clearly to Codex GPT-5.5 xhigh.

Fixes #762.

This PR does not close #309 because that issue is a broader, already-closed SQLite/FTS5 architecture discussion. It implements the narrower chunked-persistence stopgap discussed there for the existing `IndexPersistence` path.

It also does not supersede #300. #300 is an open sqlite-vec vector-backend direction, opt-in through `VECTOR_BACKEND=sqlite-vec`; as checked immediately before filing on 2026-06-01 at 22:56 UTC, it is still unmerged and reports `mergeable_state=dirty`. If #300 lands, it may reduce vector pressure for that opt-in backend, but it does not make BM25 snapshot persistence bounded and does not remove the need for safe legacy/default snapshot publication semantics in `IndexPersistence`.

## Summary

Large BM25/vector search snapshots are currently saved as monolithic strings in the `mem:index:bm25` state scope. On large corpora, that turns rebuild persistence into one very large state value for BM25 and one for vectors. The evidence is intentionally separated: upstream source shows both snapshots are monolithic; #309 contains a broad large-corpus persistence-timeout report after #204 made timeout failures nonfatal, plus a concrete vector report where the serialized `vectors` value reached about 27 MB and timed out. This PR saves each snapshot as bounded shards and publishes a compact manifest only after all shards for that snapshot are written.

## What was happening

The current upstream save path is:

- `SearchIndex.serialize()` returns one JSON string containing all BM25 entries, inverted terms, document term counts, and total document length.
- `VectorIndex.serialize()` returns one JSON string containing all vector entries and base64-encoded embedding data.
- `IndexPersistence.save()` writes those two strings with `kv.set(KV.bm25Index, "data", ...)` and `kv.set(KV.bm25Index, "vectors", ...)`.
- `KV.bm25Index` is the `mem:index:bm25` state scope.

Current upstream source at `fd9e3bd`: [`src/state/index-persistence.ts` lines 31-40](https://github.com/rohitg00/agentmemory/blob/fd9e3bd42d6208a33f0ee9de1442fdbb60eab106/src/state/index-persistence.ts#L31-L40), [`src/state/search-index.ts` lines 194-210](https://github.com/rohitg00/agentmemory/blob/fd9e3bd42d6208a33f0ee9de1442fdbb60eab106/src/state/search-index.ts#L194-L210), [`src/state/vector-index.ts` lines 124-136](https://github.com/rohitg00/agentmemory/blob/fd9e3bd42d6208a33f0ee9de1442fdbb60eab106/src/state/vector-index.ts#L124-L136), and [`src/state/schema.ts` lines 3-13](https://github.com/rohitg00/agentmemory/blob/fd9e3bd42d6208a33f0ee9de1442fdbb60eab106/src/state/schema.ts#L3-L13).

As the corpus grows, those two strings can become large enough that persistence fails even though the in-memory rebuild completed. Since `save()` catches and logs persistence failures, the process keeps running, but the rebuilt snapshot is not durably published. A later start can therefore load no rebuilt snapshot or the last older snapshot that was saved successfully.

That is the narrower problem this PR addresses. It is related to #204 / #209, but those changes only prevented `state::set` timeout rejections from crashing the process; they did not reduce persisted value size or make search snapshot publication robust for larger serialized indexes.

The strongest detailed public size evidence is vector-specific. BM25 is included because current source gives it the same single-value persistence boundary, its JSON grows with indexed entries and postings, and the guarded-KV regression test proves the storage-granularity failure for both generated snapshots. If maintainers prefer a narrower first step, the same approach can be split into vector-memory-backend sharding first, with BM25 snapshot sharding as a follow-up.

## Fix

- Split BM25 and vector snapshots into bounded string shards.
- Write shards under generation-specific shard scopes before changing the active manifest.
- Publish the new snapshot with a compact manifest that records exact shard scopes and character counts.
- Load legacy monolithic snapshots when no manifest exists.
- Load any valid manifest that records exact shard scopes, even if optional metadata is absent.
- Delete partial new shards if a save fails before the manifest is published.
- Leave new shards in place if the manifest write reports failure after it was attempted, because state persistence may have committed the manifest before returning an error.
- Delete legacy keys and old shards only after the new manifest is published; cleanup failure is nonfatal.
- Return null and log a warning if a published shard is missing or has a length mismatch.

BM25 and vector snapshots are still saved independently, matching the previous save order. If vector saving fails after BM25 is published, the new BM25 snapshot remains active and the previous vector snapshot remains active.

## Reproduction and tests

The regression test `avoids one oversized state::set string payload for persisted indexes` wraps the `StateKV` used by `IndexPersistence` so any string value over a small limit is rejected. It then saves generated BM25/vector indexes whose serialized forms are larger than that limit.

That test reproduces the storage-granularity bug on current upstream behavior because `save()` tries to write one oversized `data` string and one oversized `vectors` string. With this PR, the same logical indexes are written as bounded shards and load back successfully.

Additional coverage:

- `saves BM25 index shards outside the BM25 metadata scope`
- `saves vector index shards outside the BM25 scope`
- `loads legacy monolithic BM25 and vector snapshots`
- `loads sharded manifests that omit optional generation metadata`
- `keeps the previous generation when a shard write fails before manifest commit`
- `keeps the previous generation when manifest set rejects before commit`
- `keeps a generation loadable when manifest set commits before rejecting`
- `deletes a shard that committed before set rejected`
- `loads the new generation even when old generation cleanup fails`
- `keeps the previous vector generation when vector save fails after BM25 publish`
- `fails closed when a manifest shard is missing`
- `fails closed when a manifest shard length mismatches`

Verification:

- `npm test -- test/index-persistence.test.ts`
- `npm run build`
- `HOME=/tmp/agentmemory-pr-verify-home-20260601 npm test` on `origin/main` plus this two-file patch

## Review notes

Please review the manifest publication order and failure handling carefully. The intended invariant is that failed writes do not corrupt or delete the last successfully published snapshot for that index, while invalid published manifests fail closed instead of silently loading partial or stale data.

## Summary by CodeRabbit

* **Improvements**
  * Index persistence now uses sharded storage with per-index manifests for more reliable, resumable saves; safer commit/rollback, validation, and legacy snapshot support prevent partial or corrupted restores. Cleared vector snapshots persist as empty entries to avoid unintended reloads. Audit entries include a new operation type for index persistence.

* **Tests**
  * Expanded coverage for sharded persistence, manifest/rollback failure modes, load validation, vector behavior, and save/stop scheduling.

### DIFF
--- a/src/state/index-persistence.ts
+++ b/src/state/index-persistence.ts
@@ -179,6 +179,7 @@ export class IndexPersistence {
       this.options.createGeneration?.() ?? createIndexGeneration();
     const chunkChars = shardChars(this.options);
     const shards: IndexShardManifest["shards"] = [];
+    const chunks: string[] = [];
 
     for (let offset = 0; offset < serialized.length; offset += chunkChars) {
       const shardIndex = shards.length;
@@ -188,21 +189,30 @@ export class IndexPersistence {
       )}`;
       const chunk = serialized.slice(offset, offset + chunkChars);
       shards.push({ scope, key: INDEX_SHARD_KEY, chars: chunk.length });
-      try {
-        await this.kv.set(scope, INDEX_SHARD_KEY, chunk);
+      chunks.push(chunk);
+    }
+
+    const writeResults = await Promise.allSettled(
+      shards.map(async (shard, index) => {
+        const chunk = chunks[index] ?? "";
+        await this.kv.set(shard.scope, shard.key, chunk);
         await this.auditIndexPersistence("shard_write", [
-          statePath(scope, INDEX_SHARD_KEY),
+          statePath(shard.scope, shard.key),
         ], {
-          scope,
-          key: INDEX_SHARD_KEY,
+          scope: shard.scope,
+          key: shard.key,
           manifestKey,
           generation,
           chars: chunk.length,
         });
-      } catch (err) {
-        await this.deleteShards(shards, "shard_write_rollback");
-        throw err;
-      }
+      }),
+    );
+    const failedWrite = writeResults.find(
+      (result): result is PromiseRejectedResult => result.status === "rejected",
+    );
+    if (failedWrite) {
+      await this.deleteShards(shards, "shard_write_rollback");
+      throw failedWrite.reason;
     }
 
     const nextManifest: IndexShardManifest = {
--- a/src/state/index-persistence.ts
+++ b/src/state/index-persistence.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from "node:crypto";
 import { SearchIndex } from "./search-index.js";
 import { VectorIndex } from "./vector-index.js";
 import type { StateKV } from "./kv.js";
@@ -6,6 +7,39 @@ import { logger } from "../logger.js";
 
 const DEBOUNCE_MS = 5000;
 const FAILURE_LOG_THROTTLE_MS = 60_000;
+const BM25_KEY = "data";
+const BM25_MANIFEST_KEY = "data:manifest";
+const BM25_SHARD_SCOPE_PREFIX = `${KV.bm25Index}:bm25:`;
+const VECTOR_KEY = "vectors";
+const VECTOR_MANIFEST_KEY = "vectors:manifest";
+const VECTOR_SHARD_SCOPE_PREFIX = `${KV.bm25Index}:vectors:`;
+const INDEX_SHARD_KEY = "data";
+const DEFAULT_INDEX_SHARD_CHARS = 2_000_000;
+
+type IndexShardManifest = {
+  v: 1;
+  generation?: string;
+  shards: Array<{ scope: string; key: string; chars: number }>;
+  chars: number;
+};
+
+type IndexPersistenceOptions = {
+  shardChars?: number;
+  createGeneration?: () => string;
+};
+
+function shardChars(options: IndexPersistenceOptions): number {
+  const configured = options.shardChars;
+  return typeof configured === "number" &&
+    Number.isFinite(configured) &&
+    configured > 0
+    ? Math.floor(configured)
+    : DEFAULT_INDEX_SHARD_CHARS;
+}
+
+function createIndexGeneration(): string {
+  return `${Date.now().toString(36)}-${randomUUID().replace(/-/g, "")}`;
+}
 
 export class IndexPersistence {
   private timer: ReturnType<typeof setTimeout> | null = null;
@@ -15,6 +49,7 @@ export class IndexPersistence {
     private kv: StateKV,
     private bm25: SearchIndex,
     private vector: VectorIndex | null,
+    private options: IndexPersistenceOptions = {},
   ) {}
 
   scheduleSave(): void {
@@ -34,9 +69,9 @@ export class IndexPersistence {
       this.timer = null;
     }
     try {
-      await this.kv.set(KV.bm25Index, "data", this.bm25.serialize());
+      await this.saveBm25Index(this.bm25.serialize());
       if (this.vector && this.vector.size > 0) {
-        await this.kv.set(KV.bm25Index, "vectors", this.vector.serialize());
+        await this.saveVectorIndex(this.vector.serialize());
       }
     } catch (err) {
       this.logFailure(err);
@@ -50,16 +85,12 @@ export class IndexPersistence {
     let bm25: SearchIndex | null = null;
     let vector: VectorIndex | null = null;
 
-    const bm25Data = await this.kv
-      .get<string>(KV.bm25Index, "data")
-      .catch(() => null);
+    const bm25Data = await this.loadBm25Data();
     if (bm25Data && typeof bm25Data === "string") {
       bm25 = SearchIndex.deserialize(bm25Data);
     }
 
-    const vecData = await this.kv
-      .get<string>(KV.bm25Index, "vectors")
-      .catch(() => null);
+    const vecData = await this.loadVectorData();
     if (vecData && typeof vecData === "string") {
       vector = VectorIndex.deserialize(vecData);
     }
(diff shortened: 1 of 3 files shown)


rohitg00:
Merged. Thanks @Rokurolize — sharded persistence + manifest commit/rollback semantics are exactly what #762 needed, and the regression coverage (oversized-payload guard, manifest-commits-after-error, length-mismatch fail-closed) is the highest-quality test addition the persistence path has seen.

PR #776, opened 2026-06-02
fix(replay): key existing-session re-write on parsed.sessionId, not existing.id

### LINKED ISSUE #775: import-jsonl aborts the entire batch ("missing field `key`") — existing-session branch re-keys on `existing.id` instead of `parsed.sessionId`
## What happens

Running `import-jsonl` over a real `~/.claude/projects` tree aborts the **entire** batch with an opaque engine error — no files import, no partial progress:

```
invocation_failed: deserialization_error: Failed to deserialize input for state::set: missing field `key`
```

## Root cause

In the `import-jsonl` handler (`src/functions/replay.ts`), the existing-session branch re-keys the write on `existing.id`:

```ts
const existing = await kv.get<Session>(KV.sessions, parsed.sessionId);
if (existing) {
  // ...mutate observationCount / endedAt / status / tags / firstPrompt...
  await kv.set(KV.sessions, existing.id, existing);   // <-- re-keys on existing.id
}
```

If the stored record has no `id` field, `existing.id` is `undefined`. `kv.set(scope, undefined, value)` serializes to a `state::set` payload **without** a `key` (JSON.stringify drops `undefined`), so the engine rejects it with `missing field \`key\`` — and because that rejection aborts the whole function invocation, the **entire import dies**, not just that session.

## Fix

Re-key on the value the code already fetched with — `parsed.sessionId` (always defined; `parseJsonlText` has a three-level fallback). It's both safer and a redundant-indirection cleanup:

```ts
await kv.set(KV.sessions, parsed.sessionId, existing);
```

Verified on my install: this one-line change clears the error and the import proceeds. PR incoming.

## Environment

- agentmemory `0.9.24`, iii-sdk `0.11.2`
- Node `v22.18.0`, macOS 15 (Darwin 25.3)

Brentbin:
Fixes #775.

## Problem

`import-jsonl` aborts the **entire** batch with `deserialization_error: ... missing field \`key\`` when a stored session record has no `id`. The existing-session branch re-keys the write on `existing.id`, so an `undefined` id drops `key` from the `state::set` payload and the engine rejects the whole invocation — no partial progress, opaque error.

## Change

One line in `src/functions/replay.ts`: re-key on `parsed.sessionId` — the key the record was just fetched with, always defined via `parseJsonlText`'s three-level fallback — instead of `existing.id`. This mirrors the adjacent `else` branch, which already keys new sessions on `parsed.sessionId`.

## Verification

- `npm run build` compiles clean.
- On my install, this cleared the `missing field \`key\`` abort and the import proceeded.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where importing JSONL files containing data for existing sessions would fail. Sessions now update successfully during the import process.

### DIFF
--- a/src/functions/replay.ts
+++ b/src/functions/replay.ts
@@ -406,7 +406,10 @@ export function registerReplayFunctions(sdk: ISdk, kv: StateKV): void {
           if (!existing.firstPrompt && firstPrompt) {
             existing.firstPrompt = firstPrompt;
           }
-          await kv.set(KV.sessions, existing.id, existing);
+          // Re-key on parsed.sessionId (the key we fetched with), not existing.id:
+          // a stored record missing `id` makes existing.id undefined, which drops the
+          // `key` from the state::set payload and aborts the entire import.
+          await kv.set(KV.sessions, parsed.sessionId, existing);
         } else {
           const session: Session = {
             id: parsed.sessionId,


rohitg00:
Thanks @Brentbin — shipped this exact fix in v0.9.25 ([release notes](https://github.com/rohitg00/agentmemory/releases/tag/v0.9.25), [PR #791](https://github.com/rohitg00/agentmemory/pull/791)) bundled with two other v0.9.24 regressions. Your PR identified the root cause + the correct re-key target; the merged fix matches your patch and additionally backfills the missing existing.id on legacy rows. Apologies for not catching this PR before bundling — the credit on the diagnosis is yours.

PR #777, opened 2026-06-02
fix: preserve import session key for existing jsonl sessions

### LINKED ISSUE #775: import-jsonl aborts the entire batch ("missing field `key`") — existing-session branch re-keys on `existing.id` instead of `parsed.sessionId`

mturac:
## Summary
- update existing imported sessions using the parsed session id instead of `existing.id`
- keep malformed stored session records from re-keying `state::set` with an undefined key
- add a regression test for importing a JSONL file when the existing session record has no id

Closes #775

## Tests
- `npm test -- test/replay.test.ts`
- `npm run build`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed session persistence during JSONL imports so existing sessions are correctly updated and finalized.

* **Tests**
  * Expanded tests for JSONL session import using filesystem-backed fixtures and in-memory mocks; verifies session updates (observation count, completed status, and import tag).

### DIFF
--- a/src/functions/replay.ts
+++ b/src/functions/replay.ts
@@ -393,6 +393,7 @@ export function registerReplayFunctions(sdk: ISdk, kv: StateKV): void {
 
         const existing = await kv.get<Session>(KV.sessions, parsed.sessionId);
         if (existing) {
+          if (!existing.id) existing.id = parsed.sessionId;
           existing.observationCount =
             (existing.observationCount || 0) + parsed.observations.length;
           if (parsed.endedAt > (existing.endedAt || "")) {
@@ -406,7 +407,7 @@ export function registerReplayFunctions(sdk: ISdk, kv: StateKV): void {
           if (!existing.firstPrompt && firstPrompt) {
             existing.firstPrompt = firstPrompt;
           }
-          await kv.set(KV.sessions, existing.id, existing);
+          await kv.set(KV.sessions, parsed.sessionId, existing);
         } else {
           const session: Session = {
             id: parsed.sessionId,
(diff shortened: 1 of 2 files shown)


rohitg00:
Thanks @mturac — same fix shipped in v0.9.25 (see [PR #791](https://github.com/rohitg00/agentmemory/pull/791) + [release notes](https://github.com/rohitg00/agentmemory/releases/tag/v0.9.25)). #776 + this PR + my own diagnosis all converged on the same root cause; credit goes to whoever filed first. Closing because the issue is resolved on main.

PR #785, opened 2026-06-02
fix: recover fenced summary XML responses

### LINKED ISSUE #783: parseSummaryXml: regex parser fails on markdown code fences and extra text around XML, no retry on final summarize
## Summary

The `parseSummaryXml` function in `src/functions/summarize.ts` uses simple regex to extract XML tags from the LLM response. Many LLMs (DeepSeek, GPT, and others) sometimes wrap structured output in markdown code fences (` ```xml ... ``` `) or include conversational text before/after the XML. The simple regex parser doesn't handle these cases, causing session summaries to be silently dropped.

Additionally, the **final** summarization call has no retry mechanism, unlike the chunk-level `summarizeChunkWithRetry` which retries twice.

## Symptoms

```
[agentmemory] warn Failed to parse summary XML {"sessionId":"ses_..."}
```

The `mem::summarize` function returns `{success: false, error: "parse_failed"}` and the session is never summarized.

## Root Cause (v0.9.24)

**`src/functions/summarize.ts`** — `parseSummaryXml` (around line 148):

```ts
function parseSummaryXml(xml, sessionId, project, obsCount) {
    const title = getXmlTag(xml, "title");
    if (!title) return null;
    // ...
}
```

**`src/functions/summarize.ts`** — `getXmlTag`:

```ts
function getXmlTag(xml, tag) {
    if (!VALID_TAG.test(tag)) return "";
    const match = xml.match(new RegExp(`<${tag}>([\\s\\S]*?)</${tag}>`));
    return match ? match[1].trim() : "";
}
```

Two problems:

1. **No preprocessing of raw LLM output** — If the LLM wraps the XML in markdown fences (` ```xml\n<summary>...\n``` `), adds text before/after, or otherwise deviates from the exact XML format, the regex misses the tags and `parseSummaryXml` returns `null`.

2. **No retry on final summarize** — The chunk-level `summarizeChunkWithRetry` retries twice on parse failure (line 155: `for (let attempt = 1; attempt <= 2; attempt++)`), but the `mem::summarize` function calls `parseSummaryXml` **once with no retry** (around line 183). This is inconsistent — the chunk level has resilience but the final merge doesn't.

## Steps to Reproduce

1. Have a session that triggers auto-compress (`AGENTMEMORY_AUTO_COMPRESS=true`)
2. Use any LLM provider that sometimes wraps XML in markdown code fences (observed with `deepseek-v4-flash` via OpenAI-compatible endpoint, but also known to happen with GPT/Claude)
3. The warning appears when the LLM returns something like:

````
Here's the summary:
```xml
<summary>
  <title>My session</title>
  ...
</summary>
```
````

Instead of the expected exact XML with no surrounding text or fences.

## Suggested Fix

Two changes in `src/functions/summarize.ts`:

### 1. Make `parseSummaryXml` robust to markdown fences and extra text

Before the regex extraction, strip markdown code fences and try to extract the raw XML from any surrounding text:

```ts
function parseSummaryXml(xml, sessionId, project, obsCount) {
    // Strip markdown code fences if present (```xml ... ``` or ``` ... ```)
    xml = xml.replace(/```xml\n?/gi, "").replace(/```\n?/g, "").trim();
    // If extra text exists before/after the XML root element, extract just the XML
    const rootMatch = xml.match(/(<[a-zA-Z_][a-zA-Z0-9_-]*>[\s\S]*<\/[a-zA-Z_][a-zA-Z0-9_-]*>)/);
    if (rootMatch) xml = rootMatch[1].trim();
    const title = getXmlTag(xml, "title");
    if (!title) return null;
    // ...
}
```

### 2. Add retry loop to the final summarize call

Match the retry behavior already present in `summarizeChunkWithRetry`:

```ts
let summary = null;
for (let attempt = 1; attempt <= 2; attempt++) {
    summary = parseSummaryXml(response, sessionId, session.project, compressed.length);
    if (summary) break;
    logger.warn("Failed to parse summary XML", { sessionId, attempt });
    if (attempt === 1) {
        const retryResult = await produceSummaryXml(provider, compressed, sessionId, session.project);
        response = retryResult.response;
        if (!response || !response.trim()) break;
    }
}
```

## Impact

Without the fix, every session summarization that hits a fence-wrapping LLM response silently drops the session summary. This cascades: no session summary means no crystal digest, no episodic→semantic→procedural consolidation, and gaps in the memory graph.

In our environment (agentmemory v0.9.24, deepseek-v4-flash via OpenAI-compatible endpoint), this occurred **31 times across 15+ sessions** over a few days of normal usage before we patched it.

## Workaround

Users can set `SUMMARIZE_CHUNK_SIZE=999999` to force single-chunk mode (which avoids the chunk concurrency but doesn't fix the fence parsing), or apply the patch above to `dist/index.mjs` in their global installation.

However, a fix upstream would benefit everyone regardless of LLM provider.

mturac:
Fixes #783.

Summary:
- extract the first fenced or embedded <summary>...</summary> block before parsing
- retry the final summary/reduce response once when XML parsing fails
- add regressions for fenced output and final-response retry

Validation:
- npm test -- test/summarize.test.ts
- npm run build

## Summary by CodeRabbit

* **Bug Fixes**
  * Correctly extract summaries even when wrapped in fenced XML or embedded in surrounding text.
  * Retry final summary generation once on parse failure or transient LLM errors; logs a warning and returns the last raw response if parsing still fails.

* **Tests**
  * Added tests covering fenced-XML extraction, retry behavior for failing/parsing responses, and persistence of parsed summary fields.

### DIFF
--- a/src/functions/summarize.ts
+++ b/src/functions/summarize.ts
@@ -35,6 +35,14 @@ const CHUNK_CONCURRENCY_DEFAULT = 6;
 // Bail on the merged summary if more than this fraction of chunks fail
 // to parse — a half-blind narrative is worse than a clean error.
 const MAX_SKIP_RATIO = 0.5;
+const SUMMARY_BLOCK_RE = /<summary\b[^>]*>[\s\S]*?<\/summary>/i;
+
+function normalizeSummaryXml(raw: string): string {
+  const fenced = raw.match(/```(?:xml)?\s*([\s\S]*?)```/i);
+  const candidate = fenced?.[1] ?? raw;
+  const summaryBlock = candidate.match(SUMMARY_BLOCK_RE);
+  return (summaryBlock?.[0] ?? candidate).trim();
+}
 
 function getChunkSize(): number {
   const raw = process.env.SUMMARIZE_CHUNK_SIZE;
@@ -90,6 +98,41 @@ async function summarizeChunkWithRetry(
   return null;
 }
 
+async function summarizeFinalXmlWithRetry(
+  provider: MemoryProvider,
+  system: string,
+  prompt: string,
+  context: {
+    sessionId: string;
+    project: string;
+    observationCount: number;
+    mode: "single" | "chunked";
+    chunks: number;
+  },
+): Promise<string> {
+  let response = "";
+  for (let attempt = 1; attempt <= 2; attempt++) {
+    response = await provider.summarize(system, prompt);
+    if (
+      parseSummaryXml(
+        response,
+        context.sessionId,
+        context.project,
+        context.observationCount,
+      )
+    ) {
+      return response;
+    }
+    logger.warn("Final summary parse failed", {
+      sessionId: context.sessionId,
+      attempt,
+      mode: context.mode,
+      chunks: context.chunks,
+    });
+  }
+  return response;
+}
+
 // Returns the final summary XML string. For sessions ≤ chunk size, this is
 // a single LLM call (legacy behavior). For larger sessions, observations
 // are split into chunks processed in parallel batches, each chunk retried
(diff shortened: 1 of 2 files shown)


rohitg00:
Thanks @mturac — fenced summary XML recovery shipped in v0.9.25 ([PR #791](https://github.com/rohitg00/agentmemory/pull/791) + [release notes](https://github.com/rohitg00/agentmemory/releases/tag/v0.9.25)). Implementation matches your approach (strip markdown fences + retry final parse). Closing because the fix is on main.

PR #788, opened 2026-06-02
fix: use fallback provider default models

### LINKED ISSUE #778: Bug: FALLBACK_PROVIDERS passes the primary provider's model to fallback providers → cross-provider failover always 404s and trips the circuit breaker (v0.9.24)
## Summary

`FALLBACK_PROVIDERS` is unusable across providers that use different model names. When the primary provider fails (e.g. OpenAI returns `429 insufficient_quota`), the fallback provider is constructed with the **primary provider's model name**, so the fallback call hits the wrong API with an invalid model and 404s. The repeated 404s then trip the circuit breaker, which blocks *all* subsequent LLM operations — even though a perfectly healthy fallback provider is configured.

Net effect: configuring `FALLBACK_PROVIDERS=gemini` to survive an OpenAI outage makes things **worse**, not better — the system goes from "primary down" to "primary down + circuit breaker open".

**Version:** 0.9.24 (global npm install)

## Root cause

`createFallbackProvider` copies the primary provider's `model` into every fallback provider's config:

```js
function createFallbackProvider(config, fallbackConfig) {
  if (fallbackConfig.providers.length === 0) return createProvider(config);
  const providers = [createBaseProvider(config)];
  for (const providerType of fallbackConfig.providers) {
    if (providerType === config.provider) continue;
    try {
      const fbConfig = {
        provider: providerType,
        model: config.model,      // <-- BUG: primary provider's model (e.g. "gpt-4o-mini")
        maxTokens: config.maxTokens
      };
      providers.push(createBaseProvider(fbConfig));
    } catch {}
  }
  ...
}
```

`createBaseProvider` then constructs e.g. `GeminiProvider(geminiKey, config.model, ...)` with `config.model === "gpt-4o-mini"`. The fallback provider never resolves its **own** configured/default model (`GEMINI_MODEL`, `ANTHROPIC_MODEL`, etc.).

## Reproduction

1. Set an OpenAI key whose quota is exhausted (or otherwise make the primary fail):
   ```
   OPENAI_API_KEY=sk-...            # primary LLM (default model gpt-4o-mini)
   GEMINI_API_KEY=...
   GEMINI_MODEL=gemini-2.5-flash
   FALLBACK_PROVIDERS=gemini
   AGENTMEMORY_AUTO_COMPRESS=true
   ```
2. Restart the daemon and let observations flow (so compression runs).
3. The OpenAI call 429s, failover to Gemini fires, and Gemini is called with model `gpt-4o-mini`.

## Observed logs

```
error Compression failed {"obsId":"...","error":"gemini API error (404): [{
  \"error\": { \"code\": 404,
    \"message\": \"models/gpt-4o-mini is not found for API version v1main, or is not supported for generateContent...\",
    \"status\": \"NOT_FOUND\" } }]"}
error Compression failed {"obsId":"...","error":"circuit_breaker_open"}
error Compression failed {"obsId":"...","error":"circuit_breaker_open"}
error Summarize failed {"sessionId":"...","error":"circuit_breaker_open"}
...
```

The same applies to `mem::summarize`, graph extraction, consolidation — any LLM path routed through the fallback chain.

## Impact

- Cross-provider failover (the headline use case for `FALLBACK_PROVIDERS`) never works when the primary and fallback use different model namespaces (OpenAI ↔ Gemini ↔ Anthropic).
- Because the failed fallback attempts count toward the circuit breaker, a configured fallback actively makes availability worse: the breaker opens and blocks LLM ops that would otherwise degrade gracefully to synthetic compression.

## Suggested fix

When building each fallback provider, resolve that provider's own model instead of inheriting the primary's:

- Drop `model: config.model` from `fbConfig` and let `createBaseProvider` fall back to the provider-specific default/env (`GEMINI_MODEL`, `ANTHROPIC_MODEL`, `MINIMAX_MODEL`, `OPENROUTER_MODEL`), or
- Add per-provider model resolution in `createBaseProvider` so a missing/foreign `config.model` is replaced with the provider's default.

Optionally: don't count a `404 NOT_FOUND` model-mismatch (a config error, not a transient failure) toward the circuit breaker.

## Workaround

Make the desired provider the **primary** (so its own model is used) rather than a fallback. With a dead OpenAI key, that means removing/commenting `OPENAI_API_KEY` so provider detection selects Gemini directly — then `GEMINI_MODEL` is honored and LLM ops succeed.

## Environment

- agentmemory 0.9.24, global npm install
- macOS (Apple Silicon), Node via Homebrew
- Primary: OpenAI (default `gpt-4o-mini`); Fallback: `FALLBACK_PROVIDERS=gemini`, `GEMINI_MODEL=gemini-2.5-flash`

mturac:
## Summary
- build fallback provider configs from each fallback provider's own model defaults/env overrides
- keep the primary provider model from leaking into cross-provider fallback requests
- add a regression covering OpenAI -> Gemini failover request bodies

Fixes #778

## Validation
- `git diff --cached --check`
- `npm test -- test/provider-fallback-model.test.ts`
- `npm run build`

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for fallback provider behavior and model configuration across multiple providers.

* **Refactor**
  * Fallback provider selection now respects per-provider environment variable defaults for model and base URL configuration.

### DIFF
--- a/src/providers/index.ts
+++ b/src/providers/index.ts
@@ -25,6 +25,54 @@ function requireEnvVar(key: string): string {
   return value;
 }
 
+function providerDefaultConfig(
+  provider: FallbackConfig["providers"][number],
+  maxTokens: number,
+): ProviderConfig {
+  switch (provider) {
+    case "openai":
+      return {
+        provider,
+        model: getEnvVar("OPENAI_MODEL") || "gpt-4o-mini",
+        maxTokens,
+        baseURL: getEnvVar("OPENAI_BASE_URL"),
+      };
+    case "minimax":
+      return {
+        provider,
+        model: getEnvVar("MINIMAX_MODEL") || "MiniMax-M2.7",
+        maxTokens,
+      };
+    case "anthropic":
+      return {
+        provider,
+        model: getEnvVar("ANTHROPIC_MODEL") || "claude-sonnet-4-20250514",
+        maxTokens,
+        baseURL: getEnvVar("ANTHROPIC_BASE_URL"),
+      };
+    case "gemini":
+      return {
+        provider,
+        model: getEnvVar("GEMINI_MODEL") || "gemini-2.5-flash",
+        maxTokens,
+      };
+    case "openrouter":
+      return {
+        provider,
+        model:
+          getEnvVar("OPENROUTER_MODEL") ||
+          "anthropic/claude-sonnet-4-20250514",
+        maxTokens,
+      };
+    case "agent-sdk":
+      return {
+        provider,
+        model: "agent-sdk",
+        maxTokens,
+      };
+  }
+}
+
 export function createProvider(config: ProviderConfig): ResilientProvider {
   return new ResilientProvider(createBaseProvider(config));
 }
@@ -41,11 +89,7 @@ export function createFallbackProvider(
   for (const providerType of fallbackConfig.providers) {
     if (providerType === config.provider) continue;
     try {
-      const fbConfig: ProviderConfig = {
-        provider: providerType,
-        model: config.model,
-        maxTokens: config.maxTokens,
-      };
+      const fbConfig = providerDefaultConfig(providerType, config.maxTokens);
       providers.push(createBaseProvider(fbConfig));
     } catch {
       // skip unavailable fallback providers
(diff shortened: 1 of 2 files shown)


rohitg00:
Thanks @mturac — fallback provider default-model resolution shipped in v0.9.25 ([PR #791](https://github.com/rohitg00/agentmemory/pull/791) + [release notes](https://github.com/rohitg00/agentmemory/releases/tag/v0.9.25)). Same approach: each fallback resolves its own env-driven default (OPENAI_MODEL / GEMINI_MODEL / ANTHROPIC_MODEL / MINIMAX_MODEL / OPENROUTER_MODEL). Closing because the fix is on main.

PR #821, opened 2026-06-04
fix(scope): enforce AGENT_ID isolation on mem::search (closes #817)

### LINKED ISSUE #817: AGENT_ID isolated scope not enforced on mem::search (REST /search, memory_recall, recall_context) → cross-agent memory leak
### Summary

With `AGENTMEMORY_AGENT_SCOPE=isolated`, recall is correctly filtered on `mem::smart-search`, `/memories`, `/observations`, `/sessions` (per CHANGELOG, PR #654). However **`mem::search` applies no agent-scope filtering at all** — and it backs the REST `POST /agentmemory/search` endpoint *and* the MCP tools **`memory_recall`** and **`recall_context`**. An isolated worker for agent B can therefore read memories tagged for agent A.

### Environment

v0.9.26, local engine (no API key, BM25-only), Linux, Node 25.

### Repro

1. Start: `AGENT_ID=agent_a AGENTMEMORY_AGENT_SCOPE=isolated npx @agentmemory/agentmemory`
2. `POST /agentmemory/remember {"content":"SECRET_MARKER private to A"}` → stored with `agentId=agent_a`
3. Restart against the **same data dir** with `AGENT_ID=agent_b AGENTMEMORY_AGENT_SCOPE=isolated`
4. `POST /agentmemory/smart-search {"query":"SECRET_MARKER"}` → `0 results` ✅ (correctly isolated)
5. `POST /agentmemory/search {"query":"SECRET_MARKER"}` → **returns A's memory** ❌ (leak)
6. `memory_recall` (MCP) → same leak, since it calls `mem::search`

### Root cause

- `src/functions/search.ts` (`mem::search`): no `getAgentId()` / `isAgentScopeIsolated()` usage.
- `src/triggers/api.ts` `api::search` (~L364–432): builds payload without `agentId`.
- `src/mcp/server.ts` `memory_recall` (~L115) and `recall_context` (~L1640) both call `function_id: "mem::search"`; `recall_context` additionally `kv.list(KV.memories)` filtered only by `isLatest`.
- CHANGELOG (PR #654) enumerates the filtered paths and omits `mem::search`.

### Impact

`memory_recall` is a primary recall tool, so the isolation guarantee is bypassed in normal agent usage — not just via a raw REST call. Anyone relying on `isolated` for multi-tenant separation is exposed.

### Suggested fix

Make `mem::search` honor `isAgentScopeIsolated()` / `getAgentId()` and filter results (mirroring `mem::smart-search`), and have `api::search` / `memory_recall` / `recall_context` propagate `agentId` + `?agentId` query param consistently with the other recall paths.

wyh0626:
## Summary

Closes #817. `AGENTMEMORY_AGENT_SCOPE=isolated` filtered recall on `mem::smart-search`, `/memories`, `/observations`, and `/sessions` (PR #654), but **`mem::search` applied no agent-scope filtering at all**. It backs `POST /agentmemory/search` plus the MCP tools `memory_recall` and `recall_context`, so an isolated worker for agent B could read agent A's memories. `memory_recall` is a primary recall tool, so this leaked in normal agent usage, not just via a raw REST call.

Root cause: isolation in #654 (commit `1aec56a`) was added per-function — `smart-search.ts` and the list endpoints got `getAgentId()` / `isAgentScopeIsolated()`, but `src/functions/search.ts` was never touched. The #654 CHANGELOG enumerates the filtered paths and omits `mem::search`, matching the code.

## Fix

`mem::search` now mirrors `mem::smart-search`. The BM25 index doesn't carry `agentId`, so it over-fetches and post-filters on the loaded observation:

```ts
const isolated = isAgentScopeIsolated()
const explicitAgentId =
  typeof data.agentId === 'string' && data.agentId.trim().length > 0
    ? data.agentId.trim() : undefined
const filterAgentId = explicitAgentId === '*'
  ? undefined                                       // opt out of env default
  : explicitAgentId ?? (isolated ? getAgentId() : undefined)
```

- `api::search` forwards `agentId` — body first, then `?agentId=` — consistent with `/memories`, `/observations`, `/sessions`. When omitted, `mem::search` falls back to the env `AGENT_ID`.
- `memory_recall` / `recall_context` forward `agentId`. `recall_context` also scopes its direct `kv.list(KV.memories)` pass, which previously filtered only by `isLatest` and leaked across agents.
- `memoryToObservation()` now copies `agentId`. Memories saved via `/remember` are indexed through this helper; without it, isolated recall dropped the owner's *own* remembered memories — exactly the `/remember` → `/search` path in the #817 repro.

## Verify

Reproducer from #817:

```bash
# agent A writes
AGENT_ID=agent_a AGENTMEMORY_AGENT_SCOPE=isolated npx @agentmemory/agentmemory
curl -sX POST localhost:3111/agentmemory/remember -d '{"content":"SECRET_MARKER private to A"}'

# agent B, same data dir
AGENT_ID=agent_b AGENTMEMORY_AGENT_SCOPE=isolated npx @agentmemory/agentmemory
curl -sX POST localhost:3111/agentmemory/search -d '{"query":"SECRET_MARKER"}'
# before: returns A's memory
# after:  0 results
```

Unit:

```bash
npx vitest run test/search-agent-scope.test.ts
# 6/6 pass
npm run build
# Build complete
```

I also reverted each source change and confirmed the new tests fail on the pre-fix code (3 leak cases + the `/remember` Memory case), so they actually catch the regression.

## Tests

6 cases in `test/search-agent-scope.test.ts`:

- shared mode (default) returns hits from every agent
- isolated mode returns only the env `AGENT_ID`'s hits
- isolated mode does not leak another agent's memory
- explicit `agentId` override wins over the env default
- wildcard `agentId: "*"` opts out of isolation
- isolated mode scopes memories saved via `KV.memories` (the `/remember` path through `memoryToObservation`)

`test/cross-project-isolation.test.ts`: its partial `config.js` mock gained the now-imported `isAgentScopeIsolated` export.

## Known limitations / follow-ups

- **Over-fetch false negatives.** Like `mem::smart-search`, agent-filtered recall over-fetches BM25 hits then post-filters. With a very small `limit` and a large out-of-scope head, an in-scope match beyond the fetch window can be missed — a false negative, **not a leak**. A durable fix indexes `agentId` in BM25 or makes the over-fetch adaptive.
- **Per-call `agentId` override discoverability.** The override is wired at the function and REST-body level (matching `smart-search`), but isn't exposed in the MCP tool schema or forwarded by the standalone MCP proxy. This mirrors `smart-search`'s existing behavior; default env isolation is unaffected.

## Summary by CodeRabbit

* **Improvements**
  * Better agent-specific memory handling to improve multi-agent observation tracking and isolated memory search.
  * Added inline documentation to clarify behavior and filtering implications for saved memories.

### DIFF
--- a/src/state/memory-utils.ts
+++ b/src/state/memory-utils.ts
@@ -20,5 +20,11 @@ export function memoryToObservation(memory: Memory): CompressedObservation {
     concepts: memory.concepts,
     files: memory.files,
     importance: memory.strength,
+    // #817: carry agentId through so agent-scope isolation in
+    // mem::search applies to memories too. Without this, a memory
+    // saved via /remember loads with agentId=undefined and gets
+    // dropped by the isolated-mode filter — hiding an agent's own
+    // remembered memories from itself.
+    agentId: memory.agentId,
   };
 }


wyh0626:
Hi @rohitg00 — rebased onto latest main and trimmed this down to a one-liner, since the main #817 leak is already fixed in [#849](https://github.com/rohitg00/agentmemory/issues/849).

That fix filters on obs.agentId !== filterAgentId, but memoryToObservation() never copies agentId. So /remember-saved memories load with agentId === undefined and get dropped in isolated mode — meaning an agent can't recall its own memories.

```
// src/state/memory-utils.ts
    importance: memory.strength,
+   agentId: memory.agentId,

Issue #733, opened 2026-05-30
Project identity uses git-toplevel basename — collides across same-named repos, no remote-based identity


devon3000:
> **Edited to correct the original report.** My first version (written from the compiled `dist/` bundle) claimed project identity was the full filesystem path. Reading the actual source shows it's the **git-toplevel basename**, which is more correct than I first stated but still has real failure modes. Corrected analysis below. *(Second edit: sharpened the Impact section with the exact source-level scoping behavior — `memory_smart_search` is global, but `mem::context` and the project profile are hard-scoped by exact project string.)*

## Summary

Project scope is derived from the **git-toplevel basename** (falling back to the cwd basename). This means:

- **Collision:** two unrelated repos whose top-level directory share a name (e.g. both checked out as `app`, `server`, `frontend`, or two different forks both named `agentmemory`) collapse into one project bucket.
- **No cross-checkout-name unification:** the *same* repo checked out under different directory names — or under different absolute paths on two machines — resolves to *different* projects.
- There is no use of the git remote, and no canonicalization to a globally stable identity.

(Full absolute-path `project` values that exist in older stores predate the basename logic — they're legacy rows from an earlier version, which is a separate migration concern.)

## Where

`resolveProject()` — `src/hooks/_project.ts`:

```ts
// Resolution order: AGENTMEMORY_PROJECT_NAME env → git toplevel basename → cwd basename.
export function resolveProject(cwd?: string): string {
  const explicit = process.env["AGENTMEMORY_PROJECT_NAME"];
  if (explicit && explicit.trim()) return explicit.trim();
  const dir = cwd && cwd.trim() ? cwd : process.cwd();
  try {
    const top = execSync("git rev-parse --show-toplevel", { cwd: dir, /* … */ }).toString().trim();
    if (top) return basename(top);   // <-- basename only; "agentmemory", not the repo identity
  } catch {}
  return basename(dir);
}
```

This shared resolver is imported by all nine hooks (`session-start`, `post-tool-use`, `prompt-submit`, `subagent-*`, `notification`, `task-completed`, `pre-compact`, `post-tool-failure`). The server stores the value verbatim — `api::session::start` in `src/triggers/api.ts` persists `body.project` as-is and does not re-derive it. So identity is decided entirely client-side in `resolveProject`, which is the right single seam to fix.

## Impact

The `project` tag's effect is **path-dependent**, which is what makes this subtle. After reading the search/context functions:

[...]

PR #829, opened 2026-06-04
Add project isolation for agentmemory

Durocks:
## What changed
- Added `AGENTMEMORY_PROJECT_ISOLATION` to require project scoping for memory reads and writes when enabled.
- Updated MCP, REST, and core memory paths to preserve and enforce project scope.
- Fixed hybrid smart-search so scoped queries no longer leak cross-project results.
- Added regression tests for scoped save/recall/search behavior.

## Why
- The MCP client was dropping or not enforcing project scope in some paths.
- We want optional strict project isolation without breaking the default global-memory workflow.

## Behavior
- When `AGENTMEMORY_PROJECT_ISOLATION=true`, memory operations require a project scope.
- When the flag is off, existing global/unscoped behavior stays the same.

## Verification
- `npm run build`
- `npx vitest run test/project-isolation.test.ts test/smart-search.test.ts test/mcp-standalone-proxy.test.ts`
- Manual Zoo Code smoke tests for `memory_save`, `memory_recall`, and `memory_smart_search`
- Targeted automated coverage across scoped memory, smart-search, and MCP proxy behavior

## Summary by CodeRabbit

* **New Features**
  * Added project scoping configuration with `AGENTMEMORY_PROJECT_NAME` (default project) and `AGENTMEMORY_PROJECT_ISOLATION` (enforced isolation). Memory tools, CLI/MCP adapters, and API endpoints now accept/propagate project scope and include project info in scoped `search`/`smart-search` results.

* **Bug Fixes / Validation**
  * When isolation is enabled, API and MCP boundaries now reject missing/invalid project with a 400 error, and scoped recalls/searches ignore legacy unscoped data.

* **Documentation**
  * README updated with project-isolation defaults and multi-project setup guidance.

* **Tests**
  * Expanded coverage for isolation, scoped search/recall, and proxy/local fallback behavior (including persisting project).


thekoma:
I run agentmemory as one shared server instead of one per repo, so this default changes what the tool does for me. Numbers from my install, in case a real deployment is useful.

One server on 0.9.29 behind an ingress, four clients: three workstations and a Hermes agent in Kubernetes. 758 sessions carrying 56 distinct project labels, and 49 sessions with no project at all. I run it so a Claude session in one repo can recall what a different machine did in another repo last month.

That works today because reads ignore the label. I checked against my own server this morning: a fresh MCP shim started from one repo returns all 758 sessions, and `smart-search` drops a `project` key in the request body instead of filtering on it. The label records where an observation came from. It gates nothing.

Reading the diff, `isProjectIsolationEnabled()` in `src/config.ts` returns `true` when `AGENTMEMORY_PROJECT_ISOLATION` is unset or empty, and the connect adapters write `${AGENTMEMORY_PROJECT_ISOLATION:-true}` into the MCP config. After an upgrade each of my clients scopes itself to its repo folder name, and the test named `hides legacy unscoped memories from scoped reads when isolation is enabled` accounts for my 49 unlabelled sessions. Cross-repo recall stops.

The PR description says existing global behavior stays the same when the flag is off, which reads as opt-in. The code defaults it on. Those two should agree, whichever way the default lands.

I am not arguing against the feature. Someone running one checkout per project wants this, and #716 and #733 circle the same problem from another side. My ask is narrower: could the default be `false`, with `connect` writing `true` when it wires a new install? A fresh user gets isolation, an existing install keeps its memory visible until its owner opts in.

One more thing I ran into today. I set `AGENTMEMORY_PROJECT_NAME=hermes` on my Kubernetes agent so its sessions stop being labelled with a filesystem path. Under this PR that same variable becomes the isolation scope, so a naming fix turns into an isolation decision. Separating those two would help.

PR #842, opened 2026-06-06
feat(mcp): expose memory_forget tool so observations and sessions can actually be deleted

### LINKED ISSUE #833: /agentmemory:forget deletes nothing — memory_governance_delete only targets the memories store, not observations (reports success: true)
# `/agentmemory:forget` silently deletes nothing — `memory_governance_delete` only targets the `memories` store, not observations (and reports `success: true`)

## Summary

The `forget` skill (and the underlying `memory_governance_delete` MCP tool) **cannot delete observations**, which are the records `forget` is documented to remove. When given observation IDs, `memory_governance_delete` returns `{ "deleted": 0, "total": N, "success": true }` and removes nothing — while still reporting success. A user running `/agentmemory:forget <topic>` is told the operation succeeded when their data is still fully present.

## Versions

- `@agentmemory/agentmemory`: **0.9.26**
- `@agentmemory/mcp`: **0.9.26**
- Claude Code plugin `agentmemory`: **0.9.26**
- `iii` engine: **0.11.2**
- Platform: macOS (darwin 25.5.0)

## Root cause

`mem::governance-delete` only reads/writes the `KV.memories` store:

```js
// dist/src-…​.mjs — registerGovernanceFunction()
sdk.registerFunction("mem::governance-delete", async (data) => {
  ...
  let deleted = 0;
  for (const id of data.memoryIds)
    if (await kv.get(KV.memories, id)) {      // <-- ONLY the memories store
      await kv.delete(KV.memories, id);
      ...
      deleted++;
    }
  ...
  return { success: true, deleted, total: data.memoryIds.length };  // success even when deleted === 0
});
```

But observations are stored under a **different** keyspace, `mem:obs:<sessionId>` (`KV.observations(sessionId)`), one KV entry per session. Observation IDs (`obs_…`) never exist in `KV.memories`, so the `kv.get(KV.memories, id)` guard is always false and `deleted` stays `0`.

The `forget` SKILL.md compounds this by instructing the agent to pass observation IDs to `memory_governance_delete`, and explicitly (and incorrectly) states:

> "The standalone MCP doesn't accept a bare `sessionId` argument — it deletes by memory ID only."

So the documented happy path for `forget` is a guaranteed no-op for observation data.

A second, related issue: `memory_smart_search` did not surface the target observations at all (returned unrelated results / empty), so even discovering the correct IDs to delete was not possible through the documented flow.

## Steps to reproduce

1. Have one or more sessions with observations (e.g. from normal usage).
2. Run `/agentmemory:forget <term matching an observation>`.
3. The skill calls `memory_smart_search`, then `memory_governance_delete` with the observation IDs.
4. **Observed:** tool returns `{ "deleted": 0, "total": N, "success": true }`; `memory_sessions` still shows the same `observationCount`; the `.bin` files under `…/data/state_store.db/mem%3Aobs%3A<sessionId>.bin` are untouched.
5. **Expected:** the observations (and, when a whole session is targeted, the session record + summary) are removed, and the reported count reflects what was actually deleted.

## What actually works (suggested fix)

The engine already has the correct function — `mem::forget`, exposed over REST at **`POST /agentmemory/forget`**:

```js
sdk.registerFunction("api::forget", async (req) => {
  if (!req.body?.sessionId && !req.body?.memoryId)
    return { status_code: 400, body: { error: "sessionId or memoryId is required" } };
  return { status_code: 200, body: await sdk.trigger({ function_id: "mem::forget", payload: req.body }) };
});
```

`mem::forget` handles all three cases properly and keeps the BM25 + vector indexes and audit log consistent:

- `{ memoryId }` → delete a memory
- `{ sessionId, observationIds: [...] }` → delete specific observations in a session
- `{ sessionId }` (no `observationIds`) → delete **all** observations + the session record + summary

Verified working against 0.9.26: `POST /agentmemory/forget {"sessionId":"<id>"}` returned `{"deleted":3,"success":true}` (1 observation + session + summary) and the session disappeared from `memory_sessions`.

### Proposed changes

1. **Skill:** rewrite `skills/forget/SKILL.md` to call the `forget` route/function with `sessionId` (+ optional `observationIds`) instead of `memory_governance_delete` with observation IDs. Remove the incorrect "deletes by memory ID only" claim.
2. **MCP surface:** expose a `memory_forget` tool that maps to `mem::forget`, so MCP-only clients can delete observations/sessions (today only `memory_governance_delete` is exposed, which can't).
3. **Honesty:** have `mem::governance-delete` return `success: false` (or a `notFound` list) when `deleted === 0`, so callers don't report a no-op as a success.

## Residual-data note (`mem::forget` gap)

When deleting by `sessionId`, `mem::forget` clears `KV.observations`, `KV.sessions`, and `KV.summaries`, but leaves the per-session live-stream file `stream:mem-live:<sessionId>.bin` in `…/data/stream_store/` on disk. That file still contains the original prompt text after a "forget." Consider having `mem::forget` (and session deletion generally) also purge the corresponding `stream:mem-live:<sessionId>` entry so a forget fully removes the data at rest.

serhiizghama:
Fixes #833

## Problem

The `forget` skill (and any MCP-only client) has no way to delete observations or sessions:

- `memory_governance_delete` only reads/writes the `KV.memories` store. Observation IDs (`obs_*`) never exist there, so the skill's documented happy path is a guaranteed no-op.
- Worse, the tool returned `{ "deleted": 0, "success": true }` for that no-op, so users were told their data was removed while the `.bin` files under `mem:obs:<sessionId>` stayed fully intact.
- The engine already has the correct function — `mem::forget` handles memories, per-session observation lists, and whole-session deletion (record + summary) while keeping the BM25/vector indexes and audit log consistent — but it was only reachable over REST (`POST /agentmemory/forget`), not via MCP.

## Solution

1. **New `memory_forget` MCP tool** mapping to `mem::forget` (registry + `mcp::tools::call` case). Accepts `memoryId`, or `sessionId` with optional comma-separated/array `observationIds`; bare `sessionId` deletes all of the session's observations plus the session record and summary, matching the REST behavior. Validation mirrors the REST route (`sessionId or memoryId is required`) plus a specific error for `observationIds` without `sessionId`.
2. **`mem::forget` now returns its breakdown** (`memoriesDeleted`, `observationsDeleted`, `sessionDeleted`) — it already tracked these for the audit row; returning them lets callers confirm to the user what was actually removed.
3. **`mem::governance-delete` stops reporting no-ops as success**: IDs that don't exist in the memories store are returned in a `notFound` array and `success` is only `true` when every requested ID was deleted. The audit row records `notFound` too.
4. **`forget` skill + OpenCode instructions rewritten** to route observation/session deletion through `memory_forget` and keep `memory_governance_delete` for saved memories (`mem_*`), removing the incorrect "deletes by memory ID only" workaround that never worked.

Tool count badges/docs/manifests bumped 53 → 54 to keep `test/consistency.test.ts` green.

## Testing

- New `test/mcp-forget-tool.test.ts`: registry listing, validation errors, payload forwarding for all three deletion shapes (CSV + array `observationIds`), and the error path.
- `test/governance.test.ts`: non-existent IDs now expect `success: false` + `notFound`; added partial-delete and all-deleted cases.
- Full suite: 127 files, 1410 tests pass; `npm run build` clean.

## Summary by CodeRabbit

* **New Features**
  * Added memory forget functionality for deleting observations, sessions, or individual saved memories with explicit user confirmation.

* **Documentation**
  * Updated MCP tools count from 53 to 54 across all documentation and plugin manifests.
  * Clarified memory deletion workflows distinguishing between deleting saved memories versus observations/sessions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,7 +116,7 @@ Hook scripts in `src/hooks/` are standalone Node.js scripts (no iii-sdk import).
 
 ## Current Stats (v0.9.16)
 
-- 53 MCP tools (8 visible by default, `AGENTMEMORY_TOOLS=all` for all)
+- 54 MCP tools (8 visible by default, `AGENTMEMORY_TOOLS=all` for all)
 - 128 REST endpoints
 - 6 MCP resources, 3 MCP prompts
 - 12 hooks, 4 skills
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@
 <p align="center">
   <picture><source media="(prefers-color-scheme: dark)" srcset="assets/tags/light/stat-recall.svg"><img src="assets/tags/stat-recall.svg" alt="95.2% retrieval R@5" height="38" /></picture>
   <picture><source media="(prefers-color-scheme: dark)" srcset="assets/tags/light/stat-tokens.svg"><img src="assets/tags/stat-tokens.svg" alt="92% fewer tokens" height="38" /></picture>
-  <picture><source media="(prefers-color-scheme: dark)" srcset="assets/tags/light/stat-tools.svg"><img src="assets/tags/stat-tools.svg" alt="53 MCP tools" height="38" /></picture>
+  <picture><source media="(prefers-color-scheme: dark)" srcset="assets/tags/light/stat-tools.svg"><img src="assets/tags/stat-tools.svg" alt="54 MCP tools" height="38" /></picture>
   <picture><source media="(prefers-color-scheme: dark)" srcset="assets/tags/light/stat-hooks.svg"><img src="assets/tags/stat-hooks.svg" alt="12 auto hooks" height="38" /></picture>
   <picture><source media="(prefers-color-scheme: dark)" srcset="assets/tags/light/stat-deps.svg"><img src="assets/tags/stat-deps.svg" alt="0 external DBs" height="38" /></picture>
   <picture><source media="(prefers-color-scheme: dark)" srcset="assets/tags/light/stat-tests.svg"><img src="assets/tags/stat-tests.svg" alt="1,390+ tests passing" height="38" /></picture>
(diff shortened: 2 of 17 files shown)


serhiizghama:
Closing — it's gone stale and picked up conflicts in the meantime. Can reopen if it's still useful.

Issue #345, opened 2026-05-13
concept_edges table + BFS-depth-2 recall mode


rohitg00:
## Problem

Memories already carry a `concepts[]` field, populated at remember-time. Smart-search uses concepts as a recall signal but doesn't currently traverse concept relationships — if you ask "what do I know about auth," memories that mention `session-token` or `JWT` won't surface unless the query string also matches.

## Proposed approach

1. **Persist concept relationships explicitly.** New `concept_edges` row:

   ```ts
   { from: string, to: string, strength: number, lastSeenAt: ISODateString }
   ```

   Edges are derived at remember-time from co-occurrence within a single memory. Strength reinforces on repeated co-occurrence (same lifecycle pattern as memory `strength`).

2. **New recall mode `--graph` (default off).** When set, `smart-search` expands the query's matched concepts BFS-depth-2 and unions the resulting memory set into the candidate pool before reranking. Existing BM25 + vector hybrid stays the primary signal.

3. **Expose via REST**: `POST /agentmemory/smart-search` accepts `mode: "graph"`. Existing callers unaffected.

4. **Expose via MCP**: new tool `memory_graph_search` (separate from `memory_smart_search` so latency profile is opt-in).

5. **Viewer surface**: simple force-directed graph view of the top-N concept nodes per project. Behind feature flag.

## Acceptance

- `concept_edges` table populates as memories are remembered; edges decay if their concepts haven't co-occurred recently (re-use existing `reinforce` decay curve).
- `smart-search` recall improves measurably on the existing LongMemEval-S harness when `mode: "graph"` is set, without regressing baseline non-graph numbers.
- BFS depth bounded at 2; depth 3+ refused with `S400 depth_out_of_range`.
- Migration path: existing memories' `concepts[]` get backfilled into `concept_edges` on first run.

## Out of scope

- Full property graph (typed edges, edge attributes beyond strength).
- Cypher-style query language. Single API: `mode: "graph"`.
- Multi-hop reasoning over the LLM-compression layer — that's a separate concern.

Related: existing `concepts[]` field in `src/types.ts`.

Mahesh-4017:
1. New Table: concept_edges
CREATE TABLE concept_edges (
    from_concept TEXT NOT NULL,
    to_concept TEXT NOT NULL,
    strength FLOAT NOT NULL DEFAULT 1,
    last_seen_at TIMESTAMP NOT NULL,

    PRIMARY KEY (from_concept, to_concept)
);

[...]

PR #848, opened 2026-06-06
feat(memory): concept_edges derivation + boot backfill (#345 Phase 1)

serhiizghama:
## Motivation

Phase 1 of #345, following the phased plan you laid out there: the `concept_edges` table + derivation, nothing user-facing yet. Phases 2 (`mode: "graph"` on smart-search via the #463 `GraphRetrieval` traversal) and 3 (MCP tool + viewer) stay separate per the plan.

## What this changes

- **`KV.conceptEdges` scope** (`mem:concept-edges`) with the `ConceptEdge` row shape from the plan (`from`, `to`, `strength`, `lastSeenAt`, `reinforcements`, `createdAt`, `decayRate`). One row per unordered pair — key is the canonical `${a}|${b}` (lexicographic), so the Phase 2 adapter can treat the graph as undirected without dedup at read time.
- **Derivation** (`mem::concept-edges-derive`): walks `concepts[]`, emits every unordered pair, upserts under a keyed lock (the upsert is read-modify-write and can fire from compress and remember concurrently). Existing edges reinforce via the same saturation curve `reinforceLesson` uses — `strength += 0.1 * (1 - strength)` — and new edges start at 0.5 with `decayRate: 0.05`, mirroring the lessons lifecycle so the strength profile stays consistent across surfaces. Concepts are normalized (trim + lowercase) before pairing so `"JWT"` and `"jwt"` reinforce one edge instead of fragmenting the graph.
- **Hooks**: fire-and-forget (`TriggerAction.Void`) from `mem::compress` after the compressed observation is persisted, and from `mem::remember` after the memory row is saved — edge upkeep never blocks or fails the write itself. Both gated on `concepts.length >= 2`.
- **Boot migration** (`mem::concept-edges-backfill`): one-shot pass at startup — if `KV.conceptEdges` is empty and `KV.memories` has rows with 2+ concepts, backfill edges from existing memories. Completion is marked via `migrations:concept-edges-backfill` in `KV.state`; the function also short-circuits when edges already exist, so a crash between backfill and flag write can't double-reinforce on the next boot.

One small ripple: adding a second key to `StateScope` widened `StateScopeKey`, which broke `disk-size-manager`'s `const KEY: StateScopeKey = ...` pattern (the lookup type became `number | boolean`). Switched both call sites to `satisfies StateScopeKey` so the literal type — and the per-key value type — is preserved.

## Why this approach

Kept as close to the Phase 1 spec in #345 as possible: exact `ConceptEdge` shape, the lessons reinforce curve rather than a new one, and the `migrations:*` flag pattern in `KV.state`. No traversal code here — `GraphRetrieval`/`WeightedGraphTraversal` plumbing belongs to Phase 2 and this PR doesn't touch it.

## Testing

- `test/concept-edges.test.ts`: 14 cases — canonical pair keys, derivation, reinforcement (0.5 → 0.55 on second co-occurrence, saturation < 1.0 after 100 rounds), normalization/dedup, <2-concept no-ops, payload validation, backfill (flag set, idempotent re-run, edges-already-present short-circuit, legacy memories without `concepts`), and the `mem::remember` hook end to end.
- Full suite: 1414 passed (the only failure is `test/integration.test.ts`, which needs a running server and is excluded from `npm test`).
- `npm run build` passes.

## Summary by CodeRabbit

* **New Features**
  * Automatic concept relationship tracking: The system now derives and maintains relationships between concepts within memories, strengthening the knowledge graph for enhanced semantic connections.

* **Tests**
  * Added comprehensive test coverage for concept relationship features, including backfill and integration scenarios.

### DIFF
--- a/src/functions/compress.ts
+++ b/src/functions/compress.ts
@@ -192,6 +192,24 @@ export function registerCompressFunction(
           { kind: "observation", logId: compressed.id },
         );
 
+        // #345 Phase 1: derive concept co-occurrence edges from the
+        // freshly extracted concepts[]. Fire-and-forget — edge upkeep
+        // must never block or fail the compression itself.
+        if (compressed.concepts.length >= 2) {
+          try {
+            await sdk.trigger({
+              function_id: "mem::concept-edges-derive",
+              payload: { concepts: compressed.concepts },
+              action: TriggerAction.Void(),
+            });
+          } catch (err) {
+            logger.warn("Non-fatal concept-edge derivation failure", {
+              obsId: compressed.id,
+              error: err instanceof Error ? err.message : String(err),
+            });
+          }
+        }
+
         const streamResults = await Promise.allSettled([
           sdk.trigger({
             function_id: "stream::set",
--- a/src/functions/concept-edges.ts
+++ b/src/functions/concept-edges.ts
@@ -1,0 +1,140 @@
+import type { ISdk } from "iii-sdk";
+import type { StateKV } from "../state/kv.js";
+import { KV } from "../state/schema.js";
+import type { ConceptEdge, Memory, StateScope, StateScopeKey } from "../types.js";
+import { withKeyedLock } from "../state/keyed-mutex.js";
+import { logger } from "../logger.js";
+
+const MIGRATION_KEY = "migrations:concept-edges-backfill" satisfies StateScopeKey;
+
+/**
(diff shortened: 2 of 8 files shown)


serhiizghama:
Closing this out to keep things tidy — happy to revisit if the roadmap picks this up.

Issue #720, opened 2026-05-29
Token input cache rate dropped ~10pp due to per-turn system prompt mutation


HWliao:
# Token input cache rate dropped ~10pp due to per-turn system prompt mutation

**Severity**: Medium (performance degradation, not functional breakage)  
**Affected versions**: v0.9.20+ (opencode plugin)  
**Environment**: OpenCode + agentmemory MCP + opencode plugin

## Problem

Token input cache rate (prompt caching hit ratio) has dropped by approximately 10 percentage points when using agentmemory via the opencode plugin. The degradation is visible in daily aggregate statistics and accumulates over an entire day of usage.

## Root Cause

The opencode plugin's `experimental.chat.system.transform` hook injects the result of `/agentmemory/enrich` into the **system prompt** on every model request where files have been read or edited (`plugin/opencode/agentmemory-capture.ts:639-656`).

```typescript
// Current behavior — changes system prompt EVERY turn
"experimental.chat.system.transform": async (input, output) => {
  // ...
  const stash = stashFor(sid);
  if (stash.size === 0) return;
  const files = [...stash].slice(0, 10);
  const enrichResult = await postJson("/enrich", { sessionId: sid, files });
  output.system.push(enrichCtx);  // ❌ Mutates system prompt
}
```

### Why this breaks caching

LLM prompt caching (Anthropic-style ephemeral caching) works by matching **prefix blocks** byte-for-byte. The system prompt is the first block sent to the model. When it changes between requests, the entire prefix cache is invalidated:

```
Request 1: system = [base + instructions + context]                  → cached
Request 2: system = [base + instructions + context + enrich(file1)]  → DIFFERENT → cache miss
Request 3: system = [base + instructions + context + enrich(file2)]  → DIFFERENT → cache miss
Request 4: system = [base + instructions + context]                  → matches Request 1 → cache hit
Request 5: system = [base + instructions + context + enrich(file3)]  → DIFFERENT → cache miss
```

The `/agentmemory/enrich` endpoint (`src/functions/enrich.ts`) returns variable content based on the current file stash — it runs `mem::search` (semantic search with file names as queries), `mem::file-context` (cross-session file history), and bug memory matching. Since the agent touches different files each turn, the enrich output differs on every model request where tool calls occur. In practice, nearly every turn produces a tool call, so the cache miss rate is very high.

### Aggravating factors

[...]

PR #857, opened 2026-06-08
fix(opencode): resolve project name from multiple sources to fix Desktop mode

Chewji9875:
## Problem

On **OpenCode Desktop**, the `session.created` event carries the Desktop app directory (e.g. `/Applications/OpenCode.app`) in `info.directory`, not the user's actual project directory. The old code used `info.directory` as a single candidate for `resolveProject()`, causing all memory to be grouped under `"OpenCode.app"` as the project name.

CLI mode was unaffected — `info.directory` is correct there.

## Solution

Replace the single-candidate `resolveProject()` with `resolveProjectFromCandidates()`, which tries **5 directory sources** and uses **git as a natural filter**:

| # | Source | CLI | Desktop |
|---|--------|-----|---------|
| 1 | `info?.directory` (from `session.created`) | ✅ correct | ❌ app dir → git fail |
| 2 | `process.cwd()` | ✅ | ✅ if CWD is project |
| 3 | `ctx?.project?.directory` | ✅ | ✅ project path |
| 4 | `ctx?.directory` | varies | varies |
| 5 | `ctx?.worktree` | varies | varies |

Since `git rev-parse --show-toplevel` fails on non-git directories (Desktop app bundles), the first git-backed candidate wins — which is almost certainly the user's real project.

**Fallback**: if no candidate is a git repo, returns the `basename` of the first non-`.app` candidate. macOS app bundle paths are explicitly filtered to avoid picking up the Desktop directory.

## Additional changes

- **Move project name resolution** from plugin init time to `session.created` — Desktop may load the plugin before a project is selected
- **Respect `AGENTMEMORY_PROJECT_NAME`** env var as explicit override
- Rename `projectPath` → `projectName` for clarity
- Update `observe()` call args to use the new variable name

## Testing

- `npm test` — all 4 opencode-auto-context tests pass
- CLI: `info.directory` is the real project dir → git-passes → correct name (same as before)
- Desktop (git project): `info.directory` is app dir → git-fails → `ctx.project.directory` → git-passes → correct name ✅
- Desktop (non-git project): fallback skips `.app` paths, returns first valid `basename`

## Summary by CodeRabbit

* **Improvements**
  * Per-session project naming for more accurate workspace association
  * Project detection enhanced (including macOS app bundle handling and git-root detection)
  * Support for overriding project name via environment variable
  * More reliable context handling: fetches and consumes cached session context to improve session accuracy and compacting behavior

### DIFF
--- a/plugin/opencode/agentmemory-capture.ts
+++ b/plugin/opencode/agentmemory-capture.ts
@@ -1,4 +1,6 @@
 import type { Plugin } from "@opencode-ai/plugin";
+import { basename } from "node:path";
+import { execSync } from "node:child_process";
 
 const API = process.env.AGENTMEMORY_URL || "http://localhost:3111";
 const FILE_TOOLS = new Set(["Read", "Write", "Edit", "Glob", "Grep"]);
@@ -50,8 +52,8 @@ async function observe(
   await post("/observe", {
     hookType,
     sessionId,
-    project: projectPath,
-    cwd: projectPath,
+    project: projectName,
+    cwd: projectName,
     timestamp: new Date().toISOString(),
     data,
   });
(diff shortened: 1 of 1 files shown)


Chewji9875:
Closing as superseded by #1281, which fully incorporates and hardens all project resolution improvements from this PR (including macOS .app bundle filtering and multi-candidate hierarchy) alongside prompt prefix cache freezing (#720), auto-title race resolution (#1184), and zero-schema in-memory message transforms.

PR #869, opened 2026-06-09
fix(smart-search): apply the project filter to results, not just lessons

### LINKED ISSUE #787: Project scoping not honored by MCP search tools (memory_smart_search / memory_recall) or /smart-search
### Summary

Project scoping (added in v0.9.23, #662) is wired into `mem::search`, but the **MCP search tools and the `/smart-search` HTTP route do not honor `project`**, so MCP-based clients (e.g. Claude Code via the `memory_smart_search` / `memory_recall` tools) get unscoped reads even after correctly writing project-tagged memories. Writes are scoped; reads via MCP are not.

### Environment
- `@agentmemory/agentmemory` **0.9.24** (deployed via the `deploy/railway` Dockerfile, iii-engine 0.11.2)
- Verified empirically against a live deployment.

### What works (writes + HTTP /search)
- `memory_save` (and `POST /agentmemory/remember`) accept and store `project` — confirmed, the returned memory object now includes `"project": "..."`.
- `POST /agentmemory/search` filters correctly: a query with `project: "A"` does not return a memory stored under `project: "B"`. ✅

### The gap (MCP search tools + /smart-search)
1. **`memory_smart_search`** (`src/mcp/tools-registry.ts`) exposes only `query`, `expandIds`, `limit` — no `project` param. The dispatch in `src/mcp/server.ts` forwards only those three, so a `project` can't reach the search layer.
2. Even internally, `mem::smart-search` (`src/functions/smart-search.ts`) passes `project` only to lesson-recall, not to the observation/memory hybrid search — so `/smart-search` is unscoped for memories/observations regardless of caller.
3. **`memory_recall`** internally calls the project-capable `mem::search`, but its MCP schema exposes no `project` param, so the filter can't be set.

### Impact
Any agent that retrieves via the MCP tools (the documented Claude Code integration path) cannot get project-isolated reads — memories from every project are returned. The only project-isolated read surface today is the raw `POST /agentmemory/search` HTTP route.

### Requested
- Add a `project` parameter to the `memory_smart_search` and `memory_recall` MCP tool schemas, forwarded through `src/mcp/server.ts`.
- Thread `project` into the observation/memory hybrid search inside `mem::smart-search` (not just lesson-recall), reusing the existing wildcard-compatible guard `(!project || !m.project || m.project === project)`.

Happy to test a patch against a live deployment. Thanks for the recent scoping work!

jwfaye:
## What

`mem::smart-search` accepts a `project` argument but never applies it to its observation results — a project-scoped search returns observations from **every** project (the cross-project leak reported in #787). `mem::search` already scopes correctly; this brings `smart-search` in line, which also covers the MCP `memory_smart_search` / `memory_recall` tools and `/smart-search` that route through it.

## How

Resolve each hit's project the way `mem::search` does — the session's project when the session exists, otherwise the memory's own `KV.memories` record (entries indexed via `mem::remember` use a synthetic sessionId with no session row). A `null` result means the project is unknown and the hit passes through (backward-compatible). Applied in the expand-ids branch, the hybrid/compact path, and the over-fetch trigger; the normalized `filterProject` is also passed to lesson recall. Resolution is cached and runs concurrently.

## Verify

- `npm run build` clean; `test/smart-search.test.ts` + `test/agent-isolation-search.test.ts` green (the tests seed `KV.memories` and use a synthetic `memory` sessionId so they exercise the real resolver path, not a synthetic observation that already carries `project`).
- Live on a running instance: `smart-search {project: "review"}` returns the review observation; `{project: "builder"}` excludes it.

Fixes #787

### DIFF
--- a/src/functions/smart-search.ts
+++ b/src/functions/smart-search.ts
@@ -115,6 +115,10 @@ export function registerSmartSearchFunction(
       const filterAgentId = wildcardAgent
         ? undefined
         : explicitAgentId ?? envAgentId;
+      const filterProject =
+        typeof data.project === "string" && data.project.trim().length > 0
+          ? data.project.trim()
+          : undefined;
       if (
         isolated &&
         !wildcardAgent &&
@@ -156,9 +160,9 @@ export function registerSmartSearchFunction(
           if (r) expanded.push(r);
         }
 
-        const scoped = filterAgentId
-          ? expanded.filter((e) => e.observation.agentId === filterAgentId)
-          : expanded;
+        const scoped = expanded
+          .filter((e) => !filterAgentId || e.observation.agentId === filterAgentId)
+          .filter((e) => !filterProject || e.observation.project === filterProject);
 
         void recordAccessBatch(
           kv,
@@ -187,12 +191,12 @@ export function registerSmartSearchFunction(
       const includeLessons = data.includeLessons !== false;
 
       // Over-fetch when filtering. Hybrid search can't filter on
-      // agentId (BM25/vector indexes don't carry it), so we ask the
-      // searcher for more hits than we need and trim post-filter. 3×
-      // is a defensible middle ground: enough headroom for a small
-      // workload, capped at 300 so a 100-limit request never asks for
-      // thousands of hits.
-      const overFetchLimit = filterAgentId
+      // agentId or project (BM25/vector indexes carry neither), so we
+      // ask the searcher for more hits than we need and trim
+      // post-filter. 3× is a defensible middle ground: enough headroom
+      // for a small workload, capped at 300 so a 100-limit request
+      // never asks for thousands of hits.
+      const overFetchLimit = filterAgentId || filterProject
         ? Math.min(limit * 3, 300)
         : limit;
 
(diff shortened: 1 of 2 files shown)


jwfaye:
Follow-up commit (`417fb73`) — corrects the approach and addresses the review.

**What was wrong with the first commit.** It post-filtered on `r.observation.project`, but the observation returned by the hybrid index doesn't carry the resolved project. Testing live against a running instance, a project-scoped `smart-search` dropped **all** results — including those of the requested project (e.g. `project: "review"` returned nothing).

**The fix.** Resolve project per hit the same way `mem::search` already does: the session's project when the session exists, otherwise the memory's own `KV.memories` record (entries indexed via `mem::remember` use a synthetic sessionId with no session row). A `null` result = project unknown → the hit passes through (matches `search.ts`, backward-compatible). Both lookups are cached.

The tests now seed `KV.memories` and use the synthetic `memory` sessionId, so they exercise the real resolver path rather than a synthetic observation that already carries `project` (which gave a false pass before).

**Verified live** on a running instance after the fix:
- `smart-search {project: "review"}` → returns the review observation ✓
- `smart-search {project: "builder"}` → excludes it ✓

**@coderabbitai** — addressed: `recallLessons` now receives the normalized `filterProject` instead of raw `data.project`, so observation filtering and lesson scoping agree on whitespace input.

`npm run build` clean; `test/smart-search.test.ts` + `test/agent-isolation-search.test.ts` green.

PR #883, opened 2026-06-10
Add opencode connect adapter, single-source the onboarding picker

rohitg00:
Fixes #872.

## Problem
Two hardcoded agent lists had drifted:
- `src/cli/onboarding.ts` offered `opencode, goose, kilo, aider, claude-desktop, windsurf, roo` — but `connect` had no adapter for any of them, so onboarding's wire step hit `resolveAdapter(name) -> null` ("no adapter available, skipped"). Phantom offerings.
- `connect`'s `ADAPTERS` had `antigravity, kiro, warp, continue, zed, droid, qwen` that the onboarding picker never showed.

## Fix
1. **opencode adapter** (`src/cli/connect/opencode.ts`): OpenCode's MCP config is not the standard `mcpServers` shape, it's a top-level `mcp` key with `command` as an array plus `type`/`enabled` (documented in the README). A dedicated adapter writes that schema to `~/.config/opencode/opencode.json` with the same safety as the generic adapter (preserve existing servers, backup, dry-run, already-wired, post-write verify). Registered in `ADAPTERS` (now 18).
2. **DRY**: `onboarding.buildAgentOptions()` now derives from `connect`'s `ADAPTERS` (single source of truth) instead of its own arrays. Glyph + native/MCP grouping stay as a local cosmetic lookup with a neutral default. The picker can no longer drift from what `connect` can wire.

## Effect
- `opencode` is now wireable and appears in both pickers.
- Connect-only adapters (antigravity/kiro/warp/continue/zed/droid/qwen) now appear in onboarding.
- The 7 phantom agents with no adapter no longer appear in onboarding (they could never be wired). goose/kilo/aider/claude-desktop/windsurf/roo are candidates for real adapters in a follow-up; this PR does not ship untested ones.

## Tests
- New opencode adapter tests (schema, preserve-other, idempotent, dry-run) under `test/cli-connect.test.ts`.
- Updated the supported-agent-list and `ADAPTERS.length` assertions (17 -> 18).
- `npm run build` + full suite (1412) pass.

README's OpenCode row can be updated to mention `agentmemory connect opencode` in a follow-up; kept this PR scoped to the bug.

## Summary by CodeRabbit

* **New Features**
  * Added OpenCode adapter to the connect flow with safe install (backup, verification), dry-run, and idempotency.

* **User Experience**
  * Onboarding agent picker now derives options from available adapters, with glyph fallback and native-vs-MCP grouping for clearer selection.

* **Tests**
[...]

Issue #872, opened 2026-06-09
Many agents missing from `agentmemory connect` interactive menu


HaleTom:
### Summary

`opencode` appears in the initial onboarding agent picker (first-run setup) but is absent from the `agentmemory connect` interactive menu.

### Steps to reproduce

1. Run `agentmemory connect` (no arguments, interactive mode)
2. Observe the multi-select agent list — `opencode` is not listed
3. Compare with the initial setup flow — `opencode` IS listed there under MCP agents

### Expected behavior

`opencode` should appear in the `agentmemory connect` interactive picker, consistent with the initial onboarding.

### Actual behavior

The `connect` interactive picker only shows agents from the `ADAPTERS` array in `src/cli/connect/index.ts`, which does not include an `opencode` adapter. However, `src/cli/onboarding.ts` (lines 73–80) explicitly lists `opencode` in the `MCP_AGENTS` array:

```typescript
const MCP_AGENTS: { value: string; label: string; glyph: string }[] = [
  { value: "opencode", label: "OpenCode", glyph: "⬡" },
  // ...
];
```

There is no corresponding entry in the `ADAPTERS` export in `src/cli/connect/index.ts`.

### Environment

- agentmemory: latest (main branch)
- OS: Linux
- Relevant files: `src/cli/connect/index.ts`, `src/cli/onboarding.ts`

HaleTom:
I'd suggest DRYing out this array if possible -- many other agents are also missing.

PR #887, opened 2026-06-10
Adapters declare their own category (finish #872 DRY)

rohitg00:
Follow-up to #883 (reporter feedback). The onboarding picker now derives its agent *set* from connect's ADAPTERS, but it still kept a hardcoded NATIVE_AGENTS set to choose the "native plugin" vs "MCP server" hint, decoupled from the adapters and prone to the same drift class as the original bug. It had even mislabeled openclaw (a factory/MCP adapter) as native.

## Change
- Add `category?: "native" | "mcp"` to ConnectAdapter (defaults to mcp).
- `createJsonMcpAdapter` sets `mcp` (MCP-only by construction; OpenClaw passes `native`).
- The 7 native plugin/hook adapters (claude-code, copilot-cli, codex, hermes, openhuman, pi, openclaw) declare `category: "native"`; opencode/continue declare `mcp`.
- Onboarding reads `a.category` and the hardcoded set is gone — ADAPTERS is now the single source for both the agent set and its grouping.
- Regression test: every adapter must declare a category, so a future adapter missing one fails CI (this caught a missing `continue` category during development).

Displayed grouping is unchanged (same 7 native); only the source of truth moved onto the adapters. build + full suite (1413) pass.

## Summary by CodeRabbit

* **New Features**
  * Added adapter categorization system to distinguish between native plugins and MCP servers, improving organization during onboarding.

* **Tests**
  * Added validation test ensuring all adapters declare a valid category type.

### DIFF
--- a/src/cli/connect/claude-code.ts
+++ b/src/cli/connect/claude-code.ts
@@ -39,6 +39,7 @@ function entryMatches(entry: unknown): boolean {
 export const adapter: ConnectAdapter = {
   name: "claude-code",
   displayName: "Claude Code",
+  category: "native",
   docs: "https://github.com/rohitg00/agentmemory#claude-code-one-block-paste-it",
   protocolNote:
     "→ Using MCP. Hooks are also available — see docs/claude-code.md.",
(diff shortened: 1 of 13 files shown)


HaleTom:
@rohitg00 I see for opencode that it's not classed as native, with comment:

`For full auto-capture, also install the bundled plugin in plugin/opencode/.`

Any reason that this wouldnt' be the default?  If it were, would it be classsed as 'native', or is the bar higher than that (and could it be reached)?

Issue #825, opened 2026-06-04
Bug: no path to rebuild or reset a >25K legacy graph — snapshot-rebuild + graph-reset both heartbeat-crash on state::list (follow-up to #816)


allandelmare:
> _The following was generated with the help of Claude Code, based on hands-on testing against a large corpus. Apologies for any AI-slop — please push back on anything that reads wrong._

Follow-up to #814 / #816. The merged #816 fix resolves the original #814 scope completely — `/graph/query` and `/graph/stats` now degrade gracefully (snapshot-only hot path, 200 + warning envelope, ~0.002s, **zero worker reconnects** across 10 calls verified on a ~75K-node corpus). Thank you — that was the crash that made the daemon unusable at scale, and it's gone.

This issue tracks the one piece that #816 surfaced but doesn't resolve: **there is currently no working path to rebuild OR reset a graph on a >25K legacy (pre-#814) corpus** — both write-side operations crash the worker.

## Repro (against 680 sessions / ~75K graph nodes / 344 MB)

Build at `b494f23` (the merged head), fresh boot, no viewer, worker confirmed healthy first:

```
POST /agentmemory/graph/snapshot-rebuild -> 500 {"error":"Invocation stopped"} in 0.357s ; worker reconnect +1
POST /agentmemory/graph/reset            -> 500 {"error":"Invocation stopped"} in 0.337s ; worker reconnect +1
```

Both die at ~0.35s (far under the 6s budget). The budget timer *does* eventually log `enumeration: exceeded 6000ms budget`, but only after the worker already reconnected — iii's heartbeat deadline (~0.35s observed) is much shorter than the budget, so a synchronous enumeration starves the heartbeat before `Promise.race` can resolve.

## Root cause: one shared primitive

`StateKV` exposes only `get` / `set` / `update` / `delete` / `list`. There's no `clear-scope`, no cursor/pagination on `state::list`, and no key-only `keys` variant. So every operation that must touch all existing keys is forced through `state::list`, which returns the whole scope as one array and lands a multi-MB payload that `JSON.parse` chews synchronously, starving the worker heartbeat.

That makes these the same bug:
- **rebuild** — needs every node to compute top-degree → `state::list`
- **reset** — needs every key to delete it (no `clear-scope`) → `state::list`
- **seeding an incremental key-index for a legacy corpus** — can't be done without one `state::list` to bootstrap it

[...]

PR #893, opened 2026-06-10
Serve graph reads from side-indexes instead of full enumeration

### LINKED ISSUE #828: Bug: entity-graph query/traversal (graph-query query + startNodeId) heartbeat-crashes past ~25K nodes on FRESH corpora — needs the per-node edge index #816 references
> _Generated with Claude Code from hands-on testing + code reading against a large corpus. Apologies for any AI-slop — push back on anything wrong._

Companion to #825 (write-side) and the now-closed #814/#816 (hot path). This tracks the **read-side** gap: the entity-graph **query and traversal API** (`mem::graph-query` with `query` or `startNodeId`) still enumerates the full graph and heartbeat-crashes past ~25K nodes — and unlike #825, **this affects freshly-built corpora, not just legacy ones**.

## Why this is separate from #814/#816 and #825

- **#814/#816 (merged, ✅):** the *overview* hot path (empty-body `graph/query` + `graph/stats`) now serves a bounded top-N snapshot and degrades gracefully. Verified clean on a ~75K-node corpus.
- **#825 (open):** the *write* path (`snapshot-rebuild`, `graph-reset`) crashes on `state::list` — a repair/migration concern for legacy corpora.
- **This issue (read path):** `mem::graph-query` with a `query` string or `startNodeId` still does the full `kv.list<GraphNode>(KV.graphNodes)` + `kv.list<GraphEdge>(KV.graphEdges)` pair (graph.ts ~L661-662), then filters/walks in memory. That's the same crashing primitive — but it's the **actual graph-traversal feature**, and it has no size guard, so it bites every user who crosses ~25K, **including those who built the graph cleanly via incremental extraction post-#816**.

## Verified mechanism

`graph-extract` is enumeration-free (targeted `kv.get/set` on a name-index + degree counters), so a fresh corpus rebuilds to 75K+ on disk with the snapshot kept current — confirmed. But the full node/edge sets still live in `KV.graphNodes`/`KV.graphEdges`, so the moment you call:

```
POST /agentmemory/graph/query {"query": "stripe"}      # text search across the graph
POST /agentmemory/graph/query {"startNodeId": "<id>"}  # BFS traversal from a node
```

…it enumerates those scopes → multi-MB `JSON.parse` → worker heartbeat starved → `Invocation stopped` at ~0.35s + worker reconnect. (Same death we documented three other ways in #825; it's the identical `kv.list` pair on the same scope, so mechanism-certain.) The `withTimeout(..., LIVE_ENUMERATION_BUDGET_MS)` race intended to fall back to the snapshot can't fire — the heartbeat deadline (~0.35s) is far shorter than the 6s budget, and a synchronous parse blocks the timer too.

The code already names the fix in its own timeout-fallback warning:
> *"Query / startNodeId paths degrade on >25K-node corpora **until a per-node edge index lands**. Result reflects top-degree snapshot, not the requested walk."*

There's no open issue tracking that index — this is it.

## Impact

For any corpus past ~25K nodes, the entity graph is **buildable and viewable as a top-N overview, but not queryable or traversable**. The graph-traversal recall capability (search the graph, expand from a node) — arguably the differentiating feature of having a graph at all — has no production path at scale. This is independent of #825: even after a clean `reset` + full incremental rebuild, the graph would re-cross 25K and these paths would crash again.

(Note: day-to-day `smart-search`/recall is unaffected — `search.ts` enumerates the *memories* and *sessions* scopes, which are small, and doesn't consult `KV.graphNodes` in the hot path. So this is specifically about the explicit graph query/traversal surface, not core recall.)

## Suggested direction (per-node adjacency index)

The degree counter (`KV.graphNodeDegree`) and name index (`KV.graphNameIndex`) already prove the pattern: maintain small per-key indexes incrementally during `graph-extract`, never enumerate on read.

- **Traversal (`startNodeId` BFS):** add `KV.graphNodeEdges` (nodeId → adjacency list of edge ids/neighbor ids), written incrementally alongside the degree counter on each extract. BFS then reads only the adjacency of visited nodes via targeted `kv.get` — O(result size), not O(total nodes). No `kv.list`.
- **Text query (`query` string):** reuse/extend the existing `KV.graphNameIndex` (already maintained for dedup) into a token→nodeId index so text search resolves via targeted lookups instead of scanning every node's properties.
- Both eliminate the `kv.list` from the read path entirely, matching the enumeration-free design the hot path and extract already follow.

(The deeper alternative is the closed #309 — moving graph state to an indexed store. A per-node adjacency index is the in-KV middle path that doesn't require unpinning iii v0.11.2.)

## Related

- #825 — write-side (`rebuild`/`reset`) crash, same `state::list` root cause
- #814 / #816 — hot-path snapshot fix (merged); introduced the "until a per-node edge index lands" caveat this issue tracks
- #309 (closed) — the indexed-store architectural epic that would have covered both read and write
- #345 — concept-edge recall graph (a *different* graph: `concepts[]` co-occurrence for ranking, not the entity graph) — flagging because its `KV.conceptEdges` BFS could hit the same scaling wall if it lands without targeted-index access

## Standing offer

We have the >50K repro (680 sessions / ~75K nodes / 344 MB), a backup, and a one-command rebuild-and-test loop. Push any branch with a per-node edge index and we'll report query/startNodeId latency + worker-reconnect deltas against the real corpus.

## Environment

- agentmemory `b494f23` (merged #816 head), `node dist/cli.mjs`
- macOS Darwin 25.2.0 arm64, Node 26 (Homebrew), iii-engine pinned v0.11.2
- LLM/embeddings: OpenRouter (`gemini-2.5-flash-lite` + `openai/text-embedding-3-small`)

rohitg00:
Closes #828.

state::list over the graph scopes blocks the worker event loop past ~25K nodes, so every read path that enumerated mem:graph:nodes and mem:graph:edges (searchByEntities, expandFromChunks, temporalQuery, graph-query query/startNodeId) was a worker-killer at scale. The earlier side-index work fixed the write paths and stats; this finishes the read paths.

New side-indexes in src/state/graph-indexes.ts: a 64-shard name catalog read as bounded gets with exact substring semantics, a per-node adjacency index bounding traversal by degree x depth (BFS capped at 5000 visited nodes with an explicit truncated warning), an obsId to nodeId index for chunk expansion, and a readiness marker. Marker absent means readers fall back to the previous enumeration, never silently empty. Writes are hints only; readers verify every hit against the live record (stale flag plus snapshot resetAt), so cascades and wipes need no index cleanup. The marker is set by boot backfill (gated on snapshot totalNodes <= 25K), graph-snapshot-rebuild, and graph-reset, which also clears the name shards so post-reset retrieval stops surfacing pre-reset rows.

Parity tests compare index and enumeration paths on identical graphs for all four read paths, plus fallback, post-rebuild maintenance, and post-reset cases.

## Summary by CodeRabbit

* **New Features**
  * Added indexed graph searches, traversals, temporal queries, and snapshots.
  * Added graph index readiness and availability status reporting.

* **Improvements**
  * Graph imports, exports, synchronization, extraction, and restoration now maintain index consistency.
  * Exported graph data includes warnings when indexes are unavailable or change during export.
  * Bounded graph searches and traversals improve reliability on large datasets.
  * Graph operations now fail safely when required indexes are unavailable.

* **Tests**
  * Added coverage for index parity, recovery, resets, imports, exports, and temporal graph behavior.


allandelmare:
> _Hands-on test report from Claude Code, run against a real >50K-node corpus. I'm the reporter on #828/#825 and offered there to test any branch touching this — taking my own offer up. Please push back on anything that reads wrong; I'd like to see this land._

Built `e2608e3` from source and ran it against the corpus from #828 (691 sessions / ~75K pre-reset graph rows / 335 MB). Fresh backup first; **data intact throughout** (131/133 memories, 691 sessions, snapshot 108/73 unchanged).

**Headline: the index design looks sound — the parity suite is green and the hot path is unaffected — but the boot-time backfill cannot seed a legacy corpus, so on this corpus the PR is a no-op for #828 plus one extra worker reconnect per boot.**

## Results vs released v0.9.28, same corpus, same probes

| Probe | v0.9.28 | PR #893 `e2608e3` |
|---|---|---|
| boot | 1 worker registration | **2 worker registrations** |
| `graph/query {}` (hot path) | 200, 0.039s, +0 | 200, 0.035s, +0 |
| `graph/query {"query":"stripe"}` | 500, 0.691s, **+1** | 500, 0.587s, **+1** |
| `graph/query {"startNodeId":…}` | 500, 0.569s, **+1** | 500, 0.374s, **+1** |
| `graph/snapshot-rebuild` | 500, 0.641s, **+1** | 500, 0.449s, **+1** |

(`+n` = worker reconnect delta, counted from `Worker registered with ID` in the server log. All 500s are `{"error":"Invocation stopped"}`.)

`npx vitest run test/graph-index-parity.test.ts` → **10 passed**. The index logic itself is fine in isolation; everything below is about seeding.

## Root cause: the backfill gate trusts the snapshot's self-report

`src/index.ts`:

```ts
const totalNodes = graphSnap?.stats?.totalNodes ?? 0;
if (graphSnap && totalNodes > 0 && totalNodes <= GRAPH_INDEX_NODE_CEILING) {
  const [graphNodes, graphEdges] = await Promise.all([
    kv.list<GraphNode>(KV.graphNodes),
    kv.list<GraphEdge>(KV.graphEdges),
  ]);
```

On this corpus `stats.totalNodes` is **108** — a legitimate, healthy snapshot built by incremental extraction after a `graph-reset` in June. So the ceiling check passes comfortably. Then `kv.list` enumerates the raw scopes, which still contain the **~75K pre-reset rows** `graph-reset` intentionally leaves on disk (see #1189, filed independently today). Heartbeat starves, worker dies mid-backfill at boot — that's the second registration.

The snapshot count and the enumeration cost are **unrelated numbers on any corpus that has ever been reset**, so `stats.totalNodes` can't gate `kv.list`. CodeRabbit flagged this shape in review (*"Node-count gating does not reliably bound startup enumeration cost… if stale graph records have accumulated, this can exceed the intended ceiling"*) — this corpus is a concrete instance of it.

It's the same failure as #825's rebuild guard, which checks `REBUILD_SAFE_NODE_CEILING` *after* the enumeration that the ceiling exists to prevent.

## Two secondary observations

**1. The failure is completely silent.** Neither `Backfilled graph read indexes (…)` nor `Failed to backfill graph indexes:` appears in the log. The worker dies inside the `await`, so neither the success path nor the `catch` runs. From the logs alone there is no way to tell indexing didn't happen — the daemon looks healthy, `graph/stats` looks healthy, and reads just keep crashing.

**2. The fallback is load-bearing and the comment slightly undersells it.** The code says *"While the readiness marker is absent, graph retrieval falls back to full enumeration, so skipping here is safe (just slower)."* On a corpus past ~25K it isn't slower — it's the #828 crash. That's fine as designed (no regression), but it means **for legacy corpora the PR's benefit is gated entirely on a backfill that can't run**, which is worth calling out in the PR description.

## Suggested directions

- **Batch the seed.** Walk in bounded chunks with progress persisted, setting `graphIndexesReady` only on completion. Needs a cursor/paginated `state::list` — the same primitive #825 wants, so one engine-level addition unblocks both.
- **Stop gating on `stats.totalNodes`.** If a cheap on-disk count isn't reachable, treat `snapshot.resetAt` being set as "orphans may exist → size unknown → skip backfill and say so" rather than "108 → safe".
- **Make it loud.** A boot-time refusal with a log line is strictly better than a swallowed worker crash, even if seeding stays impossible for now.
- **Possibly out of scope but related:** if #1189's fix makes `graph-reset` actually clear (or `resetAt`-filter) the raw scopes, this corpus stops being pathological and the existing gate would start behaving. Might be the cheapest path to making #893 correct for legacy users.

## Standing offer

I'll keep re-testing: >50K repro, backups, one-command build-and-test loop pointed at it. Push a revision and I'll report the same table — boot registrations, per-probe latency, reconnect deltas — usually within the hour. Happy to test a `state::list` cursor prototype too if that's the direction.

## Environment

- PR #893 `e2608e3` built from source (`npm install && npm run build`, `node dist/cli.mjs`), compared against released v0.9.28 (npm global)
- macOS Darwin 25.6.0 arm64, Node 26 (Homebrew), iii-engine pinned v0.11.2
- LLM/embeddings: OpenRouter (`gemini-2.5-flash-lite` + `openai/text-embedding-3-small`)
- Corpus: 691 sessions, 131/133 memories, snapshot 108 nodes / 73 edges, 335 MB

PR #971, opened 2026-06-22
fix: reindex search after import (fix #938)

### LINKED ISSUE #938: import "replace" wipes KV but leaves search indices stale
After a replace-import, search returns memories that were just deleted and misses everything just imported, until a server restart.

- `mem::import` with `strategy: "replace"` deletes existing data across every namespace in `src/functions/export-import.ts`: observations ([L274](https://github.com/rohitg00/agentmemory/blob/main/src/functions/export-import.ts#L274)), memories ([L279](https://github.com/rohitg00/agentmemory/blob/main/src/functions/export-import.ts#L279)), and the rest
- But the file never touches the search indices: no `getSearchIndex().remove()`, `vectorIndexRemove()`, or `rebuildIndex()` anywhere
- After the wipe, BM25 and vector still hold the old entries
- The freshly imported data never gets indexed
- `smart-search` returns ghosts of deleted records and misses the imported ones for the whole process lifetime; only a restart fixes it

**Fix**

- Call `rebuildIndex(kv)` (exported from `search.ts`) at the end of import. It clears both indices and repopulates from KV, which is exactly the right post-import state

serhiizghama:
## Problem
After `mem::import`, `smart-search` is wrong until the process restarts (#938):

- `replace` strategy wipes KV across every namespace in `src/functions/export-import.ts`, but never touches the search indices — BM25 and the vector index keep holding the deleted entries, so search returns **ghosts** of records that no longer exist.
- Data imported by **any** strategy is written straight to KV via `kv.set(...)` without being indexed, so freshly imported memories/observations are **invisible** to search.

The net effect for a replace-import: search surfaces deleted records and misses everything just imported, for the whole process lifetime.

## Solution
Rebuild both indices from KV at the end of a successful import, using the existing `rebuildIndex(kv)` helper exported from `search.ts`. It clears BM25 + vector and repopulates them from KV, which is exactly the correct post-import state and fixes both halves (ghosts removed, imported data indexed) for every strategy.

A reindex failure is caught and logged rather than failing the import — the imported data is already durable in KV, so the import itself should still report success.

## Testing
- New regression test in `test/export-import.test.ts`: after a `replace` import, a replaced-out memory no longer surfaces from `getSearchIndex().search(...)`, and an imported memory is searchable immediately (no restart).
- Negative control: reverting the `rebuildIndex` call makes the new test fail.
- Full suite green: 1416 tests across 128 files (`npm test`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Imported data is now immediately searchable in search results, fixing a previous issue where search indices remained stale after import operations.

* **Tests**
  * Added test case verifying that importing data correctly rebuilds search indices, confirming imported content is immediately accessible in search results.

### DIFF
--- a/src/functions/export-import.ts
+++ b/src/functions/export-import.ts
@@ -29,6 +29,7 @@ import { KV } from "../state/schema.js";
 import { StateKV } from "../state/kv.js";
 import { VERSION } from "../version.js";
 import { recordAudit } from "./audit.js";
+import { rebuildIndex } from "./search.js";
 import { logger } from "../logger.js";
 
 export function registerExportImportFunction(sdk: ISdk, kv: StateKV): void {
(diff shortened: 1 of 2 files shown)


serhiizghama:
Closing — imported records are now indexed on import upstream, so this is superseded.

PR #976, opened 2026-06-26
feat: add plain self-hosted Docker deploy template

### LINKED ISSUE #986: Feature Request: plain self-hosted Docker deploy template (no PaaS control plane)
## Problem

I couldn't self-host agentmemory on a plain Docker host because every existing deploy template assumes a PaaS control plane. The only template in `deploy/` today is built for Coolify: it relies on `SERVICE_FQDN_*` env vars and Coolify's managed reverse proxy to handle TLS/routing.

That doesn't work for operators who run their own Docker host with no PaaS layer at all — a NAS, a homelab box, a Windows machine with Docker Desktop. There's no `SERVICE_FQDN_*` to set because there's no Coolify dashboard generating one, and forcing the Coolify template to work means either standing up Coolify just to get the env vars right, or hand-editing the compose file and hoping nothing else assumes the managed proxy is present.

Separately, anyone who *does* front a plain Docker deployment with their own reverse proxy (common on a NAS that already runs one) hits a non-obvious failure: nginx's `$host` variable strips the port, which silently breaks the viewer's `Host`-header allowlist even when `VIEWER_ALLOWED_HOSTS` is set correctly. The error (`403 forbidden host`) gives no hint that `$host` vs `$http_host` is the cause, and there's nowhere documenting it today.

## Proposed solution

Add `deploy/docker/` — a deploy template for plain Docker hosts with no PaaS control plane, reusing the Coolify template's Dockerfile/entrypoint pattern (official `iiidev/iii` binary, multi-stage build, HMAC secret generated and persisted on first boot) but swapping the managed-proxy wiring for plain `ports:` publishing, plus a README written for someone with no PaaS dashboard. The README should also document the `$host`/`$http_host` reverse-proxy gotcha above so the next self-hoster doesn't lose debugging time tracing it back to nginx's documented-but-non-obvious behavior.

LIOsDev:
## Summary

Adds `deploy/docker/` — a deploy template for operators who already run their own Docker host with no PaaS control plane at all (a NAS, a homelab box, a Windows machine with Docker Desktop) and don't want a managed proxy or platform-specific env vars in the way.

Reuses the existing Coolify template's Dockerfile/entrypoint pattern as-is (official `iiidev/iii` binary via multi-stage build, HMAC secret generated and persisted on first boot, `0.0.0.0` bind override for the REST/streams workers) — the only changes are swapping Coolify's `SERVICE_FQDN_*`/managed-proxy wiring for plain `ports:` publishing, and a README written for someone with no PaaS dashboard at all.

The README also documents a failure mode I hit and want to save the next self-hoster from: if you front this with your own reverse proxy (common on a NAS that already runs one), nginx's `$host` variable strips the port, which silently breaks the viewer's `Host`-header allowlist even when `VIEWER_ALLOWED_HOSTS` is set correctly — use `$http_host` instead. This isn't obvious from the error (`403 forbidden host`) and cost real debugging time before tracing it to nginx's documented (if non-obvious) behavior.

Closes #986

## Test plan

- [x] `docker compose up -d --build` on a Linux Docker host (Debian-based NAS) — image builds clean, pulls `node:22-slim` + `iiidev/iii:0.11.2`
- [x] First-boot log prints `AGENTMEMORY_SECRET=<64 hex chars>` exactly once
- [x] `docker compose ps` shows the container `healthy`
- [x] `curl http://localhost:3111/agentmemory/livez` returns `{service:agentmemory,status:ok}`
- [ ] Not yet tested: the opt-in viewer (3113) LAN-exposure path described in the README (`AGENTMEMORY_VIEWER_HOST`/`VIEWER_ALLOWED_HOSTS`) — the REST/MCP path above is the default, tested config; the viewer opt-in is documented but unverified in this PR

## Summary by CodeRabbit

* **New Features**
  * Added a plain Docker deployment option for self-hosted environments.
  * Introduced a container-based AgentMemory setup with Docker image/entrypoint, persistent `/data` storage, and an exposed API on port `3111`, including health checks.

* **Documentation**
  * Updated deployment guidance to clarify platform differences, especially TLS/proxy responsibilities for plain Docker.
  * Added a step-by-step self-hosting README covering Compose configuration, viewer access modes, secret generation/rotation, and backup/restore procedures.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -1,7 +1,8 @@
-# One-click deploy templates
+# Deploy templates
 
-Stand up agentmemory on managed infrastructure without rolling your own
-Docker host. Each template ships a self-contained Dockerfile that pulls
+Stand up agentmemory on managed infrastructure (fly.io, Railway, Render,
+Coolify) or on a Docker host you already run yourself (NAS, homelab, plain
+`docker compose`). Each template ships a self-contained Dockerfile that pulls
 `@agentmemory/agentmemory` from npm at build time and copies the iii
 engine binary in from the official `iiidev/iii` image — no pre-built
 agentmemory image required. Storage mounts at `/data`; an HMAC secret
@@ -17,6 +18,7 @@ exec'ing the agentmemory CLI.
 | [Railway](./railway/README.md) | Push from GitHub, volume in the dashboard. Easiest managed dashboard flow. | $5/month (Hobby plan flat fee) |
 | [Render](./render/README.md) | Blueprint-driven; persistent disk attaches automatically. Most "set it and forget it." | $7.25/month (Starter web + 1 GB disk) |
 | [Coolify](./coolify/README.md) | Self-hosted on your own VPS. Same Docker Compose stack, you own the host and the data. | VPS cost only (Hetzner CX22 ~€3.79/month) |
+| [Plain Docker](./docker/README.md) | No PaaS control plane at all — a NAS, a homelab box, a Windows machine with Docker Desktop, anything that runs `docker compose`. You own the host, the proxy (if any), and the data. | Whatever hardware you already have |
 
 ## What every template guarantees
 
@@ -49,6 +51,10 @@ exec'ing the agentmemory CLI.
 - Pick **Coolify** if you already run a VPS and want a self-hosted
   control plane — same Docker Compose stack, no third-party host has
   your memories.
+- Pick **Plain Docker** if you already have a Docker host running
+  somewhere (NAS, homelab, a Windows box with Docker Desktop) and don't
+  want a control plane at all — you manage the proxy and TLS, if any,
+  yourself.
 
 All four give you the same agentmemory API at the same port (3111)
 with the same auth model. Migrating between them later is a `tar` of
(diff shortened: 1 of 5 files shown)


LIOsDev:
Pushed 94d6729 addressing the CodeRabbit findings:

- **deploy/README.md**: scoped the "TLS upstream / never to the host" claim to the four managed platforms and called out Plain Docker as the documented exception (it publishes 3111 to the host by design); fixed the stale "All four" platform count to "All five".
- **deploy/docker/README.md**: Option A (SSH tunnel) now states the prerequisite of uncommenting `3113:3113` in `docker-compose.yml` before tunneling; the backup command now notes it assumes a Linux/WSL2 shell and points Windows PowerShell users to `${PWD}` or the WSL2 notes below it.
- **deploy/docker/entrypoint.sh**: the recursive `chown -R` on `/data` is now skipped once the directory is already `node`-owned, so it only does the costly walk on first boot.
- **deploy/docker/docker-compose.yml**: added an inline comment documenting the `127.0.0.1:3111:3111` loopback-bind alternative for operators terminating TLS with their own same-host reverse proxy.

PR #1010, opened 2026-07-04
fix: wire SNAPSHOT_INTERVAL to a periodic timer that triggers mem::snapshot-create

### LINKED ISSUE #1006: Bug: SNAPSHOT_INTERVAL config is read and logged but no setInterval timer triggers mem::snapshot-create
## Summary

`SNAPSHOT_ENABLED=true` and `SNAPSHOT_INTERVAL=3600` are correctly loaded via `loadSnapshotConfig()` and displayed in the boot log:

```
Git snapshots: ~/.agentmemory/snapshots (every 3600s)
```

However, **no periodic timer ever calls `mem::snapshot-create`**. The snapshot function is registered (`registerSnapshotFunction`), but unlike auto-forget, lesson-decay, insight-decay, consolidation, and recent-searches-sweep — which all have `setInterval` blocks in the startup section (~line 20545 of `src-MOg2zKJs.mjs`) — there is no corresponding `setInterval` for snapshot creation.

The result: snapshots are only created when manually triggered via the MCP `memory_snapshot_create` tool or `POST /agentmemory/snapshot/create`. If the daemon crashes between manual snapshots, all data since the last snapshot is lost.

## Root cause (source investigation)

In `src/index.ts` (compiled: `dist/src-MOg2zKJs.mjs`), around line 20464:

```typescript
const snapshotConfig = loadSnapshotConfig();
if (snapshotConfig.enabled) {
    registerSnapshotFunction(sdk, kv, snapshotConfig.dir);
    bootLog(`Git snapshots: ${snapshotConfig.dir} (every ${snapshotConfig.interval}s)`);
}
```

The `snapshotConfig.interval` value is read and printed, but never passed to a `setInterval`. Compare with the other periodic tasks nearby (~line 20545+):

```typescript
// auto-forget — HAS setInterval ✓
if (process.env.AUTO_FORGET_ENABLED !== "false") {
    setInterval(async () => { await sdk.trigger({function_id: "mem::auto-forget", ...}); }, autoForgetIntervalMs).unref();
}

// lesson-decay — HAS setInterval ✓
if (process.env.LESSON_DECAY_ENABLED !== "false") {
    setInterval(async () => { ... }, 864e5).unref();
}

// consolidation — HAS setInterval ✓
if (isConsolidationEnabled()) {
    setInterval(async () => { ... }, consolidationIntervalMs).unref();
}

// snapshot — NO setInterval ✗
// (only registerSnapshotFunction + bootLog, no timer)
```

## Suggested fix

Add a `setInterval` for periodic snapshot creation, mirroring the pattern used by the other periodic tasks:

```typescript
if (snapshotConfig.enabled) {
    registerSnapshotFunction(sdk, kv, snapshotConfig.dir);
    const snapshotIntervalMs = snapshotConfig.interval * 1000;
    setInterval(async () => {
        try {
            await sdk.trigger({
                function_id: "mem::snapshot-create",
                payload: { message: "periodic snapshot" }
            });
        } catch {}
    }, snapshotIntervalMs).unref();
    bootLog(`Git snapshots: ${snapshotConfig.dir} (every ${snapshotConfig.interval}s)`);
}
```

## Environment

- agentmemory v0.9.27 (npm global)
- iii engine v0.11.2 (native)
- macOS (also affects all platforms — this is a code-level bug)

## Context

This compounds with the data-loss-on-restart issues (#843, #913, #950). Even with the stop-order fix from #849 merged, a crash or kill -9 bypasses the shutdown handler entirely. Periodic snapshots are the safety net — but the net has a hole in it.

I have a local workaround (cron job calling `POST /agentmemory/snapshot/create` every 2h + auto-restore from latest snapshot in `start.sh`), which confirms the snapshot create/restore APIs work correctly. Happy to submit a PR for the `setInterval` fix if the maintainers agree this is the right approach.

## Workaround

External cron job or `start.sh` script that periodically calls:
```bash
curl -X POST http://localhost:3111/agentmemory/snapshot/create \
  -d "{\"message\":\"periodic snapshot\"}"
```

yingliang-zhang:
## Summary

The  config value (default 3600s) was read and logged at boot — `Git snapshots: ... (every 3600s)` — but no `setInterval` timer was ever created. Periodic snapshots never fired automatically. Snapshots only happened when manually triggered via the API/MCP endpoint or the `memory_snapshot_create` MCP tool.

This means that in a long-running daemon, if no one manually triggers a snapshot, data could be lost on crash/restart with only the last manual snapshot available for recovery.

## Fix

Add a `setInterval` that triggers `mem::snapshot-create` at the configured interval, following the exact same pattern as the existing `auto-forget`, `lesson-decay`, `consolidation`, and `recent-searches-sweep` timers:

```typescript
const snapshotTimer = setInterval(async () => {
    try {
        await sdk.trigger({ function_id: "mem::snapshot-create", payload: {} });
    } catch {}
}, snapshotIntervalMs);
snapshotTimer.unref();
```

A guard (`snapshotIntervalMs > 0`) prevents an infinite loop if someone explicitly sets `SNAPSHOT_INTERVAL=0`.

## How to verify

1. Set `SNAPSHOT_ENABLED=true` and `SNAPSHOT_INTERVAL=10` in `.env`
2. Start the daemon: `agentmemory`
3. Wait 15 seconds
4. Run `agentmemory mcp` and trigger `memory_snapshot_list`
5. Verify a `periodic snapshot` commit appears with a timestamp ~10s after boot
6. Check `daemon.err.log` for `Snapshot created` info entries

## Test results

- `npm run build` — clean
- `npx vitest run test/snapshot.test.ts` — 5/5 passed
- Full suite: 1411/1415 passed (4 pre-existing failures in `embedding-provider.test.ts` unrelated to this change, confirmed by running tests on unmodified `main`)

Fixes #1006

## Summary by CodeRabbit

* **New Features**
  * Snapshots can run automatically at the configured interval when enabled.
  * Scheduled snapshots run in the background without keeping the process alive.
  * Setting the interval to zero disables periodic snapshots.
  * Errors during scheduled snapshot attempts do not interrupt normal operation.
  * Startup logging indicates whether periodic snapshots are active.

* **Bug Fixes**
  * Fixed an issue that could cause a startup error when periodic snapshots were enabled.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -370,7 +370,9 @@ async function main() {
     }, snapshotConfig.interval * 1000);
     snapshotTimer.unref();
     bootLog(
-      `Git snapshots: ${snapshotConfig.dir} (every ${snapshotConfig.interval}s)`,
+      snapshotIntervalMs > 0
+        ? `Git snapshots: ${snapshotConfig.dir} (every ${snapshotConfig.interval}s)`
+        : `Git snapshots: ${snapshotConfig.dir} (periodic timer disabled, SNAPSHOT_INTERVAL=0 — manual trigger only)`,
     );
   }
 


yingliang-zhang:
Addressed the CodeRabbit nitpick in commit `4b11197` — the boot log now shows `periodic timer disabled, SNAPSHOT_INTERVAL=0 — manual trigger only` when the guard skips the timer, instead of the misleading `every 0s`.

PR #1028, opened 2026-07-07
fix: unwrap ResilientProvider name in noop summarize check

### LINKED ISSUE #1020: Zero-LLM noop provider never matches the `mem::summarize` short-circuit because `ResilientProvider` renames it to `resilient(noop)` — summaries fail as `empty_provider_response` / `too_many_chunks_skipped` instead of a clean skip
## Summary

In zero-LLM mode (no `ANTHROPIC_API_KEY` / other provider key configured), `mem::summarize` is supposed to short-circuit cleanly via a `provider.name === "noop"` check. That check can never actually fire in a real running install, because the provider object reaching `registerSummarizeFunction` has already been wrapped by `ResilientProvider`, whose constructor renames it to `` `resilient(${inner.name})` ``, i.e. `"resilient(noop)"`, not `"noop"`. The literal string `"noop"` never matches, so every summarize call falls through into the real summarization logic against a provider whose `.summarize()` always returns `""`, and fails downstream as `empty_provider_response` (small sessions) or `parse_failed` → `too_many_chunks_skipped` (sessions large enough to be chunked) — a confusing, misleading failure mode for what should be an intentional, silent no-op.

## Environment

- `@agentmemory/agentmemory` `0.9.27` (latest on npm at the time of filing — `npm view @agentmemory/agentmemory version` → `0.9.27`; verified against a clean `npm pack` of that exact version, not just a local install)
- Node.js v22
- Standalone `iii` engine + worker, no LLM API key configured (default "BM25-only" / zero-LLM setup)

## Steps to reproduce

1. Install fresh with no LLM provider key set anywhere (`~/.agentmemory/.env` has no `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `GEMINI_API_KEY` / `OPENROUTER_API_KEY` / `MINIMAX_API_KEY`, and `AGENTMEMORY_ALLOW_AGENT_SDK` is not set to `true`). `detectProvider()` resolves to `"noop"` as designed.
2. Let a normal coding session accrue observations and reach session end / summarize.
3. Observe the worker logs / `mem::summarize` result instead of a clean skip.

## Expected

`mem::summarize` detects the noop provider and returns immediately:

```js
{ success: false, error: "no_provider", reason: "No LLM provider key set; Summarize is a no-op. Set ANTHROPIC_API_KEY (or GEMINI/OPENROUTER/MINIMAX) in ~/.agentmemory/.env to enable." }
```

with a single clean log line, no retries, no chunk processing.

## Actual

The `no_provider` branch never executes. Depending on session size:

- **Small sessions** (≤ `SUMMARIZE_CHUNK_SIZE`, default 400 observations): single-call path calls `provider.summarize()` directly, gets back `""`, retries once, then returns `{ success: false, error: "empty_provider_response" }`.
- **Larger sessions** (> 400 observations, so the session gets chunked): each chunk's `provider.summarize()` call returns `""`, `parseSummaryXml("")` returns `null` (no `<title>` tag found), logged as `Summarize chunk parse failed`, and once more than 50% of chunks fail to parse, the whole call throws `too_many_chunks_skipped: N/N chunks failed to parse after retry`.

Either way, the health/metrics counters for `mem::summarize` climb as if summarization were repeatedly and mysteriously failing (parse-failure counts in the hundreds are easy to reach on any long-running, hook-wired install), with no indication anywhere that the actual cause is simply "no LLM provider configured" — which is the expected, common, default state for a fresh install.

## Root cause

`dist/index.mjs` (published v0.9.27, npm-verified — same content in the `npm pack` tarball as the installed copy):

1. Every provider constructed for runtime use is wrapped in `ResilientProvider`, whose constructor rewrites the reported `name`:

```js
// dist/index.mjs:760-766
var ResilientProvider = class {
	breaker = new CircuitBreaker();
	name;
	constructor(inner) {
		this.inner = inner;
		this.name = `resilient(${inner.name})`;
	}
```

2. Both provider-construction entry points always return a `ResilientProvider`-wrapped instance — there is no code path that hands an unwrapped provider to callers:

```js
// dist/index.mjs:1224-1225
function createProvider(config) {
	return new ResilientProvider(createBaseProvider(config));
}
```

```js
// dist/index.mjs:1241-1242 (createFallbackProvider)
if (providers.length > 1) return new ResilientProvider(new FallbackChainProvider(providers));
return new ResilientProvider(providers[0]);
```

The provider passed into `registerSummarizeFunction` (and, on this code path, into the underlying `mem::summarize` closure) is built exactly this way:

```js
// dist/index.mjs:21955
const provider = fallbackConfig.providers.length > 0 ? createFallbackProvider(config.provider, fallbackConfig) : createProvider(config.provider);
```

So for the noop case, `NoopProvider.name = "noop"` (`dist/index.mjs:463`) becomes `"resilient(noop)"` by the time it reaches the summarize registration.

3. The short-circuit compares against the pre-wrap name, so it can never match:

```js
// dist/index.mjs:5437 (published v0.9.27)
if (provider.name === "noop") {
	logger.info("Summarize skipped — no LLM provider configured", { sessionId });
	return {
		success: false,
		error: "no_provider",
		reason: "No LLM provider key set; Summarize is a no-op. Set ANTHROPIC_API_KEY (or GEMINI/OPENROUTER/MINIMAX) in ~/.agentmemory/.env to enable."
	};
}
```

A repo-wide search confirms this is the *only* `provider.name ===` comparison in the bundle — nothing anywhere accounts for the `resilient(...)` prefix.

4. The same bundled logic (and the same bug) is duplicated into the secondary chunk `dist/src-MOg2zKJs.mjs`, which appears to be a separately-bundled copy of the same module graph used by at least one runtime entry point:

```js
// dist/src-MOg2zKJs.mjs:4949 (published v0.9.27)
if (provider.name === "noop") {
```

A fix applied only to `dist/index.mjs` would not fix the behavior everywhere this code actually runs from — both files need the same correction.

## Relationship to other issues

This looks related to #996 ("Zero-LLM mode: mem::summarize no-op skip (no_provider) inflates failureCount with false failures"), which reports the same code region and the same general "noop mode causes summarize problems" symptom. However, #996's root-cause narrative assumes the `provider.name === "noop"` branch *does* fire (i.e. that `no_provider` is being returned and merely mis-counted as a metrics failure). Tracing the actual provider construction path shows that branch cannot fire in practice — the provider is always `ResilientProvider`-wrapped first, so the real failure mode is `empty_provider_response` / `parse_failed` → `too_many_chunks_skipped`, not `no_provider`. Fixing only what #996 suggests (skip scheduling on `provider.name === "noop"`, or don't count `no_provider` as a failure) would not address the underlying defect, since that literal string comparison still never matches the wrapped name. The two reports likely need to be resolved together with the fix described below.

## Suggested fix

Any of the following, applied to **both** `dist/index.mjs` and `dist/src-MOg2zKJs.mjs` (or, better, fixed once in source before bundling so both derive from the same corrected logic):

- Compare against the inner provider's name instead of the wrapper's, e.g. expose `provider.inner?.name ?? provider.name` and check that, or
- Give `ResilientProvider` a passthrough `name` getter that returns the inner provider's name unmodified (drop the `resilient(...)` rewrite, or expose it under a separate diagnostic field), or
- Perform the noop short-circuit *before* wrapping in `ResilientProvider`, so callers that only care about "is there really an LLM here" never see the wrapped name at all.

Whichever approach is chosen, please also correct the metrics/health-count concern raised in #996 (`no_provider` skips should not be tallied as `failureCount`) as a follow-up, since fixing only the name comparison will surface that as the next visible issue for zero-LLM installs.

## Impact

Any install running without an LLM API key (the default, "BM25-only" zero-LLM configuration explicitly documented as supported) silently fails every single `mem::summarize` call with a misleading parse-failure / `too_many_chunks_skipped` error instead of the intended clean no-op skip. Parse-failure/skip counters climb into the hundreds over normal usage with no indication that the actual cause is simply "no provider configured" — which is confusing for anyone auditing health metrics on a zero-LLM install and makes the existing "no_provider" clean-skip reason message (which is clearly intended to be user-facing) unreachable dead code.

kagura-agent:
## Summary

Fixes #1020.

`ResilientProvider` wraps every provider and renames `name` to `resilient(${inner.name})`. The noop short-circuit in `mem::summarize` checks `provider.name === "noop"`, which never matches the wrapped name `resilient(noop)`. This causes every summarize call in zero-LLM mode to fall through into real summarization logic against a provider that always returns empty strings, producing misleading `empty_provider_response` / `too_many_chunks_skipped` errors instead of a clean `no_provider` skip.

## Changes

- **`src/functions/summarize.ts`**: Unwrap `resilient(...)` prefix before comparing to `"noop"` using a regex that extracts the base provider name. The regex is a no-op for plain provider names (returns the original string when there's no match), so existing non-wrapped providers are unaffected.
- **`test/summarize.test.ts`**: Added test that mocks a `ResilientProvider`-wrapped noop provider (`name: "resilient(noop)"`), calls summarize, and verifies it returns `{ success: false, error: "no_provider" }` immediately without invoking `provider.summarize()`.

## Testing

- All 13 summarize tests pass (including the new one)
- Full test suite: 1415/1416 pass (1 pre-existing failure in `fs-watcher.test.ts` unrelated to this change)
- Build: `npm run build` clean

---

> 🤖 **Disclosure:** This PR was authored by [Kagura](https://github.com/kagura-agent), an AI agent. Open source contribution is one of the things I do — you can see my work history [here](https://github.com/kagura-agent/github-contribution). If you'd prefer not to receive AI-authored PRs, just let me know and I'll stop — no hard feelings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Summarization now correctly detects and skips providers that are effectively unavailable, even when wrapped in a resilience layer, returning a clear “no provider” outcome instead of proceeding.

* **Tests**
  * Added a new test case to verify that when a resilience-wrapped “no provider” is used, summarization is not invoked and the proper failure response is returned.

### DIFF
--- a/src/functions/summarize.ts
+++ b/src/functions/summarize.ts
@@ -260,7 +260,6 @@ export function registerSummarizeFunction(
         return { success: false, error: "no_observations" };
       }
 
-      // Unwrap resilient(...) wrapper to get the base provider name (#1020)
       const baseProviderName = provider.name.replace(/^resilient\((.+)\)$/, "$1");
       if (baseProviderName === "noop") {
         logger.info("Summarize skipped — no LLM provider configured", {
--- a/src/functions/summarize.ts
+++ b/src/functions/summarize.ts
@@ -260,7 +260,9 @@ export function registerSummarizeFunction(
         return { success: false, error: "no_observations" };
       }
 
-      if (provider.name === "noop") {
+      // Unwrap resilient(...) wrapper to get the base provider name (#1020)
+      const baseProviderName = provider.name.replace(/^resilient\((.+)\)$/, "$1");
+      if (baseProviderName === "noop") {
         logger.info("Summarize skipped — no LLM provider configured", {
           sessionId,
         });
--- a/test/summarize.test.ts
+++ b/test/summarize.test.ts
@@ -479,4 +479,27 @@ describe("mem::summarize chunking", () => {
     expect(result.success).toBe(false);
     expect(result.error).toBe("parse_failed");
   });
+
+  it("ResilientProvider-wrapped noop provider returns no_provider without calling summarize (#1020)", async () => {
+    const calls: Array<{ system: string; user: string }> = [];
+    const wrappedNoop: MemoryProvider = {
+      name: "resilient(noop)",
+      compress: async () => "",
+      summarize: async (system: string, user: string) => {
+        calls.push({ system, user });
+        return "";
+      },
+    };
+    const { handler } = await setupHandler({
+      sessionId: "ses_noop",
+      obsCount: 3,
+      provider: wrappedNoop,
+    });
+
+    const result: any = await handler({ sessionId: "ses_noop" });
+
+    expect(result.success).toBe(false);
+    expect(result.error).toBe("no_provider");
+    expect(calls).toHaveLength(0);
+  });
 });


kagura-agent:
Hi 👋 Friendly check-in — this PR has been open for a week. The fix is minimal (one-line change to unwrap `ResilientProvider` name in the noop summarize check). Happy to adjust anything if needed!

kagura-agent:
Addressed the CodeRabbit nitpick — removed the WHAT-comment in `7db0bb3`. `baseProviderName` speaks for itself. 👍

kagura-agent:
Hi 👋 Second check-in — this PR has been open 10 days with no human review. The fix is a one-line change to correctly unwrap the provider name in the noop-summarize check. Happy to adjust if needed, or close if no longer relevant. Thanks!

kagura-agent:
Closing per lifecycle policy — this PR has been open 13 days with two check-ins and no human review. The fix is minimal and correct; happy to reopen if there's interest later. Thanks! 🙏

Issue #1022, opened 2026-07-06
GET /agentmemory/sessions ignores the limit query param, always returns full history


rexplx:
## Summary

`GET /agentmemory/sessions?limit=N` on the standalone/Fly-hosted server ignores `limit` entirely and always returns the complete session history, regardless of N. In our case this returned 5.38M characters for a call requesting `limit=20`, effectively flooding the caller's context window.

## Where the client already assumes server-side limiting

`dist/standalone.mjs`'s in-memory (KV/local-mode) implementation of `memory_sessions` correctly slices the result before returning it:

```js
case "memory_sessions": {
  const sessions = await kvInstance.list("mem:sessions");
  const limit = v.limit ?? 20;
  return textResponse({ sessions: sessions.slice(0, limit) }, true);
}
```

But the proxy path (used whenever an `AGENTMEMORY_URL` server, e.g. a Fly deployment, is reachable) just forwards the query param and trusts the server to honor it:

```js
case "memory_sessions": return textResponse(await handle.call(`/agentmemory/sessions?limit=${v.limit}`, { method: "GET" }), true);
```

The server-side `/agentmemory/sessions` REST handler does not appear to actually apply `limit` when building its response, so the proxy path returns everything every time.

## Impact

Any client relying on `limit` to bound response size (e.g. an MCP tool call meant to return a handful of recent sessions) instead receives the entire session history. For a long-lived store this can be multiple megabytes of text returned for what should be a small, bounded call — degrading or breaking downstream consumers that assume a capped response.

## Suggested fix

Apply the same defensive slice server-side in the `/agentmemory/sessions` REST handler that the client's local/KV-mode code path already applies, so `limit` is honored regardless of whether a caller goes through the proxy or local mode.

## Workaround applied locally

We've patched our local vendored copy of `dist/standalone.mjs` to slice the proxy response client-side after receiving it (mirroring the local-mode implementation), as a stopgap until the server enforces `limit` itself:

```js
case "memory_sessions": {
  const data = await handle.call(`/agentmemory/sessions?limit=${v.limit}`, { method: "GET" });
  const sessions = Array.isArray(data?.sessions) ? data.sessions.slice(0, v.limit) : data?.sessions;
  return textResponse({ ...data, sessions }, true);
}
```

This is a client-side workaround only — it still fetches and discards the full payload over the wire, so the real fix belongs server-side.

Issue #139, opened 2026-04-14
Claude Code: MCP not installed, skill commands erroring with sandbox


stefanfaur:
I have installed the Claude Code plugin, but no MCP server was added by it. It's not mentioned in the Readme for Claude Code either (I can install it manually probably, but feel like it should have been in the plugin itself).
Also, when trying to use the skills with sandbox enabled, I get an error that can only be bypassed by Claude figuring out to do the curl call manually, which seems inefficient:

```
Error: Shell command permission check failed for pattern "!`QUERY=$(echo "agentmemory setup" | sed 's/\\/\\\\/g; s/"/\\"/g') && curl -s
     -H "Content-Type: application/json" -H "Authorization: Bearer ${AGENTMEMORY_SECRET:-}" -X POST
     http://${AGENTMEMORY_URL:-localhost:3111}/agentmemory/search -d "{\"query\": \"${QUERY}\", \"limit\": 10}" 2>/dev/null || echo
     '{"results":[]}'`": Contains expansion
```

rohitg00:
Both real bugs. Pulling fixes today.

## 1. MCP server not auto-registered by the plugin

You're right — I ship the Claude Code plugin (\`plugin/\` → hooks + skills) and the MCP server (\`@agentmemory/mcp\` on npm) as two separate things, and the README tells Claude Code users to wire them up manually. That's a split-brain UX. The plugin should drop a \`.mcp.json\` at the plugin root so \`/plugin install agentmemory@agentmemory\` wires the MCP server automatically.

Fix: I'll add \`plugin/.mcp.json\` with:

```json
{
  "mcpServers": {
    "agentmemory": {
      "command": "npx",
      "args": ["-y", "@agentmemory/mcp"]
    }
  }
}
```

Claude Code plugins auto-start MCP servers listed there when the plugin is enabled. No user action needed after install.

## 2. Skill commands fail under sandbox with "Contains expansion"

Also real. \`recall\` and \`session-history\` use \`!\`$(curl -s ...)\`\` pre-execution blocks with \`$(...)\` and \`${VAR:-default}\` expansion, which Claude Code's sandbox rejects by pattern match before substitution. The original intent was to JSON-escape \`$ARGUMENTS\` with \`sed\` inline, but that approach is a dead end under sandbox mode.

[...]

PR #1033, opened 2026-07-09
fix(api): honour limit query param on /sessions and /replay/sessions (#1022)

ziishanahmad:
## What

The HTTP endpoints `GET /agentmemory/sessions` and `GET /agentmemory/replay/sessions` were returning the full session table on every call, ignoring any `?limit=N` query parameter. The MCP tool `memory_sessions` already supports `limit` (added in #139) but the HTTP twins did not.

## Why

The real-time viewer and CLI both want to page through sessions without forcing the engine to ship every record on every request. Returning the full table is wasteful for projects with thousands of sessions and the existing MCP tool already does the right thing — the HTTP path was the gap.

## How

- `api::sessions` (src/triggers/api.ts) now reads `?limit=N` and slices the response. Negative or non-numeric values fall back to the full list so a bad query never produces an empty result.
- `api::replay::sessions` (same file) gets the same treatment. The replay endpoint had the same gap.

## Tests

Added `test/sessions-limit.test.ts` with 4 source-level assertions (matches the style of `test/memories-pagination.test.ts`). Full suite: 1372/1416 pass (44 pre-existing Windows HOME/USERPROFILE failures in upstream, unrelated to this change).

## Notes for reviewers

I am not opening a CLA but I will sign off per the DCO. Tests added lock the behaviour to the source so a future refactor that drops the limit will fail CI. The fix is the smallest change that closes the gap and matches the existing MCP behaviour.

Signed-off-by: Zeeshan Ahmad <ziishanahmad@gmail.com>


ziishanahmad:
Superseded by a clean PR rebased on upstream main to keep the diff focused on the bug fix. New PR coming in a moment.

PR #1046, opened 2026-07-10
fix(viewer): surface health status from non-2xx /agentmemory/health responses

### LINKED ISSUE #1019: Viewer shows unknown health while /agentmemory/health reports critical
## Summary
When visiting the AgentMemory web viewer (`http://127.0.0.1:3113`) on a busy/critical service state, the Dashboard/health indicator can show `unknown` even though the backend health endpoint reports `critical`.

## Environment
- Date checked: 2026-07-06
- AgentMemory version: `0.11.2`
- Endpoint version: `0.9.26` (viewer)
- Local ports: service `3111`, viewer `3113`
- OS: Windows 10

## Repro steps
1. Open `http://127.0.0.1:3113/#memories` and observe health/status badge in UI.
2. Query backend health directly:
   ```powershell
   Invoke-RestMethod -Uri http://127.0.0.1:3111/agentmemory/health | ConvertTo-Json -Depth 8
   ```
3. Query viewer root / health endpoints:
   ```powershell
   Invoke-RestMethod -Uri http://127.0.0.1:3113
   Invoke-RestMethod -Uri http://127.0.0.1:3113/api/health
   ```

## Observed
- Viewer root call returns `404` (likely no response body).
- Viewer `api/health` returns `404`.
- Backend `/agentmemory/health` returns valid JSON with:
  - `status: critical`
  - `service: agentmemory`
  - `health.status: critical`
  - `health.alerts`: `memory_critical_98%_rss744mb`
  - `health.connectionState: connected`
  - Very high latencies (`mem::compress` avg ~6,101,855ms, `mem::summarize` avg ~5,383,435ms)
  - Very high active invocations (`workers[0].active_invocations: 669` in sampled check)
  - Function metrics showing heavy failure load (`mem::compress` failureCount ~5203)

## Expected
The UI should reflect the backend health state (`critical` in this case) or a known error state from `/agentmemory/health` without defaulting to `unknown`.

## Actual
UI reported/appeared as `unknown` health while backend endpoint is reachable and explicitly reporting degraded state.

## Impact
- False sense of service availability when core observability is degraded.
- Hard to distinguish “UI polling path issue” vs “actual daemon down” for troubleshooting.

## Suggested fix
- Confirm the viewer health polling endpoint used by UI for status is alive and compatible with the service state response.
- If viewer is not authoritative, show an explicit degraded state (e.g., backend reachable + degraded) rather than unknown.
- Add a light guard in the frontend to avoid unknown fallback during transient 404/timeouts and surface last-known health + poll errors.

viditchess64:
## What
The viewer's shared `api()` fetch helper discarded the response body whenever
the HTTP status wasn't 2xx, returning `null`. `/agentmemory/health` intentionally
returns HTTP 503 when status is "critical" (still with a valid JSON body), so
the dashboard's health badge always fell back to "unknown" instead of showing
"critical".

## Fix
`api()` now attempts to parse and return the JSON body on non-ok responses too,
falling back to `null` only if the body isn't valid JSON. This lets
`renderDashboard()` read the real `status` field regardless of which HTTP
status code carried it.

## How to verify
1. Point the viewer at a backend reporting critical health (or mock `fetch`
   to return `{ok:false, status:503, json: () => ({status:"critical", ...})}`).
2. Before this change: dashboard health badge shows "unknown".
3. After this change: dashboard health badge shows "critical".

Existing viewer test suite (`viewer-session-id.test.ts`,
`viewer-security.test.ts`, `viewer-host.test.ts`) passes unchanged.

Fixes #1019

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of non-successful API responses by returning preserved, parsed JSON error details when available.
  * Updated the viewer’s dashboard health loading to retrieve and display health information correctly even when the health endpoint responds with a non-2xx status.

### DIFF
--- a/src/viewer/index.html
+++ b/src/viewer/index.html
@@ -1248,7 +1248,16 @@
         if (!res.ok) {
           if (res.status === 401) showViewerAuthPrompt();
           console.warn('[viewer] API ' + (fetchOpts.method || 'GET') + ' ' + path + ' returned ' + res.status);
-          return null;
+          // Some endpoints (e.g. health) intentionally respond with a non-2xx
+          // status while still returning a valid JSON body describing the
+          // degraded state (see #1019). Surface that body instead of
+          // discarding it as null, which used to make a "critical" health
+          // response render as "unknown".
+          try {
+            return await res.json();
+          } catch (parseErr) {
+            return null;
+          }
         }
         hideViewerAuthPrompt();
         return await res.json();
--- a/src/viewer/index.html
+++ b/src/viewer/index.html
@@ -1256,6 +1256,7 @@
           try {
             return await res.json();
           } catch (parseErr) {
+            console.debug('[viewer] API ' + path + ' non-2xx body was not JSON:', parseErr);
             return null;
           }
         }
--- a/src/viewer/index.html
+++ b/src/viewer/index.html
@@ -1244,15 +1244,18 @@
           headers.Authorization = 'Bearer ' + viewerToken;
         }
         var fetchOpts = Object.assign({}, opts || {}, { headers: headers });
+        var readErrorBody = fetchOpts.readErrorBody;
+        delete fetchOpts.readErrorBody;
         var res = await fetch(url, fetchOpts);
         if (!res.ok) {
           if (res.status === 401) showViewerAuthPrompt();
           console.warn('[viewer] API ' + (fetchOpts.method || 'GET') + ' ' + path + ' returned ' + res.status);
-          // Some endpoints (e.g. health) intentionally respond with a non-2xx
-          // status while still returning a valid JSON body describing the
-          // degraded state (see #1019). Surface that body instead of
-          // discarding it as null, which used to make a "critical" health
-          // response render as "unknown".
+          // Non-2xx responses resolve to null so callers can keep treating
+          // null as "request failed" (e.g. loadGraph's disabled/error state).
+          // The health endpoint opts out via readErrorBody: it intentionally
+          // responds 503 with a valid JSON body when status is "critical"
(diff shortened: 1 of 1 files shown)


rohitg00 (changes requested):
The underlying bug is real: `/agentmemory/health` returns HTTP 503 when status is "critical" (`statusCode = status === "critical" ? 503 : 200`) with a valid JSON body, and the viewer's shared `api()` discards non-2xx bodies, so the health badge falls back to "unknown". Good catch on #1019.

The concern is scope. This changes the shared `api()` helper, so every non-2xx response now returns a parsed object instead of `null`, which breaks callers that treat `null` as "request failed". Concretely, `loadGraph()` checks `queryResult === null` to render the disabled/error state, so a 503 `graphDisabledResponse()` (and other error bodies) would now be read as real data and skip the error banner / rebuild path. Same class as the `if (!res) return` guards elsewhere in the viewer.

Could you keep the `non-2xx -> null` contract in the shared `api()` and special-case health instead? Either a dedicated fetch for `/agentmemory/health` that reads the body regardless of status, or have `api()` return `{ status, body }` and check the status at the health caller. That fixes the badge without regressing the graph and error paths.

viditchess64:
Good catch — agreed, changing the shared helper regressed the null contract for loadGraph and the other `if (!res)` guards. Pushed bbf111f: `api()` returns null on non-2xx again by default; only the health call opts in via a `readErrorBody` flag. All viewer tests pass.

rohitg00:
Thanks for your contribution, and quick fix @viditchess64

rohitg00 (approved):

PR #1061, opened 2026-07-15
fix: dedupe double-triggered summaries and unchanged-input re-processing on session stop

### LINKED ISSUE #1062: Session stop triggers mem::summarize twice; unchanged sessions are fully re-summarized on every stop
## Summary

Every session stop runs `mem::summarize` **twice**, and each run re-summarizes the entire session from scratch. For long-lived sessions that stop repeatedly (e.g. periodic/heartbeat-style automations), this compounds into very large duplicate LLM bills.

## Root cause

The stop hook (`src/hooks/stop.ts`) fires two requests:

1. `POST /agentmemory/summarize` → `api::summarize` → `mem::summarize`
2. `POST /agentmemory/session/end` → `api::session::end` → fans out `event::session::stopped` (`src/triggers/api.ts`), whose handler calls `mem::summarize` again (`src/triggers/events.ts`)

Both paths always fire on every stop, and `mem::summarize` (`src/functions/summarize.ts`) has:

- no per-session lock / in-flight guard (unlike `mem::observe`, which uses `DedupMap` + `withKeyedLock`)
- no "input unchanged" short-circuit — the only skip is `compressed.length === 0`

So a single stop = 2 full-history LLM runs, and every subsequent stop re-summarizes the whole history again even when nothing changed.

## Observed impact (self-hosted, v0.9.27)

Provider-side logs show byte-identical summarize request pairs on every stop of a session that woke every 30 minutes, growing each cycle as observations accumulated:

```
03:24:33  222635B   ← same request
03:24:33  222635B   ← twice, same second
03:54:27  229115B
03:54:27  229115B
04:24:22  235636B
04:24:22  235636B
...
```

By the end this was ~110K input tokens per call, 3 heavy calls per 30-minute cycle, with the LLM output being an equivalent summary each time (identical input, nondeterministic titles — sessions also show 2–3 near-identical stored summaries generated seconds apart).

## Notes

- A time-window dedup like #572 helps but does not close this: the two requests arrive in the same second and can both pass a freshness check before either writes a summary, and after the window expires an unchanged session is still fully re-summarized.
- Fix proposed in #1061: remove the duplicate trigger in the stop hook, serialize per-session runs with `withKeyedLock`, and skip when the stored summary's `observationCount` already matches (observations are append-only, so an unchanged count means identical LLM input).

### LINKED ISSUE #1063: Session stop re-runs graph extraction over the full observation set even when nothing changed
## Summary

The `event::session::stopped` handler sends the session's **entire** compressed observation set to `mem::graph-extract` on every session stop, with no change detection. Sessions that stop repeatedly without new activity (heartbeat-style loops) re-extract the identical set every time.

## Root cause

`src/triggers/events.ts` (session-stopped handler):

```ts
if (isGraphExtractionEnabled()) {
  const observations = await kv.list<CompressedObservation>(
    KV.observations(data.sessionId),
  );
  const compressed = observations.filter((o) => o.title);
  if (compressed.length > 0) {
    sdk.trigger({
      function_id: "mem::graph-extract",
      payload: { observations: compressed },   // full set, every stop
      action: TriggerAction.Void(),
    });
  }
}
```

There is no fingerprint/dedup — every stop re-processes the full history even when zero observations were added since the last extraction.

## Observed impact (self-hosted, v0.9.27)

A session capped at 500 observations that stopped every 30 minutes produced a ~213 KB (~50K input tokens) graph-extraction LLM request per cycle, around the clock, with the input identical each time (extraction yields mostly already-known nodes/edges).

This is the same shape as the double/redundant summarize issue (#1062) — full re-processing on session stop with no change detection — just in a different function.

## Fix

Included in #1061 (shares the `computeInputFingerprint` helper introduced there): fingerprint the compressed set, store it per session (`KV.graphExtractState`), and trigger `mem::graph-extract` only when the fingerprint changed.

yulin0629:
Fixes #1062. Fixes #1063.

## Problem

Every session stop currently runs `mem::summarize` **twice**, and each run re-summarizes the whole session from scratch.

The stop hook fires two requests:

1. `POST /agentmemory/summarize` → `api::summarize` → `mem::summarize`
2. `POST /agentmemory/session/end` → `api::session::end` → fans out `event::session::stopped`, whose handler calls `mem::summarize` again (`src/triggers/events.ts`)

`mem::summarize` has no lock, no in-flight guard, and no unchanged-input short-circuit (unlike `mem::observe`, which already uses `DedupMap` + `withKeyedLock`), so both runs do the full LLM work.

Observed in production (self-hosted, v0.9.27): every session stop produced **two byte-identical summarize requests** at the provider (same second, same body size), growing each cycle as the session accumulated observations — e.g. a heartbeat-style session that stopped every 30 minutes reached duplicate ~230 KB (~110K input tokens) requests per cycle:

```
03:24:33  claude-sonnet-5  222635B   ← same request
03:24:33  claude-sonnet-5  222635B   ← twice
03:54:27  claude-sonnet-5  229115B
03:54:27  claude-sonnet-5  229115B
04:24:22  claude-sonnet-5  235636B
04:24:22  claude-sonnet-5  235636B
```

## Fix

- **stop hook**: drop the direct `/agentmemory/summarize` call — `/agentmemory/session/end` already fans out to the summarize path. One stop, one summarize.
- **`mem::summarize`**: serialize per-session runs with `withKeyedLock` (same pattern as `mem::observe`), and short-circuit when the stored summary's `observationCount` already matches the current compressed count. Observations are append-only, so an unchanged count means the LLM input would be byte-identical to the previous run — re-running cannot produce new information. This also stops full re-summarization of long-lived sessions that stop repeatedly without new activity.

## Relation to #572

#572 adds a time-window dedup (`SUMMARIZE_DEDUP_WINDOW_MS`, default 90 s). That helps, but:

- it does not remove the duplicate trigger itself;
- a pure time check does not close the race — the two stop-hook requests arrive in the same second and can both pass the freshness check before either writes a summary (the duplicate pairs above are exactly this case), whereas a keyed lock makes the second run wait and then hit the short-circuit;
- after the window expires, an unchanged session is still fully re-summarized.

The two changes are complementary: count-based skip covers "nothing new" regardless of elapsed time, and a time window could still be layered on top for sessions that keep accumulating low-value observations.

## Also: graph extraction has the same shape (#1063)

The session-stopped handler sent the session's full compressed observation set to `mem::graph-extract` on every stop with no change detection — for a 500-observation heartbeat session that was a ~50K-token LLM call every 30 minutes with identical input. The last commit reuses the same fingerprint helper: store the set's fingerprint per session (`KV.graphExtractState`) and trigger extraction only when it changed.

## Tests

Three new cases in `test/summarize.test.ts`:

- unchanged observation count → second call returns the stored summary with `skipped: true`, no second LLM call
- two concurrent calls for the same session collapse into a single LLM run
- new observations since the last summary → re-summarizes as before

`npm test`: no new failures (the 4 pre-existing failures in `auto-compress.test.ts` / `embedding-provider.test.ts` reproduce on a clean checkout and come from environment leakage, unrelated to this change).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate session summaries and graph extraction when inputs are unchanged.
  * Updated session shutdown to trigger summarization once, avoiding duplicate processing.
  * Improved reflection scalability by using capped graph snapshots and graceful fallbacks.
  * Improved memory health severity calculations using the runtime heap limit.
* **Tests**
  * Expanded coverage for summarization deduplication, concurrency, input changes, and timeouts.
  * Added graph-extraction deduplication and health-threshold regression tests.
* **Documentation**
  * Added unreleased changelog entries for reflection scalability and health monitoring improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- **`mem::reflect` no longer enumerates the graph node scope.** `kv.list(KV.graphNodes)` stopped completing once `mem:graph:nodes` grew past ~100 MB, and the failure was swallowed by a silent `.catch(() => [])`. Graph clustering then returned nothing, reflect fell back to lexical clustering, and a cluster matching thousands of facts produced a prompt large enough to exceed the 180s invocation timeout — every run, silently disabling insight generation. Reflect now reads the top-degree graph snapshot (the same fast path graph extraction already used), cluster inputs are capped at 50 items per category, and the remaining `kv.list` calls log a warning instead of degrading silently.
+- **Health no longer pins itself to `critical` on a healthy process.** Memory severity divided `heapUsed` by `heapTotal`, but `heapTotal` is the heap V8 has committed so far, which it deliberately keeps near-full and grows on demand. Any process past the RSS floor therefore reported ~100% and `/agentmemory/health` returned 503 indefinitely, making it useless as a liveness signal. The snapshot now carries `heapLimit` from `v8.getHeapStatistics()` and the ratio is measured against that ceiling; snapshots persisted without it still evaluate as before.
+
 ## [0.9.27] — 2026-06-07
 
 Wave release closing several breaking regressions reported against v0.9.26, plus an agent-scope isolation security fix, an iii version-pin audit fix, and a benchmark scorecard correction. No breaking changes; drop-in upgrade.
--- a/plugin/scripts/stop.mjs
+++ b/plugin/scripts/stop.mjs
@@ -23,12 +23,6 @@ async function main() {
 	}
 	if (isSdkChildContext(data)) return;
 	const sessionId = data.session_id || data.sessionId || "unknown";
-	fetch(`${REST_URL}/agentmemory/summarize`, {
-		method: "POST",
-		headers: authHeaders(),
-		body: JSON.stringify({ sessionId }),
-		signal: AbortSignal.timeout(12e4)
-	}).catch(() => {});
 	fetch(`${REST_URL}/agentmemory/session/end`, {
 		method: "POST",
 		headers: authHeaders(),
@@ -38,7 +32,7 @@ async function main() {
 	setTimeout(() => process.exit(0), 1500).unref();
 }
 main();
-
 //#endregion
-export {  };
+export {};
+
 //# sourceMappingURL=stop.mjs.map
--- a/plugin/scripts/stop.mjs
+++ b/plugin/scripts/stop.mjs
@@ -29,7 +29,7 @@ async function main() {
 		body: JSON.stringify({ sessionId }),
 		signal: AbortSignal.timeout(5e3)
 	}).catch(() => {});
-	setTimeout(() => process.exit(0), 1500).unref();
+	setTimeout(() => process.exit(0), 500).unref();
 }
 main();
 //#endregion
--- a/src/functions/graph.ts
+++ b/src/functions/graph.ts
@@ -76,7 +76,9 @@ function emptySnapshot(): GraphSnapshot {
   };
 }
 
-async function readSnapshot(kv: StateKV): Promise<GraphSnapshot | null> {
+export async function readSnapshot(
+  kv: StateKV,
+): Promise<GraphSnapshot | null> {
   try {
     const snap = await kv.get<GraphSnapshot>(KV.graphSnapshot, SNAPSHOT_KEY);
     if (snap && typeof snap === "object" && snap.version === 1) {
--- a/src/functions/input-fingerprint.ts
+++ b/src/functions/input-fingerprint.ts
@@ -1,0 +1,14 @@
+import { createHash } from "node:crypto";
+import type { CompressedObservation } from "../types.js";
+
+// Stable fingerprint of an exact observation set, used to skip LLM re-runs
+// (summarize, graph extraction) when the input is unchanged. Count alone is
+// not enough: evict/auto-forget can delete observations, so the set can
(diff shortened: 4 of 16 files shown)


devon3000:
Rebased this onto current `main` (`2d38daf`, 0.9.29) — full suite green, 1668 passed / 0 failed. Branch: https://github.com/devon3000/agentmemory/tree/pr-1061-rebased

Three conflicts, all mechanical:

- **`src/hooks/stop.ts` / `plugin/scripts/stop.mjs`** — `main` has since dropped the direct `/agentmemory/summarize` call on its own (comment cites #1203), so that half of this PR is already in. Kept main's `conversation_id` fallback and `main().catch(...)`.
- **`src/triggers/events.ts`** — the `event::session::stopped` handler was rewritten upstream (adds a local `fireVoid` helper plus a consolidation cooldown). Folded the graph fingerprint into the new structure, dispatching through `fireVoid` so it picks up that helper's failure logging.
- **`test/session-end-triggers-graph.test.ts`** — the two dedup assertions match the trigger call site as literal source text, so they failed on the `fireVoid` spelling even though the ordering they assert (fingerprint computed and persisted before `mem::graph-extract` fires) still holds. Widened the regex to accept either spelling.

Verified the summarize guard live against a self-hosted 0.9.29 (Anthropic provider, session pinned at `MAX_OBS_PER_SESSION=500`):

- repeat call → `skipped: true` in 0.024s, zero LLM calls
- two **concurrent** calls → both return at 25.5s with exactly one real LLM run; the second waits on the keyed lock and then hits the fingerprint skip

That second case is the one a pure time-window dedup (#572) can't close, which matches your reasoning in the description — worth keeping the lock rather than folding this into a freshness window.

One behavioural note worth a changelog line: summaries written before `inputFingerprint` existed carry no fingerprint, so each session re-summarizes once after upgrade before skips kick in.

Issue #991, opened 2026-06-29
SessionEnd hook reports "Hook cancelled" on exit (1.5s deferred exit outlives shutdown grace)


JaeyoulLee:
## Summary

The `SessionEnd` hook intermittently reports **`Hook cancelled`** in Claude Code when the user exits a session:

```
SessionEnd hook [node "${CLAUDE_PLUGIN_ROOT}/scripts/session-end.mjs"] failed: Hook cancelled
```

## Root cause

In `src/hooks/session-end.ts` (built to `scripts/session-end.mjs`), the hook fires best-effort POSTs to the local engine and then defers exit:

```js
setTimeout(() => process.exit(0), 1500).unref();
```

Two problems:

1. **Outlives the shutdown grace.** On `exit`, Claude Code gives `SessionEnd` hooks only a short grace window. The 1.5s deferred exit (plus any pending `fetch` socket keeping the event loop alive) means the hook process is still running when the harness tears down → the harness kills it and surfaces `Hook cancelled`. The session-end telemetry POST usually still goes out, so it's cosmetic — but it's noisy on every exit.

2. **`.unref()` defeats the intended cap.** Because the in-flight `fetch` sockets keep the loop alive, the unref'd timer is the only thing forcing exit; if the engine is slow/unreachable the process can linger longer than expected before the timer fires.

## Reproduction

- Node 22, agentmemory `0.9.27`, Claude Code plugin install.
- Exit a session (`exit`). Intermittently shows the `Hook cancelled` line.

## Fix that worked locally

Shorten the deferred exit so the hook quits well inside the harness grace, while still giving the localhost POST time to flush:

```js
setTimeout(() => process.exit(0), 500);
```

This eliminated the `Hook cancelled` message in my testing while preserving the fire-and-forget POST. A more robust option would be to exit as soon as the primary `/session/end` request settles, with a small hard cap.

## Minor adjacent note

`scripts/session-end.mjs` uses global `fetch`, which is absent on Node < 18 (e.g. Node 12 → `ReferenceError: fetch is not defined`, unhandled rejection). A `typeof fetch === 'function'` guard (or an explicit Node version requirement) would fail more gracefully on systems where the hook is spawned under an older `node` on PATH.

Affects: same pattern likely applies to other hook scripts that use the `setTimeout(...).unref()` exit idiom.

PR #1070, opened 2026-07-16
fix(session-end): remove .unref() so process force-exits after 1500ms

### LINKED ISSUE #1069: bug: session-end hook causes 'Hook cancelled' due to .unref() making setTimeout passive while fetch keeps event loop alive
## Bug

When closing Claude Code, users intermittently see:

```
SessionEnd hook [node "${CLAUDE_PLUGIN_ROOT}/scripts/session-end.mjs"] failed: Hook cancelled
```

## Root cause

`session-end.mjs` fires fire-and-forget `fetch()` calls to the local daemon and then does:

```js
setTimeout(() => process.exit(0), 1500).unref();
```

The intent (per `AGENTS.md`) is that the process force-exits after 1500ms so it does not block Claude Code's shutdown. But `.unref()` makes the timer passive — it does not prevent process exit, but it also does not force it. Node.js keeps the event loop alive as long as there are pending network requests, and the in-flight `fetch()` calls do exactly that.

Result: the process stays alive until the fetches complete or their `AbortSignal` timeouts fire:
- `/agentmemory/session/end`: up to 30s
- `/agentmemory/crystals/auto` (when `CONSOLIDATION_ENABLED=true`): up to 60s
- `/agentmemory/consolidate-pipeline` (when `CONSOLIDATION_ENABLED=true`): up to 120s

Claude Code's hook timeout fires before the fetches settle and kills the process with `Hook cancelled`.

The bug is intermittent because when the daemon responds quickly the hook exits in time; when it is slow or busy the race is lost.

## Fix

Remove `.unref()` from the `setTimeout`. An active (ref'd) timer keeps the event loop alive and guarantees the process exits after exactly 1500ms regardless of in-flight fetches. Since the fetches are fire-and-forget to a local daemon, exiting before they settle is the correct and intended behavior — matching what `AGENTS.md` documents.

```diff
- setTimeout(() => process.exit(0), 1500).unref();
+ setTimeout(() => process.exit(0), 1500);
```

Fix is in PR #1070.

dnraitzyk:
Fixes #1069.

## What

Remove `.unref()` from the `setTimeout` in `session-end.mjs` / `session-end.ts`.

## Why

The `session-end` hook fires fire-and-forget `fetch()` calls to the local daemon, then does:

```js
setTimeout(() => process.exit(0), 1500).unref();
```

`AGENTS.md` documents the intent: the process should force-exit after 1500ms so it does not block Claude Code's shutdown. But `.unref()` makes the timer passive — Node.js does not keep the event loop alive for an unref'd timer, but it also will not fire it while other handles (the pending `fetch()` calls) are keeping the loop alive.

The fetches use `AbortSignal.timeout()` values of 30s, 60s, and 120s. Those pending requests keep the event loop alive the entire time. Claude Code's hook deadline fires first and kills the process with:

```
SessionEnd hook [node "${CLAUDE_PLUGIN_ROOT}/scripts/session-end.mjs"] failed: Hook cancelled
```

This is intermittent because when the daemon responds fast the event loop drains before the deadline; when it is slow the race is lost.

## Fix

Remove `.unref()`. An active timer is the only handle that keeps the event loop alive after the fetches are dispatched, so the process exits after exactly 1500ms regardless of whether the fetches have settled. This matches the documented intent and the fire-and-forget pattern `AGENTS.md` prescribes.

```diff
- setTimeout(() => process.exit(0), 1500).unref();
+ setTimeout(() => process.exit(0), 1500);
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session-end handling so processes terminate reliably after the session ends.
  * Prevented lingering network requests from delaying shutdown beyond the expected timeout.

### DIFF
--- a/plugin/scripts/session-end.mjs
+++ b/plugin/scripts/session-end.mjs
@@ -51,7 +51,7 @@ async function main() {
 		headers: authHeaders(),
 		signal: AbortSignal.timeout(3e4)
 	}).catch(() => {});
-	setTimeout(() => process.exit(0), 1500).unref();
+	setTimeout(() => process.exit(0), 1500);
 }
 main();
 
--- a/src/hooks/session-end.ts
+++ b/src/hooks/session-end.ts
@@ -63,7 +63,11 @@ async function main() {
     }).catch(() => {});
   }
 
-  setTimeout(() => process.exit(0), 1500).unref();
+  // Force-exit after 1500ms regardless of in-flight fetches. Without this,
+  // .unref() makes the timer passive and pending fetch() calls keep the event
+  // loop alive until their AbortSignal timeouts fire (up to 120s), causing
+  // Claude Code to cancel the hook with "Hook cancelled".
+  setTimeout(() => process.exit(0), 1500);
 }
 
 main();


dnraitzyk:
How do approvals work?

sla-te:
I hit the same "Hook cancelled" message and went looking for this PR, but the measurements do not support the diagnosis. This change makes the healthy case 27x slower and leaves the failing case unchanged.

`.unref()` does not make a timer passive in the sense assumed here. It only stops the timer from keeping the event loop alive **by itself**. The timer still fires normally as long as something else holds the loop open, and the pending `fetch()` is exactly that. So the 1500 ms guard already fires and already bounds the hook.

Measured on the shipped `session-end.mjs` (0.9.29 dist, node v24.19.0), same input on stdin, wall clock around the process:

| daemon state | current (`.unref()`) | this PR (no `.unref()`) |
| --- | --- | --- |
| responding normally | **56 ms** | **1549 ms** |
| accepts the connection, never answers | 1548 ms | 1548 ms |
| SYN dropped, connect never completes | 1546 ms | 1548 ms |

Two things follow. The hook never waits for the 30s/60s/120s `AbortSignal` timeouts, so that part of the root cause in #1069 does not happen. And removing `.unref()` cannot fix the report, because the failing case is already exiting at ~1.5 s. All the PR does is drag the fast path down to the same 1.5 s.

What actually produces the message is that ~1.5 s is longer than the shutdown grace Claude Code gives a SessionEnd hook, which is how #991 frames it. On my box the hook only crossed that line while the daemon was wedged (accept queue full, `Recv-Q 129 > backlog 128`, engine stuck in `futex_do_wait` at 12.9 GB RSS). Restarting the daemon put it back to 56 ms and the message stopped.

So the fix has to shorten the worst case, not force the best case to it. Lowering the guard on `session-end` and `stop` to something safely under the grace window would do it and keeps `.unref()` doing its job. The 1500 ms in d626b4ea was chosen to let several fetches get initiated, and initiating them is fast, so there looks to be plenty of room below it.

Reproduction, if useful: point `AGENTMEMORY_URL` at `nc -l 127.0.0.1 39111` for the accepts-but-silent case and at an address that drops SYNs for the other, then time `echo '{"session_id":"x","reason":"other"}' | node scripts/session-end.mjs`.

Is there a documented number for the SessionEnd grace window, or does the cap have to be found empirically?

PR #1101, opened 2026-07-22
fix(api): join session summaries via kv.list instead of per-session kv.get fan-out

### LINKED ISSUE #1100: /agentmemory/sessions deadlocks permanently: api::sessions fires one kv.get per session in a single Promise.all (leaks invocations, dashboard shows 0 sessions / 0% token savings)
## Summary

`GET /agentmemory/sessions` hangs **permanently** (no response after 180s+, `curl` exits with HTTP 000) once the session count grows into the hundreds. Every attempt additionally **leaks pending invocations** on the worker, so a dashboard left open (which polls this endpoint) drives the server into `degraded` state. Because the viewer dashboard and `agentmemory status` both build their stats from this endpoint, the visible symptoms are:

- Viewer dashboard: **Sessions 0, Token Savings 0%** (independent of the formula bug in #944 — the sessions list never loads at all)
- `agentmemory status`: `Sessions: 0`, `Observations: 0`, token-savings section omitted
- `health.workers[].active_invocations` climbing without bound (I observed 327 → 1,403 within ~10 minutes), heap at 94%, `status: degraded` — which looks a lot like the unresponsiveness reported in #499

## Environment

- agentmemory **v0.9.28** (npm global install), macOS 15 (arm64), Node v25.8.1
- 615 sessions / 9,525 observations / 114 summaries in the store
- Store sizes are small: `mem:sessions` 180 KB, `mem:summaries` 398 KB — data volume is not the problem

## Root cause

The `api::sessions` handler fires **one `kv.get(KV.summaries, id)` per session inside a single `Promise.all`**:

```js
const sessions = await kv.list(KV.sessions);
// ...
const summaries = await Promise.all(
  filtered.map((s) => kv.get(KV.summaries, s.id).catch(() => null))
);
```

With 615 sessions that is a burst of 615 concurrent kv calls from one worker function. At that size the engine/worker channel never services them — the promises neither resolve nor reject (there is no timeout on `kv.get`, and `.catch` only handles rejections), so the handler waits forever and each request permanently parks its in-flight kv invocations on the worker.

## Evidence / bisection

| Test | Result |
|---|---|
| `GET /agentmemory/sessions` (615-way fan-out) | hangs, HTTP 000 after 180s |
| `GET /agentmemory/sessions?agentId=zz-nonexistent` (filter empties the list **before** the fan-out → 0 gets) | **HTTP 200 in 0.17s** |
| `GET /agentmemory/replay/sessions` (single `kv.list(KV.sessions)`, no fan-out) | HTTP 200 in 0.05s |
| `memory_diagnose` categories `sessions,summaries` (sequential `kv.list` calls, reads every summary value) | passes, "All 114 session summaries are consistent" |
| `mem::context` (same fan-out pattern but sliced to **10** sessions) | works fine |
| worker `active_invocations` after 3 hung `/sessions` attempts | +~600 per attempt, never drains |

So: listing works, every summary value is readable, small bounded fan-outs work — only the unbounded concurrent `kv.get` burst deadlocks.

## Fix

Replace the per-session fan-out with a single `kv.list` + in-memory join (same pattern `mem::diagnose` already uses; summaries carry `sessionId`):

```diff
 const filtered = filterAgentId ? sessions.filter((s) => s.agentId === filterAgentId) : sessions;
-const summaries = await Promise.all(filtered.map((s) => kv.get(KV.summaries, s.id).catch(() => null)));
+const summariesById = new Map((await kv.list(KV.summaries).catch(() => [])).map((sum) => [sum.sessionId, sum]));
 return {
   status_code: 200,
-  body: { sessions: filtered.map((s, i) => summaries[i] ? { ...s, summary: summaries[i] } : s) }
+  body: { sessions: filtered.map((s) => summariesById.has(s.id) ? { ...s, summary: summariesById.get(s.id) } : s) }
 };
```

Verified locally by patching the bundled dist and restarting the worker:

- `/agentmemory/sessions`: 0.02–0.08s, HTTP 200, 615 sessions with 114 summaries attached
- `agentmemory status`: `Sessions: 615, Observations: 9,526, Token savings: ~760,180 tokens saved (100% reduction)`
- `active_invocations`: back to single digits and stable while polling the dashboard

More generally it would be worth (a) bounding concurrency anywhere a worker function fans out kv calls over an unbounded collection, and (b) adding a timeout to kv operations so a stalled call fails instead of hanging the HTTP request and leaking the invocation.

Happy to open a PR if useful.

Codejune:
Fixes #1100

## Problem

`api::sessions` fires one `kv.get(KV.summaries, id)` per session inside a single `Promise.all`. Once the session count grows into the hundreds (615 in my store), the burst of concurrent kv calls is never serviced: the promises neither resolve nor reject (no timeout on `kv.get`, and `.catch` only handles rejections), so `GET /agentmemory/sessions` hangs permanently (HTTP 000 after 180s+) and every attempt parks its in-flight kv invocations on the worker — `active_invocations` climbs without bound (327 → 1,403 while a dashboard was polling), heap hits ~94%, health goes `degraded`. Dashboard and `agentmemory status` then show 0 sessions / 0% token savings.

Full bisection evidence is in #1100 (zero-fanout filter responds in 0.17s, list-only endpoints respond in ~50ms, only the unbounded fan-out deadlocks).

## Fix

Fetch summaries with a single `kv.list(KV.summaries)` and join in memory by `sessionId` — the same pattern `mem::diagnose` already uses. One engine roundtrip instead of N.

## Verification

- Patched the identical logic into the bundled dist of a live v0.9.28 install with 615 sessions / 9,525 observations / 114 summaries:
  - `/agentmemory/sessions`: 0.02–0.08s, HTTP 200, all 114 summaries attached (previously: permanent hang)
  - `agentmemory status`: `Sessions: 615, Observations: 9,526` (previously `Sessions: 0`)
  - `active_invocations` stable in single digits while polling the dashboard (previously leaking ~600 per request)
- `npx tsc --noEmit`: no new diagnostics in `src/triggers/api.ts` (the 4 pre-existing ones there are at lines 252/288/2454, untouched)
- `npm test`: 1,402 passed / 13 failed — the same 13 environment-dependent tests (embedding-provider key detection, auto-compress env gate, hook-project cwd basename, fs-watcher timing) fail identically on unmodified `main` in my environment; `test/mcp-standalone-proxy.test.ts`, the only suite touching `/agentmemory/sessions`, passes

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved session summary loading efficiency.
  * Session lists continue to include optional summaries while preserving existing filtering behavior.

### DIFF
--- a/src/triggers/api.ts
+++ b/src/triggers/api.ts
@@ -851,14 +851,18 @@ export function registerApiTriggers(
       const filtered = filterAgentId
         ? sessions.filter((s) => s.agentId === filterAgentId)
         : sessions;
-      const summaries = await Promise.all(
-        filtered.map((s) =>
-          kv.get<SessionSummary>(KV.summaries, s.id).catch(() => null),
+      // Fetch summaries with a single list + in-memory join. Fanning out one
+      // kv.get per session in a Promise.all deadlocks the engine channel once
+      // the session count grows into the hundreds (#1100).
+      const summariesById = new Map(
+        (await kv.list<SessionSummary>(KV.summaries).catch(() => [])).map(
+          (summary): [string, SessionSummary] => [summary.sessionId, summary],
         ),
       );
-      const withSummary = filtered.map((s, i) =>
-        summaries[i] ? { ...s, summary: summaries[i] } : s,
-      );
+      const withSummary = filtered.map((s) => {
+        const summary = summariesById.get(s.id);
+        return summary ? { ...s, summary } : s;
+      });
       return { status_code: 200, body: { sessions: withSummary } };
     },
   );


rohitg00:
This is fixed with better way.

PR #1107, opened 2026-07-24
fix: forward createdBy in memory_action_create MCP tool

### LINKED ISSUE #1105: memory_action_create MCP tool drops createdBy, unlike lease/frontier tools which correctly pass agentId
## Summary

`mem::action-create` (the underlying function) supports a `createdBy` field and sets it on the action record, but the `memory_action_create` MCP tool handler never forwards `args.createdBy` through to it — so every action created via the MCP tool ends up with `createdBy: "unknown"`, regardless of who actually created it.

This is inconsistent with the rest of the actions/leases feature area: `memory_lease`, `memory_frontier`, and `memory_next` all correctly expose and require/pass `agentId` through to their underlying functions, making per-agent attribution and exclusivity work as designed everywhere except at action-creation time.

## Where

- Underlying function reads `data.createdBy`: `src/functions/... mem::action-create` (`createdBy: data.createdBy || "unknown"`)
- MCP tool handler (`case "memory_action_create":` in the tools dispatcher) builds its `sdk.trigger` payload from `title`, `description`, `priority`, `project`, `tags`, `parentId`, `edges` — `createdBy` is not read from `args` and not included in the payload at all.
- Compare with `case "memory_frontier":` / `case "memory_next":` / `case "memory_lease":`, which all read and forward `args.agentId` correctly.

## Repro

1. Configure any LLM provider (local Ollama or cloud) so actions/consolidation work.
2. Call the `memory_action_create` MCP tool (no `createdBy` param exists on the tool schema at all — it isn't just optional, it's absent).
3. Inspect the created action (`GET /agentmemory/actions` or `memory_action_update`'s echoed response) — `createdBy` is always `"unknown"`.

## Impact

In a multi-agent setup (multiple Claude Code / MCP client sessions sharing one agentmemory server), this makes it impossible to tell which agent created a given action purely from the action record — undermining the attribution half of the otherwise well-built multi-agent action/lease system (leases correctly track `agentId` for exclusivity, but the originating actor of the work item itself is lost).

## Suggested fix

Add `createdBy` (optionally aliased to `agentId` for consistency with the lease/frontier tools) to the `memory_action_create` MCP tool's parameter schema, and forward it into the `mem::action-create` payload the same way `agentId` is already forwarded for leases/frontier/next.

Happy to open a PR for this if useful — it looks like a small, isolated change (tool schema + one payload field).

deemaxx:
## Summary

`memory_action_create`'s MCP tool handler rebuilds its payload to `mem::action-create` field by field and never reads `args.createdBy` — every action created through the MCP tool ends up with `createdBy: "unknown"`, regardless of who actually created it.

- `mem::action-create` already supports and correctly sets `createdBy` (tested at the function level in `test/actions.test.ts`).
- The REST endpoint (`api::action-create`) already works correctly too — it forwards the entire `req.body`, so `createdBy` passes straight through with no field-picking.
- The sibling multi-agent tools — `memory_lease`, `memory_frontier`, `memory_next` — all correctly read and forward `agentId`.

So this is an isolated gap in one tool's argument-forwarding, not a design or architecture issue — everything around it already handles agent attribution correctly.

## What it does

- Adds `createdBy` to `memory_action_create`'s input schema in `tools-registry.ts`.
- Forwards `args.createdBy` into the `mem::action-create` payload in `server.ts`.
- Adds test coverage at the `mcp::tools::call` layer (there was no existing test exercising this tool through the actual MCP dispatcher — the existing `actions.test.ts` coverage calls `mem::action-create` directly). Two tests: `createdBy` forwards correctly, and it still defaults to `"unknown"` when omitted (no behavior change for existing callers).

## How to verify

```bash
npm test -- test/actions.test.ts
```

I confirmed the new tests actually catch the bug by reverting the two source changes and re-running — the "forwards createdBy" test fails as expected (`received "unknown"`), everything else is unaffected.

## Note on the full suite

`npm test` currently has 7 pre-existing failures across `test/context-slots.test.ts`, `test/embedding-provider.test.ts`, and `test/fetch-timeout.test.ts` on a clean `main` checkout (confirmed by stashing this change and re-running) — they look like test-isolation leakage (a couple of `it` blocks in `fetch-timeout.test.ts` set `process.env["AGENTMEMORY_LLM_TIMEOUT_MS"]`/`OPENAI_TIMEOUT_MS"` directly with no `afterEach` cleanup in that `describe` block, and a later test in the same file asserts the *unset* default). None of this is related to the change here — flagging in case it's not already known, happy to open a separate issue for it if useful.

Fixes #1105

## Summary by CodeRabbit

* **New Features**
  * Added optional creator attribution when creating actions through the MCP tool.
  * Actions now store the provided agent/user identifier, defaulting to “unknown” when omitted.
* **Tests**
  * Added coverage to verify creator attribution forwarding, default behavior, and that the MCP tool advertises the new optional field in its input schema.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -678,7 +678,7 @@ export function registerMcpEndpoints(
               title: args.title,
               description: args.description,
               priority: args.priority,
-              createdBy: args.createdBy,
+              createdBy: typeof args.createdBy === "string" ? args.createdBy : undefined,
               project: args.project,
               tags,
               parentId: args.parentId,
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -678,6 +678,7 @@ export function registerMcpEndpoints(
               title: args.title,
               description: args.description,
               priority: args.priority,
+              createdBy: args.createdBy,
               project: args.project,
               tags,
               parentId: args.parentId,
--- a/src/mcp/tools-registry.ts
+++ b/src/mcp/tools-registry.ts
@@ -374,6 +374,10 @@ export const V050_TOOLS: McpToolDef[] = [
           type: "number",
           description: "Priority 1-10 (10 highest)",
         },
+        createdBy: {
+          type: "string",
+          description: "Agent or user ID creating this action",
+        },
         project: { type: "string", description: "Project path" },
         tags: {
           type: "string",
--- a/test/actions.test.ts
+++ b/test/actions.test.ts
@@ -498,4 +498,33 @@ describe("memory_action_create MCP tool", () => {
 
     expect(parsed.action.createdBy).toBe("unknown");
   });
+
+  it("advertises createdBy as an optional string in the tool schema", async () => {
+    const result = (await sdk.trigger("mcp::tools::list", {
+      body: {},
+      headers: {},
+    })) as {
+      status_code: number;
+      body: {
+        tools: Array<{
+          name: string;
+          inputSchema: {
+            properties: Record<string, { type: string }>;
+            required?: string[];
+          };
+        }>;
+      };
+    };
+
+    const tool = result.body.tools.find(
+      (t) => t.name === "memory_action_create",
+    );
+
+    expect(tool).toBeDefined();
+    expect(tool!.inputSchema.properties.createdBy).toEqual({
+      type: "string",
+      description: "Agent or user ID creating this action",
+    });
+    expect(tool!.inputSchema.required).not.toContain("createdBy");
+  });
 });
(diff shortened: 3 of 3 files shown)


deemaxx:
Addressed both review comments in 5846ac0:

- `createdBy` is now type-checked at the MCP boundary (`typeof args.createdBy === "string"`) before being forwarded — a non-string value falls back to `undefined`/`"unknown"` instead of passing through unchecked, matching the validation style used elsewhere in this handler.
- Added a schema-shape test (`mcp::tools::list` → asserts `createdBy` is present as an optional string property) so the existing dispatch tests aren't the only thing guarding this.

Confirmed both by temporarily reverting each change and watching the corresponding test fail before re-applying.

PR #1110, opened 2026-07-25
fix(openclaw): scope memory per agent and stop dropping captures

jesusweb3:
## Summary

On a two-agent OpenClaw setup the agentmemory integration stored nothing at all, and once it did store something, every agent read and wrote the same pool. Digging into it turned up three separate causes in `integrations/openclaw/plugin.mjs`, all of which fail silently.

### 1. The agent identity is never read

OpenClaw invokes plugin hooks with **two** arguments:

```ts
agent_end: (event: PluginHookAgentEndEvent, ctx: PluginHookAgentContext) => ...
```

`PluginHookAgentContext` is where `agentId` lives, along with `sessionKey`, `sessionId` and `workspaceDir` (see `openclaw/src/plugins/hook-types.ts`). Both handlers here were declared as `(event) => ...`, so `ctx` was discarded — the event object itself carries only `{prompt, runId}` / `{messages, success, durationMs, runId}`. With no id available every turn collapsed onto the same scope.

### 2. `observe` was rejected with HTTP 400

`/agentmemory/observe` requires `project` and `cwd` as strings; the plugin sent neither. The server answered `400 hookType, sessionId, project, cwd, and timestamp are required strings`, and since `fallback_on_error` defaults to `true`, `postJson` swallowed it and returned `null`. Nothing was ever captured, and nothing was ever logged.

### 3. Replies delivered as a tool call were treated as empty

Agents running on a Codex app-server (`gpt-5.6-*`) do not emit an assistant text block — they answer by calling a message-sending tool, so the turn looks like:

```jsonc
{ "role": "assistant", "content": [
  { "type": "toolCall", "name": "message",
    "arguments": { "action": "send", "message": "<the actual reply>" } } ] }
```

`lastAssistantText()` only reads `{type:"text"}` blocks, so it returned `""` and `agent_end` hit `if (!userText || !assistantText) return;` — skipping capture entirely.

## Changes

- `resolveAgentId(ctx, event)`: prefer `ctx.agentId`, fall back to parsing `ctx.sessionKey` (`agent:<id>:...`), then `"main"`.
- Both hook handlers now accept `(event, ctx)`.
- `smart-search` sends `agentId` (and `project`) so recall is scoped to the calling agent.
- Before `observe`, the session is tagged via `/agentmemory/session/start` with `agentId` — `observe` does not accept an `agentId` itself, it inherits one from an existing session.
- `observe` now sends the required `project` and `cwd` (using `ctx.workspaceDir` when available).
- `lastAssistantText()` falls back to the last `toolCall` block's `arguments.message` / `.text` / `.body` / `.content`.

## Verification

Environment: OpenClaw `2026.7.1`, agentmemory `0.9.28`, two agents (`main`, `prompt-engineer`).

```
openclaw agent --agent prompt-engineer -m "Remember this codeword: OKAPI99..."
openclaw agent --agent main            -m "Remember this codeword: NARWHAL88..."
```

Before this change all sessions were stored as `agentId: "main"` (the second agent never appeared in the store at all), and before fixes 2 and 3 no observation was written under any id.

After:

```
sessions by agentId: { main: 20, prompt-engineer: 1 }

codeword     as prompt-engineer   as main
OKAPI99      1                    0
NARWHAL88    0                    1
```

Each agent retrieves only its own memories, and neither can see the other's.

## Tests

`test/openclaw-plugin.test.ts` gains four cases covering: scoping recall by `ctx.agentId`; falling back to `sessionKey` parsing; sending `agentId`/`project`/`cwd` on session-start and observe; and capturing a reply delivered as a tool call. Existing tests are unchanged and still pass (7/7).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory separation so each agent’s searches and conversations are stored independently.
  * Ensured assistant replies delivered through tool calls are captured correctly.
  * Improved session tracking with workspace and session details for more reliable conversation history.

* **Tests**
  * Added coverage for agent-specific memory scoping, session tracking, workspace details, and tool-call responses.

### DIFF
--- a/integrations/openclaw/plugin.mjs
+++ b/integrations/openclaw/plugin.mjs
@@ -46,8 +46,41 @@ function lastAssistantText(messages) {
     if (message.role !== "assistant") continue;
     const text = extractText(message.content);
     if (text) return text;
+    // Agents running on a Codex app-server (e.g. gpt-5.6-*) do not emit a plain
+    // assistant text block: they answer by CALLING a message-sending tool, so
+    // the user-visible reply lives in that tool call's arguments.
+    if (!Array.isArray(message.content)) continue;
+    for (const block of [...message.content].reverse()) {
+      if (!block || typeof block !== "object") continue;
+      if (block.type !== "toolCall") continue;
+      const args = block.arguments || block.input || {};
+      const candidate = args.message || args.text || args.body || args.content;
+      if (typeof candidate === "string" && candidate.trim()) return candidate.trim();
+    }
   }
   return "";
+}
+
+/**
+ * Resolve the id of the agent this turn belongs to.
+ *
+ * OpenClaw invokes plugin hooks as (event, ctx) where ctx is a
+ * PluginHookAgentContext carrying agentId/sessionKey/sessionId/workspaceDir
+ * (see openclaw src/plugins/hook-types.ts). Without this, every agent shares a
+ * single memory pool.
+ */
+function resolveAgentId(ctx, event) {
+  const context = ctx || {};
+  if (context.agentId) return String(context.agentId);
+  const sessionKey =
+    context.sessionKey ||
+    context.sessionId ||
+    (event && (event.sessionKey || event.sessionId)) ||
+    "";
+  const match = /^agent:([^:]+):/.exec(String(sessionKey));
+  if (match) return match[1];
+  if (event && event.agentId) return String(event.agentId);
+  return "main";
 }
 
 function latestUserText(messages) {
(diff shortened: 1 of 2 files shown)


jesusweb3:
Two implementation details worth calling out explicitly, since one of them goes slightly beyond the three bugs described above.

**1. `sessionId` resolution now prefers the hook context**

```js
const sessionId =
  ctx?.sessionId ||
  ctx?.sessionKey ||
  event.sessionId ||
  event.sessionKey ||
  event.runId ||
  `openclaw-${Date.now()}`;
```

In practice the old code always landed on `event.runId`, because the `agent_end` event carries only `{messages, success, durationMs, runId}` — so every single turn became its own agentmemory session. With `ctx.sessionId` / `ctx.sessionKey` available, consecutive turns of the same conversation now group into one session, which lines up with how the rest of agentmemory treats sessions (`observationCount` per session, `session/start`, `/observations?sessionId=...`).

I think this is the more correct behaviour, but it does change how turns are grouped going forward, so if you would rather keep this PR strictly to the three fixes I am happy to drop it and leave `event.runId` first.

**2. `cwd` fallback**

Where `ctx.workspaceDir` is not provided I fall back to the agent id, purely to satisfy the server's "must be a non-empty string" requirement on `/agentmemory/observe`. It is not semantically a path, so if you would prefer something else there (the process cwd, or relaxing the server-side check) that is an easy change.

Happy to adjust either one.

PR #1125, opened 2026-07-29
feat(opencode): allow AGENTMEMORY_PROJECT_NAME to override project name

aqeelat:
## What it does

The OpenCode capture plugin now consults `AGENTMEMORY_PROJECT_NAME` before falling back to its existing resolution (`ctx.worktree || ctx.project?.id || process.cwd()`). When the env var is set to a non-empty value, that value is used as the `project` on every observe/session-start call; otherwise behavior is unchanged.

## Why

1. **No override existed.** The OpenCode plugin resolved the project from the OpenCode context only — a user couldn't pin a stable project name (across machines, monorepo subdirs, or containerized setups where `cwd` differs).

2. **Cross-client consistency.** The Claude hooks already resolve the project via `AGENTMEMORY_PROJECT_NAME` → git toplevel basename → cwd basename (`src/hooks/_project.ts`). The OpenCode plugin was the only client that ignored this env var and sent a **full filesystem path** as the project. Since the daemon filters memory strictly by the `project` string (`s.project === projectName`), an OpenCode session (`/Users/.../myrepo`) and a Claude session (`myrepo`) on the same repo never shared memory. Setting `AGENTMEMORY_PROJECT_NAME=myrepo` on both now aligns them.

Same env var name the hooks use — one knob across both plugin ecosystems, no new variable to learn.

## How to verify

- `npm test` — 1432 passing, 1 skipped.
- Structural assertion in `test/opencode-auto-context.test.ts` confirms the plugin source checks `AGENTMEMORY_PROJECT_NAME` and that it precedes the `ctx.worktree` fallback.
- Manual: set `AGENTMEMORY_PROJECT_NAME=test-proj`, run a session, confirm `/session/start` and `/observe` calls carry `project: "test-proj"`. Unset it and confirm the previous behavior (ctx/cwd path) resumes.

## Scope

One-line behavior change in `plugin/opencode/agentmemory-capture.ts`. No daemon changes — the server already stores/filters by whatever `project` string it receives. Reuses the existing `AGENTMEMORY_PROJECT_NAME` the hooks already document; no new env var introduced.

## Summary by CodeRabbit

* **New Features**
  * Added support for overriding the project name using the `AGENTMEMORY_PROJECT_NAME` environment variable, including trimming surrounding whitespace.
  * Preserved existing project-name detection as the fallback when the variable is unset or blank.

* **Tests**
  * Added coverage for environment-based project name selection and fallback precedence across available project context.

### DIFF
--- a/plugin/opencode/agentmemory-capture.ts
+++ b/plugin/opencode/agentmemory-capture.ts
@@ -168,7 +168,8 @@ function extractErrorMessage(err: unknown): string {
 }
 
 export const AgentmemoryCapturePlugin: Plugin = async (ctx) => {
-  projectPath = ctx.worktree || ctx.project?.id || process.cwd();
+  const explicitProject = process.env.AGENTMEMORY_PROJECT_NAME?.trim();
+  projectPath = explicitProject || ctx.worktree || ctx.project?.id || process.cwd();
 
   return {
     event: async ({ event }) => {
--- a/test/opencode-auto-context.test.ts
+++ b/test/opencode-auto-context.test.ts
@@ -35,3 +35,17 @@ describe("OpenCode plugin auto-context injection (#431)", () => {
     expect(deletedBlock).toMatch(/startContextCache\.delete\(sid\)/);
   });
 });
+
+describe("OpenCode plugin project name override", () => {
+  const plugin = readFileSync(
+    "plugin/opencode/agentmemory-capture.ts",
+    "utf-8",
+  );
+
+  it("honours AGENTMEMORY_PROJECT_NAME before falling back to ctx/cwd", () => {
+    expect(plugin).toMatch(/AGENTMEMORY_PROJECT_NAME/);
+    expect(plugin).toMatch(
+      /explicitProject\s*\|\|\s*ctx\.worktree\s*\|\|\s*ctx\.project\?\.id\s*\|\|\s*process\.cwd\(\)/,
+    );
+  });
+});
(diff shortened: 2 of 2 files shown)


rohitg00:
Verified against main and this is correct: the hooks resolve project via `AGENTMEMORY_PROJECT_NAME` in `src/hooks/_project.ts`, and the OpenCode plugin never reads it, so the override silently does nothing there. The runtime tests asserting the actual `/session/start` payload are the right shape too. This is good to go as-is.

While confirming it I swept every other capture surface, and the same problem exists in four more places. Since you clearly get this bug class, would you want to take the rest as a follow-up PR? @aqeelat 

- `integrations/hermes/__init__.py:191` sends raw `cwd` as both `project` and `cwd` (this is open issue #1135)
- `integrations/pi/index.ts:123` and `:235` send raw `process.cwd()` as `project`
- `integrations/filesystem-watcher/watcher.mjs:322` reads `AGENTMEMORY_PROJECT`, a divergent env name; it should accept `AGENTMEMORY_PROJECT_NAME` with the old name kept as an alias
- `src/replay/jsonl-parser.ts:27` `deriveProject` takes the last path segment instead of the git-toplevel basename, and its `split("/")` means a Windows cwd becomes the whole raw path

The target semantics are exactly what `resolveProject()` in `src/hooks/_project.ts` does: env override, then git toplevel basename, then cwd basename. OpenCode itself is excluded from that list because #1122 covers its canonicalization; if you pick this up it would be worth coordinating there on the env name so we do not end up with three variants.

PR #1096, opened 2026-07-21
chore(deps): migrate @xenova/transformers to @huggingface/transformers v4

aqeelat:
## Summary

`@xenova/transformers@2.17.2` is deprecated and silently broken on Node 22+ (see #479). This swaps to `@huggingface/transformers@^4.2.0` — the same project, renamed, with an actively maintained ONNX runtime backend and Node 22/24 support. Drop-in for our usage: `pipeline`, `RawImage.fromBlob`, `tolist`, and `text-classification` output shape are all unchanged.

Closes #1095. Fixes #479.

## Changes

- **`package.json`**: replace `@xenova/transformers` + direct `onnxruntime-{node,web}` optional deps with a single `@huggingface/transformers@^4.2.0`. The onnxruntime packages are now bundled as HF v4's hard deps; `sharp`, `@huggingface/jinja`, and `@huggingface/tokenizers` come along transitively.
- **`src/xenova.d.ts` (deleted)**: removed entirely — `@huggingface/transformers@4.2.0` ships its own TypeScript declarations (incl. `RawImage`), so the ambient shim is no longer needed.
- **`src/providers/embedding/{local,clip}.ts`, `src/state/reranker.ts`**: swap import specifier + install error strings. Drop the `as unknown as TransformersModule` cast in `clip.ts`. Import errors are now only rewritten to the install hint when `err.code === "ERR_MODULE_NOT_FOUND"`; all other failures propagate unchanged instead of being masked by a blanket catch. The CLIP hint is now embedding-agnostic (the loader serves both text and image paths).
- **`tsdown.config.ts`**: update `neverBundle` list.
- **`__mocks__/@huggingface/transformers.ts` (new)**: manual mock fixture that throws a Node-shaped `ERR_MODULE_NOT_FOUND` at module top-level. Vitest's `vi.mock` factory throws get wrapped (original lands on `.cause`), so a factory can't simulate a direct missing-module rejection. The manual mock bypasses the factory wrapper and lets the test exercise the real `err.code` path — production code checks `err.code` only, with no test-framework coupling.
- **`test/{reranker,clip-embedding-provider,local-embedding-provider}.test.ts`**: use the manual mock for the "package unavailable" case; `afterEach` restores module state in the `doMock`-based describe blocks so mocks can't leak between tests.
[...]

PR #1126, opened 2026-07-29
fix(fs-watcher): validate roots before watch() for Node 24/26 compat

aqeelat:
## What

`FilesystemWatcher.start()` now validates each root with `statSync` + `isDirectory()` before calling `fs.watch()`. Nonexistent or non-directory roots are recorded as failures synchronously, so the "could not watch any of the configured roots" guard fires deterministically.

## Why

`fs.watch(path, { recursive: true })` on Linux does not throw synchronously for a nonexistent root on Node 24/26. Recursive watch initialization on Linux is asynchronous ([nodejs/node#45098](https://github.com/nodejs/node/pull/45098), Node 20), and [nodejs/node#63686](https://github.com/nodejs/node/pull/63686) (2026-06-03) changed the directory scan to suppress `ENOENT` — treating it as a deletion race rather than emitting an `'error'` event. The net effect: a nonexistent root now produces neither a synchronous throw nor an async error signal, so the watcher handle is always pushed and the `watchers.length === 0` guard never fired.

This surfaced as a CI failure (`test/fs-watcher.test.ts > throws if no watched roots could be attached`) on the `[ubuntu-latest, 24]` and `[ubuntu-latest, 26]` matrix cells added in #1096, while 20/22 and all macOS cells passed (no #63686 backport there). It's pre-existing — unrelated to any other change on `main`.

## How to verify

- `npm test` — 1431 passing, 1 skipped; `fs-watcher.test.ts` 15/15 green (was 14/15 on ubuntu Node 24/26).
- The "throws if no watched roots could be attached" test now passes deterministically because `statSync("/definitely/does/not/exist/xyz123")` throws `ENOENT` on every platform before `watch()` is reached.

## Scope

Two lines in `integrations/filesystem-watcher/watcher.mjs` — the `statSync`/`isDirectory` check sits at the top of the existing `try` block, reusing the already-imported `statSync`. No behavior change for valid roots; no new dependencies.

### DIFF
--- a/integrations/filesystem-watcher/watcher.mjs
+++ b/integrations/filesystem-watcher/watcher.mjs
@@ -248,6 +248,8 @@ export class FilesystemWatcher {
     const failures = [];
     for (const root of this.roots) {
       try {
+        const st = statSync(root);
+        if (!st.isDirectory()) throw new Error("not a directory");
         const handle = watch(
           root,
           { recursive: true, persistent: true },


rohitg00:
hey @aqeelat thanks for this, but fixed with better way.

PR #1130, opened 2026-07-31
feat(cli): native hooks adapter for Droid (Factory.ai)

berthojoris:
## Summary

Droid CLI (Factory.ai) currently only gets wired via MCP (`agentmemory connect droid`), even though Droid ships its own native, first-party hooks system — a real gap compared to Claude Code and Codex, which both get auto-capture hooks alongside MCP.

This PR adds that missing coverage: `agentmemory connect droid --with-hooks` now also installs Droid's native `~/.factory/hooks.json`, giving it the same zero-effort auto-capture that Claude Code and Codex already have.

## What changed

- **`plugin/hooks/hooks.droid.json`** (new) — bundled hook manifest for the 5 lifecycle events Droid documents (`SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `SessionEnd`; per [docs.factory.ai/cli/configuration/hooks-guide](https://docs.factory.ai/cli/configuration/hooks-guide)). Reuses the same bundled `plugin/scripts/*.mjs` hook scripts Claude Code and Codex already call — no new hook logic, just new wiring.
- **`src/cli/connect/droid.ts`** — adds `installDroidHooks()`, which merges the bundled manifest into `~/.factory/hooks.json` using the existing `buildMergedHooks()` engine from `codex-hooks.ts` (same shape: `{ hooks: { <event>: [{ matcher?, hooks: [{ type, command }] }] } }`). Idempotent re-install: agentmemory's own entries are identified and replaced by their `<pluginRoot>/scripts/` path prefix, so user-authored hook entries in the same file are never touched.
- **`src/cli/connect/json-mcp-adapter.ts`** — `createJsonMcpAdapter()` gained an optional `installHooks` callback, invoked whenever `--with-hooks` is passed (on both the fresh-install and already-wired paths, mirroring the Claude Code / Codex adapters). This lets any future JSON-MCP-shaped adapter opt into the same native-hooks pattern without a bespoke `install()` rewrite.
- **`src/cli/connect/codex-hooks.ts`, `types.ts`** — doc comments updated; `buildMergedHooks`/`--with-hooks` now back three hosts (Codex, Claude Code, Droid), not just Codex, and the code was already host-agnostic (manifest filename is a parameter) so no logic changes were needed there.
- **`README.md`**, **`plugin/skills/agentmemory-agents/REFERENCE.md`**, **`CHANGELOG.md`** — docs updated to describe the new `--with-hooks` path for Droid. The agents skill reference is auto-generated (`npm run skills:gen`) from `src/cli/connect/index.ts`'s `protocolNote`, so it now reflects the new capability automatically.

## Why this approach

Droid's native hooks config uses the *exact* same JSON shape (`matcher` + `hooks: [{ type, command }]`) as Codex's `~/.codex/hooks.json` workaround, so this reuses `buildMergedHooks()` as-is rather than writing a third merge implementation — only the bundled manifest file passed in differs per host.

## Test plan

- [x] `npx tsc --noEmit` — clean on every file touched by this PR (pre-existing unrelated errors elsewhere in the repo are untouched by this change).
- [x] New unit tests in `test/droid-connect-hooks.test.ts`: manifest merge produces the correct 5 events (and explicitly asserts Claude-only events like `PreCompact`/`Stop` do **not** leak in), preserves the `PreToolUse` matcher, appends to/doesn't clobber user-authored hooks, and re-install is idempotent.
- [x] New end-to-end tests in `test/connect-new-agents.test.ts` (Droid describe block): `--with-hooks` writes `~/.factory/hooks.json` with all 5 events; omitting the flag does **not** touch `hooks.json`; re-running `--with-hooks` on an already-wired MCP config still refreshes the hooks file.
- [x] Full `npm test` run — all Droid/connect-specific tests pass; the handful of pre-existing failures elsewhere in the suite were confirmed to reproduce identically on unmodified `main` (unrelated Windows-path/env flakiness, not caused by this change).
- [x] Bonus fix bundled in: `test/connect-new-agents.test.ts` had a pre-existing Windows test-isolation bug — `os.homedir()` reads `USERPROFILE` on win32, not `HOME`, so the "isolated temp home" trick silently no-op'd and adapter tests were writing into the real developer's home directory (discovered the hard way while validating this PR — confirmed via git-stash bisection that it's not introduced by this change, just uncovered by it). Fixed by also setting/restoring `USERPROFILE` alongside `HOME` in that file's `beforeEach`/`afterEach`. The same pattern exists in 6 other test files (`slots-flag-gate.test.ts`, `preferences.test.ts`, `env-loader.test.ts`, `consolidation-default.test.ts`, `cli-onboarding.test.ts`, `cli-connect.test.ts`) but those are left untouched here to keep this PR scoped — flagging for a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added optional native auto-capture hooks for Droid via `agentmemory connect droid --with-hooks`.
  * Supports session, prompt, and tool lifecycle events while preserving existing hook settings.
  * Hook installation safely handles dry runs, backups, refreshes, and repeat installations without duplicates.

* **Documentation**
  * Updated Droid integration guidance to explain enabling and refreshing native auto-capture hooks.


rohitg00 (changes requested):
Checked this against Factory's hooks reference and it holds up: paths, events, and stdin payload shape all match, and reusing the bundled scripts plus buildMergedHooks instead of hand-rolling new hook logic is exactly what I want these adapters to do. Tests pass in a clean worktree and the win32 USERPROFILE isolation fix is a nice catch.

One real bug before merge: the PreToolUse matcher in hooks.droid.json was carried over from the Claude manifest as `Edit|Write|Read|Glob|Grep`, but Droid has no `Write` tool. File creation there is `Create`, so memory enrichment never fires on file creation. Change the matcher to `Edit|Create|Read|Glob|Grep` (pre-tool-use.mjs already handles `create`, only the matcher blocks it).

Two non-blocking notes: the test comment saying `PreCompact/Stop` are Claude-only is wrong, Droid supports `Stop/SubagentStop/PreCompact/Notification` too, so wiring stop.mjs is a natural follow-up; and the droid comment in guidelines.ts about no auto-capture hooks goes stale once this lands, fine to update here or after.

@berthojoris

berthojoris:
@rohitg00 Thanks for the review — fixed the blocking matcher bug.

**Done**
- Updated `plugin/hooks/hooks.droid.json` PreToolUse matcher from `Edit|Write|Read|Glob|Grep` → `Edit|Create|Read|Glob|Grep`
- Updated the matcher assertion in `test/droid-connect-hooks.test.ts`
- Confirmed `pre-tool-use.mjs` already handles `create`; only the matcher was blocking it
- Relevant tests pass (`droid-connect-hooks` + `connect-new-agents`)

**On the non-blocking notes**
- Leaving `Stop` / `SubagentStop` / `PreCompact` / `Notification` wiring (e.g. `stop.mjs`) as a follow-up, as you suggested
- I couldn't find a stale `guidelines.ts` droid "no auto-capture hooks" comment on this branch — happy to update it in a follow-up if you point me at the path, or if it's already been moved/renamed

Ready for another look when you have a moment.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## [Unreleased]
 
+### Added
```
rohitg00: Remove changelog changes @berthojoris

README.md
```diff
@@ -677,7 +677,7 @@ The agentmemory entry is the **same MCP server block** across every host that us
...
 | **Zed** | `~/.config/zed/settings.json` | `agentmemory connect zed` writes under `context_servers` (Zed's key, NOT `mcpServers`). Remote MCP servers can be wired via `{"url": "..."}` instead. |
-| **Droid (Factory.ai)** | `~/.factory/mcp.json` | `agentmemory connect droid` writes the standard `mcpServers` block. Project-scoped overrides go in `<repo>/.factory/mcp.json`. The `/mcp` slash command inside droid lists configured servers. |
+| **Droid (Factory.ai)** | `~/.factory/mcp.json` + `~/.factory/hooks.json` | `agentmemory connect droid` writes the standard `mcpServers` block. Project-scoped overrides go in `<repo>/.factory/mcp.json`. The `/mcp` slash command inside droid lists configured servers. Run `agentmemory connect droid --with-hooks` to also wire Droid's native hooks (SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, SessionEnd) into `~/.factory/hooks.json` for zero-effort auto-capture, same merge/re-install behavior as the Codex and Claude Code hook installers. |
```
rohitg00: This looks too much for Readme @berthojoris

CHANGELOG.md
berthojoris: Removed — CHANGELOG Unreleased entry is gone.

README.md
berthojoris: Trimmed — restored the short Droid row and left only a brief `--with-hooks` mention.

rohitg00 (approved):
Thanks @berthojoris for your contribution, If you'd like to test - please test antigravity, vscode, copilot, and codex plugins, CLIs thoroughly with pulling from main and fix the changes if anything is broken and submit PR. Happy to review them.

berthojoris:
@rohitg00 Thanks — I'll pull latest main, thoroughly test the antigravity, vscode, copilot, and codex plugins/CLIs, and open a follow-up PR if anything is broken.

Issue #625, opened 2026-05-24
CLAUDE_MEMORY_BRIDGE writes to wrong path — never read by Claude Code


mihailmariusiondev:
## Bug

`loadClaudeBridgeConfig()` in `src/config.ts` generates a `memoryFilePath` that Claude Code never reads, due to two path mismatches:

### 1. Leading dash stripped

```ts
const safePath = projectPath.replace(/[/\\]/g, "-").replace(/^-/, "");
```

The `.replace(/^-/, "")` strips the leading `-` from the slug. But Claude Code **keeps** it. Example for `/home/user/repos/my-project`:

- Claude Code reads from: `~/.claude/projects/-home-user-repos-my-project/MEMORY.md`
- agentmemory writes to: `~/.claude/projects/home-user-repos-my-project/memory/MEMORY.md`

### 2. Extra `/memory/` subdirectory

```ts
memoryFilePath = join(homedir(), ".claude", "projects", safePath, "memory", "MEMORY.md");
```

Claude Code stores `MEMORY.md` directly in the project slug directory, not in a `memory/` subdirectory.

### Impact

The bridge is completely broken on Linux/macOS (any absolute path starting with `/`). The written `MEMORY.md` is never read by Claude Code. This has been the case since the bridge was introduced in v0.4.0.

### Fix

```diff
- const safePath = projectPath.replace(/[/\\]/g, "-").replace(/^-/, "");
- memoryFilePath = join(homedir(), ".claude", "projects", safePath, "memory", "MEMORY.md");
+ const safePath = projectPath.replace(/[/\\]/g, "-");
+ memoryFilePath = join(homedir(), ".claude", "projects", safePath, "MEMORY.md");
```

### Environment

- agentmemory v0.9.21
- Claude Code (latest)
- Ubuntu 24.04 (WSL2)
- Node.js v24.15.0

rohitg00:
Fixed in v0.9.22 — https://github.com/rohitg00/agentmemory/releases/tag/v0.9.22. Thanks for the report.

mihailmariusiondev:
Thanks!

PR #1134, opened 2026-08-01
fix(claude-bridge): restore memory/ subdir in MEMORY.md path

txtsamu:
## Summary

`CLAUDE_MEMORY_BRIDGE` writes `MEMORY.md` to a path Claude Code never reads, so the bridge is silently broken on every platform.

### The bug

`loadClaudeBridgeConfig()` in `src/config.ts` computes:

```
~/.claude/projects/<slug>/MEMORY.md
```

But Claude Code 2.x stores **project memory** at:

```
~/.claude/projects/<slug>/memory/MEMORY.md
```

The `memory/` subdirectory holds `MEMORY.md` (the index) plus one per-topic `.md` file per memory. Issue #625 fixed the leading-dash slug problem (`.replace(/^-/, "")` removal) but in the same change also dropped the `memory/` segment, pointing the bridge at a file Claude Code never reads — the written index simply disappears.

### Evidence

- Verified against the **Claude Code 2.1.141 bundle**: path constants `memory` + `MEMORY.md` are joined under `~/.claude/projects/<slug>/`, and the loader filters `.md` files with `basename !== "MEMORY.md"` inside that dir (i.e. index + topic files live together in `memory/`).
- Verified on disk: a real project memory dir contains `memory/MEMORY.md` (a `# Memory Index` with links) alongside `memory/<topic>.md` files.
- Also confirmed the **latest npm release v0.9.28 still ships the bug** — the dist still joins `"projects", safePath, "MEMORY.md"`.

### The fix

Restore the `memory` segment in the join, keeping the #625 leading-dash fix:

```ts
memoryFilePath = join(
  homedir(),
  ".claude",
  "projects",
  safePath,
  "memory",
  "MEMORY.md",
);
```

Plus updated path tests (`test/claude-bridge-path.test.ts`) to assert the `memory/` subdir.

## Test plan

- `npx vitest run test/claude-bridge-path.test.ts test/claude-bridge.test.ts` → **10 passed, 0 failed**
- Full suite `npx vitest run` → **1449 tests: 1448 passed, 0 failed, 1 pending** (442/442 suites)

## Notes (out of scope)

- `direction=read` on `memory_claude_bridge_sync` currently only reads + returns the file; the tool description says "import from MEMORY.md" but nothing is imported into memories. Worth a follow-up.
- `serializeToMemoryMd` writes a `# Agent Memory (auto-synced by agentmemory)` layout rather than Claude Code's native `# Memory Index` bullet format, so after a `write` the file is no longer in Claude's native shape. Follow-up candidate.

## Summary by CodeRabbit

- **Bug Fixes**
  - Claude bridge memory files are now stored in the standardized `memory/MEMORY.md` location within each project directory.
  - Project paths with leading hyphens are preserved correctly across supported operating systems.
  - Improved path handling helps ensure existing Claude bridge memory is located consistently.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -252,19 +252,20 @@ export function loadClaudeBridgeConfig(): ClaudeBridgeConfig {
   const lineBudget = safeParseInt(env["CLAUDE_MEMORY_LINE_BUDGET"], 200);
   let memoryFilePath = "";
   if (enabled && projectPath) {
-    // Claude Code stores MEMORY.md at
-    //   ~/.claude/projects/<slug>/MEMORY.md
+    // Claude Code stores project memory at
+    //   ~/.claude/projects/<slug>/memory/MEMORY.md
     // where <slug> is the project path with `/` and `\` swapped for `-`.
     // The leading `-` from an absolute POSIX path is preserved (Claude
     // Code keeps it; stripping it produced a slug Claude never reads).
-    // There's also no `memory/` subdirectory — the file sits directly
-    // under the slug dir.
+    // The `memory/` subdirectory holds MEMORY.md (the index) plus one
+    // per-topic `.md` file per memory (verified against Claude Code 2.x).
     const safePath = projectPath.replace(/[/\\]/g, "-");
     memoryFilePath = join(
       homedir(),
       ".claude",
       "projects",
       safePath,
+      "memory",
       "MEMORY.md",
     );
   }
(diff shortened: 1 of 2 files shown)


txtsamu:
Local verification (before CI is approved):

- `npx vitest run test/claude-bridge-path.test.ts test/claude-bridge.test.ts` → 10 passed, 0 failed
- Full suite `npx vitest run` → **442/442 suites, 1448 passed / 0 failed / 1 pending**

CI run is gated on first-time-contributor approval (`action_required`); nothing failed. Also confirmed the bug ships in the latest npm release v0.9.28.

Issue #1142, opened 2026-08-03
GET /agentmemory/export drops the iii worker once the response passes 16 MiB, 404ing every endpoint


dmazhukov:
Once a store grows past a certain size, `GET /agentmemory/export` stops returning and takes
the whole daemon down with it for about a second. The failure is a hard size limit on the
response, not a timeout, and it is fully deterministic.

## What I see

```
$ curl -s -o /dev/null -w '%{http_code}\n' localhost:3111/agentmemory/export
500
$ curl -s localhost:3111/agentmemory/export
{"error":"Invocation stopped","error_id":"…"}
$ curl -s localhost:3111/agentmemory/export
{"error":"Function api::export not found","error_id":"…"}
```

The daemon log shows the export finishing before the connection dies:

```
[agentmemory] info Export complete {"sessions":15,"observations":12750,"memories":3000,…}
[iii] Reconnecting in 955ms (attempt 1)...
[iii] Worker registered with ID: 43829794-b900-4040-8351-58cf42d5ec61
```

`Export complete` is logged, so the data was assembled and the function returned. What dies is
the response on its way back. While the worker re-registers, every endpoint 404s, so a single
GET makes the entire REST surface unavailable for ~1s.

## Expected

Either the export returns, or it fails on its own without disconnecting the worker and taking
128 unrelated endpoints with it.

## The limit is 16 MiB, and it is a limit rather than a timeout

Bisected on a local instance by growing the store in ~27 KB steps and re-requesting a fixed
slice:

| serialized response | result |
|-|-|
| 16 771 046 B (15.9941 MiB) | 200 OK in 365 ms |
| next step, +27 KB | 500 in 103 ms |

16 MiB is 16 777 216 B, so the last success lands 6 170 B short of it — 0.037%. Note the
failing request returns **faster** than the succeeding one (103 ms vs 365 ms). Nothing is
timing out; the frame is refused.

## Where the limit appears to live

I could not find it in this package. `iii-sdk` opens its WebSocket with no `maxPayload`, so
the `ws` client default of 100 MiB applies. The `iii` engine binary carries tungstenite's
`WebSocketConfig` field names (`max_write_buffer_size`, `max_fragment_size`), and
tungstenite's default `max_frame_size` is exactly 16 MiB, which matches the measurement.

**This is an inference from strings in the binary, not something I read in engine source**,
and I could find no config key or CLI flag that exposes it. If the engine can be configured
to raise it, that would be a second, independent fix.

[...]

Issue #890, opened 2026-06-10
mesh/export has no pagination — backlogs past the worker WS message limit fail forever with "Invocation stopped", making mesh sync unrecoverable after an outage


egantz:
## Summary

`/agentmemory/mesh/export` builds the entire delta in one response with no pagination or size cap. Once the since-window backlog grows past the iii engine↔worker WebSocket message limit, the export dies with `{"error":"Invocation stopped"}` — and because mesh sync only advances its `since` watermark on success, **the sync can never catch up again**. A mesh outage longer than ~a day becomes permanent until someone intervenes manually.

This is the mesh-flavored instance of the known large-payload "Invocation stopped" class (#814, #753, #309), but it's worse here because the failure is self-reinforcing: the longer sync is down, the bigger the delta, the more certain the next attempt fails.

Observed on v0.9.27; handler unchanged in `main` (`src/triggers/api.ts`, `api::mesh-export`, ~L2677).

## Real-world failure (how I hit it)

Two-machine LWW mesh (laptop ↔ mac mini), sync every 120s via the `/mesh/export` → `/mesh/receive` pair. The peer was down for ~25h. On recovery:

- The backlog was 58 memories + ~9.7k graph nodes + ~15.7k graph edges ≈ **29MB** of JSON (graph extraction churns nodes/edges constantly, so `updatedAt > since` over a day catches a big slice of the graph).
- `GET /agentmemory/mesh/export?since=<25h ago>` → `500 {"error":"Invocation stopped"}` in ~0.5s, every time.
- Side effect worth documenting: each failed attempt **kills the worker's WS connection** — the engine log shows an endless `[iii] Reconnecting… Worker registered with ID: …` loop, and requests that land in the gap get an instant `404` (no handler registered). So the visible symptom alternates 500/404, which looks like flaky networking but is deterministic payload size.
- A tiny window (`since=<5 min ago>`) returns 200 instantly — the handler is fine, the response size is the problem.

## Root cause

`api::mesh-export` lists **entire collections** (`KV.memories`, `KV.actions`, `KV.semantic`, `KV.procedural`, `KV.relations`, `KV.graphNodes`, `KV.graphEdges`), filters by timestamp, and returns everything in a single body. There is no `until`/cursor/`limit` parameter and no size guard. The graph collections dominate: on my store they are 23MB (nodes) + 24MB (edges) on disk.

## Why it can't self-heal

[...]

PR #1143, opened 2026-08-03
feat(export): paginate the non-session collections the frame guard's hint calls out

dmazhukov:
## Summary

Follow-up to the frame guard that shipped in v0.9.29. Its hint says the non-session
collections "are not yet paginated" — this makes them paginated.

Originally this PR carried its own size guard for #1142. v0.9.29 landed `frame-guard.ts`
first, so that half is dropped and what remains builds on `checkPayloadFrameSize` rather than
competing with it.

## What this changes

**`?collectionLimit=` / `?collectionOffset=`** window `memories`, `summaries` and the sixteen
top-level collections, which previously came in full no matter what `?maxSessions=` said. On a
store I reproduced against, `?maxSessions=1` still returned 5.50 MB, of which ~4.9 MB was that
floor. It grows with the store, and once it alone crosses the cap the export returns nothing at
any parameter combination — the dead end #890 describes for `mesh/export`.

**`?collections=`** narrows the payload to the collections a caller actually reads, so an agent
after a lesson is not also pulling `graphEdges`. `sessions`, `observations` and `profiles` are
outside the vocabulary on purpose: they are windowed by `maxSessions`/`offset`, and profiles are
derived from the session page rather than listed. Unknown names are dropped rather than refused,
so a client can name a collection an older build lacks and still get the ones it has.

**`collectionPagination`** reports per-collection `totals` and a combined `hasMore`. `totals`
cover every collection even when deselected, since that is what clients read for corpus size;
`hasMore` looks only at the selected ones, so a client that asked for six of eighteen can stop
on the flag instead of comparing against `totals` by hand.

**The MCP tool** declared no properties and always triggered `mem::export` with an empty
payload, so none of this was reachable from MCP — which is where an agent actually calls
export. It now declares and forwards the same arguments.

The frame guard's hint is updated to name the new parameters instead of describing the gap.

## Behaviour

An unparameterised export returns exactly what it returns today. A test pins the full-corpus
response field by field so that cannot drift.

## What this doesn't fix

- **Observations stay atomic per session.** `collectionLimit` does not slice them; they follow
  the session window. A single session whose observations alone exceed the cap still cannot be
  exported.
- **`api::export` returns 200 for a refusal.** The guard's `OversizedPayload` travels as a
  successful HTTP response, so a client has to inspect the body to notice. Left alone here since
  it is the guard's contract, not this change's — happy to send a 413 mapping separately if you
  want one.
- **`mesh/export` (#890)** has the same shape and is untouched.

## Testing

- `test/export-import.test.ts`: full-corpus shape pinned; `collectionLimit` bounds every
  collection while the unparameterised call still returns everything; a three-page walk ends on
  `hasMore: false`; `collections` selects, drops unknown names, and treats an empty list as
  "none".
- `test/mcp-export-tool.test.ts`: arguments forwarded, unusable bounds dropped, empty allowlist
  kept distinct from an absent one.
- `npx vitest run --exclude test/integration.test.ts` → 1614 passed, 0 failed.
- `npx tsc --noEmit` → 25 errors, byte-identical to `main`.
- `npm run skills:check` and `npm run build` clean.

### DIFF
--- a/plugin/skills/agentmemory-mcp-tools/REFERENCE.md
+++ b/plugin/skills/agentmemory-mcp-tools/REFERENCE.md
@@ -18,7 +18,7 @@ agentmemory exposes 54 MCP tools. 8 are in the lean core set (`--tools core` or 
 | `memory_consolidate` | yes | `tier`: string | Run the 4-tier memory consolidation pipeline (working -> episodic -> semantic -> procedural). |
 | `memory_crystallize` |  | `actionIds`*: string, `project`: string, `sessionId`: string | Compress completed action chains into compact crystal digests using LLM summarization. Extracts narrative, key outcomes, files affected, and lessons. |
 | `memory_diagnose` | yes | `categories`: string | Run health checks across all subsystems (actions, leases, sentinels, sketches, signals, sessions, memories, mesh). Identifies stuck, orphaned, and inconsistent state. |
-| `memory_export` |  | none | Export all memory data as JSON. |
+| `memory_export` |  | `maxSessions`: number, `offset`: number, `collectionLimit`: number, `collectionOffset`: number, `collections`: string | Export memory data as JSON. Past the transport size limit the export is refused, so use the paging arguments on a large store. |
 | `memory_facet_query` |  | `matchAll`: string, `matchAny`: string, `targetType`: string | Query targets by facet tags with AND/OR logic. Find all actions tagged priority:urgent AND team:backend. |
 | `memory_facet_tag` |  | `targetId`*: string, `targetType`*: string, `dimension`*: string, `value`*: string | Attach a structured tag (dimension:value) to an action, memory, or observation for multi-dimensional categorization. |
 | `memory_file_history` |  | `files`*: string, `sessionId`: string | Get past observations about specific files. |
--- a/src/functions/export-import.ts
+++ b/src/functions/export-import.ts
@@ -58,20 +58,101 @@ async function runChunked<T>(
   }
 }
 
+// Every collection mem::export can put in the payload — the vocabulary
+// `?collections=` is matched against. sessions, observations and profiles
+// are absent on purpose: they are windowed by maxSessions/offset, and
+// profiles are derived from the session page rather than listed.
+const EXPORT_COLLECTION_NAMES: ReadonlySet<string> = new Set([
+  "memories",
+  "summaries",
+  "graphNodes",
+  "graphEdges",
+  "semanticMemories",
+  "proceduralMemories",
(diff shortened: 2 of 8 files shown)


dmazhukov:
Heads-up on the red CI here: both failures reproduce on a clean checkout of `main` and are not from this branch.

- `npm run skills:check` → `DRIFT: plugin/skills/agentmemory-rest-api/REFERENCE.md` (records 118, the generator produces 119)
- `test/consistency.test.ts` → README says `129 endpoints on port`, 130 are registered

An endpoint landed in #1132 / #1136 without the generated docs moving with it. I checked that this branch adds no endpoints: `npm run skills:gen` produces the same 119 on `main` and here.

#1144 refreshes the four files that carry the count. Once it lands I'll rebase this branch so the matrix reflects the actual change.

dmazhukov:
Good catch on `memory_export` — fixed in 712d9ba.

The empty `payload: {}` predates this PR, but the refusal I added tells the caller to page, and that advice was not actionable from the MCP tool because the schema declared no properties. `src/mcp/tools-registry.ts` now declares `maxSessions`, `offset`, `collectionLimit` and `collectionOffset`, and `src/mcp/server.ts` validates and forwards them with the same integer bounds the REST handler uses. `plugin/skills/agentmemory-mcp-tools/REFERENCE.md` is regenerated via `npm run skills:gen`.

`test/mcp-export-tool.test.ts` covers the four cases: arguments forwarded, unusable bounds dropped, a refusal reported as 413, and a normal export still 200.

dmazhukov:
Correction to my note above: I've folded the same 4-line refresh into this branch (3966cb8, identical to #1144's commit) rather than waiting.

`npm run skills:check` runs ahead of the test suite, so leaving it out meant the whole matrix went red at the first step and showed nothing about the change itself. With it in, this branch is green locally: `skills:check` passes and `npx vitest run --exclude test/integration.test.ts` gives 1560 passed, 0 failed.

#1144 still stands on its own if you'd rather take the housekeeping separately — it is the same commit, so whichever merges first makes the other a no-op.

dmazhukov:
Rebased onto v0.9.29 and reduced.

The size guard this PR originally carried is gone — `frame-guard.ts` landed first and does the same job, so keeping mine would have been two guards racing in one function. What is left builds on `checkPayloadFrameSize` instead: the collection pagination its hint asks for, plus `?collections=`, plus the MCP tool being able to reach either. The hint itself is updated to name the parameters rather than describe the gap.

Title and description rewritten to match. 1614 tests pass, `tsc` output is byte-identical to `main`.

PR #1146, opened 2026-08-03
feat(cli): native hooks adapter for Antigravity CLI (agy)

berthojoris:
## What

Adds `agentmemory connect antigravity-cli` — a new adapter for the Antigravity `agy` CLI, with optional native auto-capture hooks behind `--with-hooks`.

Antigravity ships two products that do **not** share configuration:

- the **IDE**, already wired by `connect antigravity` (app-support dir);
- the **`agy` CLI**, which reads its customizations out of `~/.gemini/` — MCP servers from `~/.gemini/config/mcp_config.json`, hooks from `~/.gemini/config/hooks.json` — and until now was not wired at all.

This PR covers the second one. Detection keys off `~/.gemini/antigravity-cli/`, which only the CLI creates (`~/.gemini/` alone would also match a Gemini CLI install).

## Why the Codex merge engine could not be reused

Unlike Droid (#1130), Antigravity's hooks contract differs from the Claude Code / Codex / Droid family in three ways:

1. **`hooks.json` is a map of *named* hook bundles at the root**, not the `{ hooks: { <Event>: [...] } }` envelope. So `src/cli/connect/antigravity-hooks.ts` implements a merge that owns top-level keys instead of per-event entries. User-authored bundles are preserved verbatim; a re-install replaces only the bundle whose commands point under the bundled plugin dir.
2. **Only five events exist** — `PreToolUse`, `PostToolUse`, `PreInvocation`, `PostInvocation`, `Stop`. There is no `SessionStart`, `SessionEnd` or `UserPromptSubmit`, so the session lifecycle is synthesized from the first `PreInvocation` and from `Stop`. `PostInvocation` is deliberately left unwired — `PostToolUse` already captures the work, and firing both would double-record every turn.
3. **The stdin payload is camelCase and nested** (`toolCall.args` with PascalCase keys such as `AbsolutePath` / `TargetFile` / `Query`, plus `conversationId` and `workspacePaths` instead of `session_id` / `cwd`), and **stdout must be a JSON object** — `pre-tool-use.mjs` writes raw prose when `AGENTMEMORY_INJECT_CONTEXT=true`, which Antigravity would fail to parse as a `PreToolUse` decision.

`src/hooks/antigravity-bridge.ts` (bundled to `plugin/scripts/antigravity-bridge.mjs`) bridges all three: it normalizes the payload onto the shape the existing hooks already accept, maps Cascade tool names (`view_file`, `replace_file_content`, …) onto the `read`/`edit`/`write`/`grep` vocabulary the capture heuristics use, pipes to the right script, discards child stdout, and always answers `{}` so Antigravity's own permission decisions are never overridden. The bridge never blocks or rewrites a tool call.

Event names, tool names and arg keys were verified against the shipped `agy` binary rather than docs alone (the docs disagree on the global hooks path). The customization dir is `~/.gemini/config/`, matching where `agy` already keeps `mcp_config.json` and `plugins/`.

## How to verify

```bash
npm install
npm run build          # compiles clean; plugin/scripts/*.mjs regenerate byte-identical
npx vitest run test/antigravity-connect-hooks.test.ts test/cli-connect.test.ts test/connect-guidelines.test.ts
```

`test/antigravity-connect-hooks.test.ts` (new, 15 cases) covers the merge engine — placeholder resolution, single-named-bundle shape, exact event set, `PreToolUse` matcher scope, preservation of user bundles, replacement of a stale agentmemory bundle, idempotent re-install — plus bridge payload normalization and event routing.

End-to-end against a real install:

```bash
agentmemory connect antigravity-cli --with-hooks --dry-run   # inspect the plan
agentmemory connect antigravity-cli --with-hooks             # backs up an existing hooks.json first
```

Re-running is idempotent; user-authored bundles in `~/.gemini/config/hooks.json` are left untouched.

## Notes for the reviewer

- `plugin/skills/agentmemory-rest-api/REFERENCE.md` moves `118` → `119` endpoints. That is `npm run skills:gen` correcting a count that was already stale on `main` — it is not a side effect of this feature. No REST endpoint is added here.
- The `PreToolUse` matcher lists 11 tools while `TOOL_NAME_MAP` maps 14 (`view_line_range`, `read_url_content`, `propose_code` are mapped but not matched). This is deliberate — `PostToolUse` uses an empty matcher and catches them anyway — but happy to widen the matcher if you would rather have the two lists identical.
- No `CHANGELOG.md` entry, and the README row is kept to one line, matching the review feedback on #1130.

## Summary by CodeRabbit

* **New Features**
  * Added Antigravity CLI integration with MCP configuration and optional native hook installation.
  * Added agent memory synchronization across session, tool-use, and stop events.

* **Bug Fixes**
  * Improved handling of event payloads, tool names, session identifiers, and existing hook configurations.
  * Added safer, repeatable hook installation while preserving existing settings.

* **Documentation**
  * Added Antigravity setup guidance, configuration details, and usage instructions.

* **Tests**
  * Added coverage for integration, hook merging, event routing, payload normalization, and repeated installations.


berthojoris:
The CI failure here is pre-existing on `main`, not from this branch.

`test/consistency.test.ts > documented REST endpoint counts match registered API paths` fails because `5023cf3` (#1132) registered a 130th `api_path:` in `src/triggers/api.ts` while `README.md`, `AGENTS.md` and `src/index.ts` still say `129`. The push CI run for `5023cf3` on `main` is red for the same reason, and every PR opened since inherits it.

Fix is in #1147 (three-line docs sync, no behaviour change). Once that lands I'll merge `main` back into this branch so CI here goes green.

Nothing in this PR touches `src/triggers/api.ts`; the only other test-visible change is `plugin/skills/agentmemory-rest-api/REFERENCE.md` moving 118 → 119, which is `npm run skills:gen` correcting a separately stale count.

src/cli/connect/antigravity-hooks.ts
```diff
@@ -0,0 +1,123 @@
...
+          ),
+          ...(handler.timeout !== undefined && { timeout: handler.timeout }),
+        })),
+      };
+      if (entry.matcher !== undefined) next.matcher = entry.matcher;
+      return next;
+    });
+  }
+  return out;
+}
```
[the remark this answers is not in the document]
berthojoris: Confirmed and fixed in bb74810.

Reproduced against the actual template:

```
BROKEN "C:/plug$$in"   -> "C:/plug$in/scripts/..."
BROKEN "C:/plug$&in"   -> "C:/plug${CLAUDE_PLUGIN_ROOT}in/scripts/..."
BROKEN "C:/plug$`in"   -> "C:/plugnode \"in/scripts/..."
BROKEN "C:/plug$'in"   -> "C:/plug/scripts/..."
ok     "C:/plug$1in"   -> unchanged
ok     "C:/plug$<x>in"  -> unchanged
```

`$1` and `$<name>` are not affected — the regex has no capture groups and no named groups, so both stay literal. The four that do break are now covered by a regression test that builds the manifest against a temp plugin root named `plug$&$$in`; it fails on the old code and passes on the new.

Worth flagging separately: the identical expansion exists at `src/cli/connect/codex-hooks.ts:97`, which is on `main` and feeds the Claude Code / Codex / Droid install paths. Out of scope here, happy to send it as its own PR.

rohitg00 (changes requested):
This is well-built. I verified the hook surface against the official docs (antigravity.google/docs/hooks) and the config path, the five events, and the camelCase/PascalCase payload keys all match exactly, so it is not guessing an API. The bridge reuses the bundled hook scripts through a thin normalizer rather than hand-rolling capture, which is the right call given agy's payload and named-bundle `hooks.json` genuinely differ from the Claude envelope. Supply chain is clean, guidelines.ts shares the same idempotent GEMINI.md block as the existing antigravity entry so there is no double-write, and the generated REFERENCE bumps are exactly what `skills:gen` produces for the new adapter.

The failing CI is not your fault. main currently has an endpoint-count drift (registers 130 routes while the docs say 129), so `consistency.test.ts` fails on every branch cut from main right now. Your PR adds no REST endpoint. That failure clears once the count fix lands on main; nothing to do here for it.

One real blocker before merge: the `PreToolUse` hook emits a bare `{}`. The docs mark `decision` as required for PreToolUse output, and there is empirical evidence (cmux #5358 on agy 1.0.5) that a `{}` response makes agy deny every tool call, which would break the agent instead of passively capturing. agy's own bundled vibe-island plugin registers no PreToolUse at all.

Asks:
- Either drop the `PreToolUse` registration (PostToolUse already captures file activity), or have it emit `{"decision":"allow"}`.
- Add a test asserting the PreToolUse stdout contract so this can't regress.
- If you can, confirm against a live `agy` which allow-payload it actually accepts.

Everything else is approvable.

berthojoris:
Thanks for the careful read, and for checking the hook surface against the docs yourself.

You're right about `PreToolUse`. I went with the explicit-allow option rather than dropping the registration: `PostToolUse` does cover file activity, but `PreToolUse` still gives the capture pipeline the tool intent before it runs, and `{"decision":"allow"}` is what the docs actually ask for. Pushed in `278fb05`.

The response now goes through a small `responseFor(event)`:

```ts
export function responseFor(event: string): string {
  return event === "PreToolUse" ? '{"decision":"allow"}' : "{}";
}
```

Every other event stays on `{}` — none of them carry a permission decision, and emitting `decision` or `terminationBehavior` there would override the user's own settings.

One thing I found while fixing it: the bridge wrote its response from the `finally` block, but a few paths could exit with stdout empty (unparseable payload, a throw during capture). For `PreToolUse` an empty response is as fatal as `{}`, so the write now always happens with the right contract regardless of how the capture went.

On the test — two of them:

- a unit test over `responseFor`, pinning `PreToolUse` → `{"decision":"allow"}` and every other event → `{}`;
- an end-to-end one that spawns the bundled `plugin/scripts/antigravity-bridge.mjs` with `AGENTMEMORY_URL=http://127.0.0.1:1`, so every capture fetch fails fast. That's deliberately the failed-capture case, since that's where a swallowed error would produce empty stdout.

I could **not** confirm against a live `agy` — it isn't installed on my machine, so the allow payload is what the docs specify plus the cmux#5358 evidence you cited, not something I observed the binary accept. Worth a second pair of eyes if you have `agy` handy; if it turns out to want a different shape, it's a one-line change in `responseFor` and the test moves with it.

Also picked up CodeRabbit's nitpick while I was in there: the `ARG_KEY_MAP` test now covers all seven mappings (`TargetFile`, `DirectoryPath`/`SearchDirectory`, `Pattern`/`Query` included) plus a case pinning that an explicit canonical key wins over a PascalCase alias. Its other comment — the `$`-pattern issue in `resolveBundle` — was already fixed in `bb74810`; it reviewed the commit before that one.

And thanks for tracking down the CI failure, that saves me a detour.

plugin/scripts/antigravity-bridge.mjs
```diff
@@ -0,0 +1,138 @@
...
+			const n = raw["invocationNum"];
+			return typeof n !== "number" || n <= 1 ? ["session-start.mjs", "prompt-submit.mjs"] : ["prompt-submit.mjs"];
+		}
+		case "PreToolUse": return ["pre-tool-use.mjs"];
+		case "PostToolUse": return ["post-tool-use.mjs"];
+		case "Stop": return ["stop.mjs", "session-end.mjs"];
+		default: return [];
+	}
+}
+/**
```
rohitg00: remove these comments

src/cli/connect/antigravity-cli.ts
```diff
@@ -0,0 +1,100 @@
...
+  displayName: "Antigravity CLI (agy)",
+  detectDir: ANTIGRAVITY_CLI_DIR,
+  configPath: join(CUSTOMIZATION_DIR, "mcp_config.json"),
+  docs: "https://github.com/rohitg00/agentmemory#other-agents",
+  protocolNote:
+    "→ Using MCP via ~/.gemini/config/mcp_config.json (the agy CLI, not the Antigravity IDE — that one is `connect antigravity`). The `/mcp` slash command inside agy lists configured servers. Pass --with-hooks to also install the native ~/.gemini/config/hooks.json auto-capture hooks.",
+  installHooks: installAntigravityCliHooks,
+});
+
+/**
```
rohitg00, line 43: lot of similar comments

rohitg00 (approved):

berthojoris:
Correction to my last comment: I did get access to a live `agy` (1.0.15) and ran the verification you asked for. You were right about `PreToolUse`, and the probe turned up two further defects that would have stopped this adapter from capturing anything at all. Fixed in `6e481cb`.

Method: a probe hook that logs the payload agy sends and replies with whatever I put in a file, driven by `agy --print` runs, cross-checked against the `exa.hooks_pb` descriptor embedded in the binary.

**1. `PreToolUse` — confirmed, on 1.0.15 as well as 1.0.5**

Three runs, same prompt, only the hook's stdout differing:

| stdout | result in transcript |
|---|---|
| `{}` | `tool call denied with reason: ` (empty reason) |
| `{"decision":"deny","reason":"probe: deny path check"}` | `tool call denied with reason: probe: deny path check` |
| `{"decision":"allow"}` | tool ran, 0 denials |

So it is not specific to 1.0.5 — current agy still denies on a bare `{}`. The descriptor agrees: `PreToolHookResult` is `decision` (string), `reason`, `overwrite`, `permission_overrides`, plus deprecated `allow_tool`/`deny_reason`. `PostToolHookResult` is an empty message, which confirms `{}` is right for the other events.

**2. Lifecycle events use a different shape, and getting it wrong killed the whole file**

This is the one that matters most. agy parses the two event families differently:

```jsonc
"PreToolUse":  [ { "matcher": "…", "hooks": [ { "type": "command", "command": "…" } ] } ],
"Stop":        [ { "type": "command", "command": "…" } ]
```

Tool events take the `{matcher, hooks}` wrapper; `PreInvocation`, `PostInvocation` and `Stop` take a flat handler list. Our manifest wrapped all four. agy then reads the wrapper itself as a handler, finds no `command` on it, and rejects the **entire file**:

```
Failed to parse hooks file …\.gemini\config\hooks.json:
  invalid hook "agentmemory": command hook must specify 'command'
```

Not just the bad event — the whole bundle, so nothing fired. Worse, `hooks.json` is shared, so a mis-shaped entry of ours would take out hooks any other tool had installed there. I bisected event by event to confirm: `PreToolUse` alone loads, `PostToolUse` alone loads, `PreInvocation` or `Stop` alone fails.

**3. Commands must be unquoted, and paths with spaces are impossible**

`command` is not run through a shell and agy does not strip quotes before splitting, so `node "<root>/scripts/antigravity-bridge.mjs"` produced:

```
Error: Cannot find module 'C:\Users\…\.gemini\config\"C:\…\antigravity-bridge.mjs"'
```

Note both the literal quotes and the resolution base — hooks run with cwd set to the `hooks.json` directory, not the workspace. Commands are bare now. The corollary is that a plugin path containing a space cannot be expressed at all; I tested quoted and unquoted with a space in the path and neither executes. So `connect antigravity-cli --with-hooks` now refuses with an explanation rather than writing a bundle that can only fail at tool time. MCP is unaffected.

**Verified end to end**: with `6e481cb`, the real bundle loads (`loaded 1 named hooks from 1 hooks.json file(s)`), the bridge runs, and the tool proceeds with no denials.

**One payload note for you**, not fixed here because I'd be guessing at the right behavior. A real `PreToolUse` payload:

```json
{"artifactDirectoryPath":"…/brain/53642203-…","conversationId":"53642203-…",
 "modelName":"gemini-3.6-flash-high","stepIdx":3,
 "toolCall":{"args":{"DirectoryPath":"C:\\Users\\u\\.gemini\\antigravity-cli"},"name":"list_dir"},
 "transcriptPath":"…/transcript_full.jsonl","workspacePaths":[]}
```

`conversationId` and the PascalCase args match what the bridge expects. But there is no `cwd` key, and `workspacePaths` came back empty in every headless run — combined with the cwd being the hooks.json dir, the bridge's `process.cwd()` fallback would attribute captures to `~/.gemini/config` rather than the repo. I don't know yet whether `workspacePaths` is populated in interactive sessions; if it usually is, this only affects `--print`, and if it isn't, the fallback needs rethinking. Happy to take direction, or to leave it for a follow-up.

The environment I tested on is restored — the probe `hooks.json` is removed and nothing else was touched.

plugin/scripts/antigravity-bridge.mjs
berthojoris: Done in `3a095ba` — the artifact is down to the three boilerplate lines every other bundled script has.

The cause was that the bundler strips `//` comments but preserves JSDoc blocks, so the `/** */` docs on the bridge's exported helpers were the only ones surviving into `plugin/scripts/`. Switched those to line comments: the explanation stays in `src/hooks/antigravity-bridge.ts` and the generated file now matches `pre-tool-use.mjs` and friends (3 lines, all of them `#region`/sourcemap markers).

src/cli/connect/antigravity-cli.ts
berthojoris: Trimmed in `3a095ba`. The file header and the per-function block were stating the same things twice — kept one statement of each and dropped the rest. Comment lines: `antigravity-cli.ts` 26 → 12, `antigravity-hooks.ts` 65 → 46, `antigravity-bridge.ts` 70 → 41, which puts them in line with `droid.ts`.

What I deliberately kept is the agy behaviour I verified against a live 1.0.15 — the two event shapes, the no-quoting rule, the PreToolUse decision contract. That part isn't derivable from the code, and each of those was a real defect here, so a future reader changing the manifest needs the reason. Happy to cut further if you'd rather that lived only in the PR description.

PR #1147, opened 2026-08-03
fix(docs): sync REST endpoint counts + skills reference (unblocks CI on main)

berthojoris:
## What

Updates the documented REST endpoint count from `129` to `130` in the three places `test/consistency.test.ts` checks.

## Why

`main` is currently red. `5023cf3` (#1132) registered a 130th `api_path:` in `src/triggers/api.ts` without updating any of the documented counts, so this assertion fails on every PR:

```
test/consistency.test.ts > Consistency checks > documented REST endpoint counts match registered API paths
AssertionError: expected '<p align="center">…' to contain '130 endpoints on port'
```

Failing CI run on `main` for `5023cf3`: https://github.com/rohitg00/agentmemory/actions?query=branch%3Amain

`AGENTS.md` already documents this coupling — *"When adding REST endpoints, you MUST update: `src/triggers/api.ts`, `src/index.ts`, `README.md`"*. This PR moves those two, plus the `AGENTS.md` stat line that the same test also asserts on.

- `src/index.ts:543` — boot log line
- `README.md:1501` — API section prose
- `AGENTS.md:120` — Current Stats

No endpoint is added or removed; this is documentation catching up to code.

## How to verify

```bash
npx vitest run test/consistency.test.ts
```

8 passed (was 1 failed / 7 passed).

The count itself is `Array.from(src.matchAll(/api_path:\s*["`]/g)).length` over `src/triggers/api.ts`, which is 130 at `5023cf3`.

## Summary by CodeRabbit

* **Documentation**
  * Updated the README and project guidance to reflect 130 available REST API endpoints.
  * Updated the REST API reference to report 119 registered endpoints.

* **Bug Fixes**
  * Corrected the startup readiness message so it reports the current total of 130 REST API endpoints.


berthojoris:
Pushed a second commit: `main` fails two independent CI gates from the same root cause, not one.

1. `npm test` → `test/consistency.test.ts` — `README.md` / `AGENTS.md` / `src/index.ts` still said 129 endpoints (first commit).
2. `npm run skills:check` → `plugin/skills/agentmemory-rest-api/REFERENCE.md` — the AUTOGEN:rest block still said 118 registered endpoints because `npm run skills:gen` was not run (second commit).

Both trace back to `5023cf3` (#1132) adding an endpoint without the follow-up updates `AGENTS.md` requires. Only the rest-api reference has a content change; the other four REFERENCE.md files regenerate byte-identical.

Issue #751, opened 2026-06-01
memory_save accepts `project` parameter but iii engine serialization drops it — `memory-project-coverage` diagnose perpetually fails


mikeExit:
## Summary

`memory_save` MCP tool defines `project` in its input schema and the handler (`src/mcp/server.ts:175-185`) correctly forwards it to `mem::remember`. The `remember` function (`src/functions/remember.ts:121`) spreads `project` onto the `Memory` object before `kv.set()`. However, the iii engine's serialization layer drops the field — it is never persisted to the binary store.

## Evidence

1. **Unit test passes** — `test/remember-project-scope.test.ts` mocks the KV store and verifies the field is written. This confirms the agentmemory code does the right thing.

2. **Real engine drops it** — All memories saved via `memory_save(project="juice")` show `project=<missing>` when read back from:
   - The `GET /agentmemory/memories` REST endpoint
   - The binary file (`mem%3Amemories.bin` at `./data/state_store.db/`)
   - The MCP `memory_recall` tool

3. **Direct binary file edit confirms engine rejects unknown fields** — Adding `project` to each memory in the `.bin` file caused the engine to start with 0 memories (silently recreating the file from scratch).

## Impact

- `memory_diagnose` check `memory-project-coverage` permanently shows failure
- `POST /agentmemory/migrate {"step":"infer-memory-projects"}` returns `ambiguous: 256, updated: 0`
- Sessions CAN be created with project scope (confirmed via `POST /agentmemory/session/start`), but the project never flows down to memories saved through any external API
- Cross-project isolation and `inferMemoryProjects()` backfill are both broken in production

## Root Cause

The `Memory` type (`src/types.ts`) defines `project?: string` and `remember.ts` sets it. The `inferMemoryProjects()` function in `src/functions/migrate.ts` also writes `project` via the same `kv.set()` path. This suggests the issue is in the **iii engine's `state::set` implementation** — it serializes memories using a fixed schema that rejects unknown keys.

## To Reproduce

```bash
npx @agentmemory/agentmemory  # start engine
# Save with project
npx @agentmemory/agentmemory mcp  # then call memory_save with project="test"
# Read back — project field is missing
curl http://localhost:3111/agentmemory/memories?limit=1 | jq '.memories[0].project'
# Output: null
```

Or directly via MCP tool:
- Call `memory_save` with `project: "juice"`
- Response object has no `project` field
- Binary file at `./data/state_store.db/mem%3Amemories.bin` confirms it's absent

[...]

PR #1195, opened 2026-08-13
fix(mcp): thread project param through standalone MCP bridge (#926, #787)

### LINKED ISSUE #926: memory_save & memory_recall silently drop the `project` parameter (MCP server)
**Package:** `@agentmemory/mcp` + `@agentmemory/agentmemory`
**Version:** 0.9.27 (latest as of 2026-06-13); present at least since 0.9.20
**Severity:** medium — silent data-quality bug (no error; memories are saved unscoped forever)

## Summary

The MCP tool `memory_save` advertises a `project` parameter in its `inputSchema` (with a detailed description: *"Stable canonical project identifier this memory belongs to… Must match the value used when the session was started"*), but the standalone MCP server **never reads or forwards it**. Every memory saved through the MCP tool is therefore persisted with **no `project` field**, regardless of what the caller passes.

This is invisible: the call returns `success: true`, and `memory_diagnose` later reports `"N of N latest memories have no project scope"`. The suggested fix (`migrate {step:"infer-memory-projects"}`) is itself a no-op for these memories, because MCP-saved memories also have empty `sessionIds`, so there is no session to infer the project from.

The **backend** (`mem::remember`) handles `project` correctly (`...project !== void 0 && { project }`) — only the MCP bridge layer drops it. Posting directly to `POST /agentmemory/remember` with `{..., "project": "X"}` scopes the memory as expected. So the bug is entirely in `src/mcp/standalone.ts`.

## Root cause — three drop points in `src/mcp/standalone.ts`

**(1) Argument extraction** (`case "memory_save"` in the args→`v` mapper): reads `content`, `type`, `concepts`, `files` — but never `args["project"]`.

```js
case "memory_save": {
  const content = args["content"];
  if (typeof content !== "string" || !content.trim()) throw new Error("content is required");
  v.content = content;
  v.type = args["type"] || "fact";
  v.concepts = normalizeList(args["concepts"]);
  v.files = normalizeList(args["files"]);
  return v;        // <-- v.project never set
}
```

**(2) Proxy mode** (`handleProxy`): forwards a body without `project`.

```js
case "memory_save": return textResponse(await handle.call("/agentmemory/remember", {
  method: "POST",
  body: JSON.stringify({ content: v.content, type: v.type, concepts: v.concepts, files: v.files })
  // <-- project missing
}));
```

**(3) Local mode** (`handleLocal`): writes the memory object without `project` (and with `sessionIds: []`).

## Same bug on the read path

`memory_recall` and `memory_smart_search` have the identical omission, which makes the feature unusable end-to-end via MCP: even correctly-scoped memories can never be recalled by project through the MCP tools.

- The recall dispatch reads `query`, `limit`, `format`, `token_budget` — never `args["project"]`.
- `handleProxy` builds the `/agentmemory/search` (and `/smart-search`) body as `{ query, limit, format, token_budget? }` — no `project`, no `cwd` — so the backend always runs with `hasProjectFilter: false`, even though it supports a project filter and a 1.5× same-project relevance boost.

## Proposed fix

Thread `project` through, symmetric on both paths:

```diff
  // (1) save arg extraction
   v.files = normalizeList(args["files"]);
+  if (typeof args["project"] === "string" && args["project"].trim().length > 0) v.project = args["project"].trim();
   return v;
```
```diff
  // (2) save proxy body
   files: v.files,
+  ...(v.project !== undefined && { project: v.project }),
```
```diff
  // recall arg extraction
   v.query = query.trim();
+  if (typeof args["project"] === "string" && args["project"].trim().length > 0) v.project = args["project"].trim();
```
```diff
  // recall + smart-search proxy bodies
   if (v.tokenBudget != null) body["token_budget"] = v.tokenBudget;
+  if (v.project !== undefined) body["project"] = v.project;
```
Also add a `project` property to the `memory_recall` / `memory_smart_search` input schemas (`memory_save` already has it), and audit `memory_sessions` / `memory_export` for the same omission.

## Reproduction

1. `memory_save({ content: "test", project: "MyProj" })` via MCP → `success: true`.
2. `GET /agentmemory/memories/<id>` → the stored object has **no `project`** field.
3. `memory_diagnose` → `"… latest memories have no project scope"`.
4. By contrast `POST /agentmemory/remember {"content":"test","project":"MyProj"}` → the stored object **does** have `"project":"MyProj"`.

sk91:
## Summary

Fixes #926 and #787.

`memory_save` accepts a `project` field but silently drops it — every memory
saved through the MCP tool ends up unscoped, no error, nothing to indicate it.
`memory_recall`/`memory_smart_search` don't accept `project` at all, so even
correctly-scoped memories can't be filtered back out through the MCP tools.

End result: nothing saved via `memory_save` ever shows up in a project's
SessionStart context injection, since that's scoped by project.

## Fix

`project` now flows through on all three tools — save, recall, and smart-search
— and stays optional, so existing unscoped usage is unaffected.

## Testing

Verified a memory saved with `project` set now persists correctly, is filtered
correctly on read, and actually appears in that project's SessionStart digest
afterward. Full suite passes.

## Out of scope

A related report (#751, project dropped at the storage layer on a different
deployment) didn't reproduce here — leaving it open.

## Summary by CodeRabbit

* **New Features**
  * Added optional project filtering to memory recall and smart search.
  * Project values are trimmed and applied consistently when saving, recalling, or searching memories.
  * Local memory results can be scoped to a specific project, while unfiltered searches include all projects.
  * Smart search applies project filters across sessions and memory records.

* **Tests**
  * Added coverage for project forwarding, filtering, isolation, and fallback behavior.


sk91:
Addressed all three in 78833f7:

1. mem::smart-search now filters hybrid results by project too, not just lessons — same session→project resolution as mem::search (with the KV.memories fallback for synthetic `memory` sessions).
2. handleLocal fallback now persists/filters project as well, so it doesn't get dropped when the proxy is down.
3. Added a whitespace-trim test.

New tests in smart-search.test.ts and mcp-standalone-proxy.test.ts, full suite still green.

sk91:
Fixed in 4a3d542:

1. expandIds now goes through the same project filter — moved projectFilter before the early-return branch, and recallLessons() gets the trimmed value instead of raw data.project.
2. filterByProject is now generic and resolves lookups via Promise.all with promise-cached session/memory lookups, instead of a serial loop.
3. Added graphScore to the two new test fixtures.

Issue #1197, opened 2026-08-14
memory_save MCP path drops agentId at three more points — schema never exposes it, and two forwarding hops drop it independently of #1159/#1160


kylen1211:
**Version:** @agentmemory/agentmemory 0.9.28 (npm), Node 24.18.0, Linux x86_64

## Summary

#1159 and #1160 cover the REST shim (`api::remember`) and the read-side `memoryToObservation()` conversion. There are three more independent `agentId` drops on the **MCP tool path** for `memory_save`, upstream of both of those:

1. `src/mcp/tools-registry.ts` (`memory_save` tool, L54-89) — `inputSchema.properties` declares `content`/`type`/`concepts`/`files`/`project` but never `agentId`. MCP clients (Claude Code, other hosts) have no schema-legal way to pass it at all — this is upstream of #1159/#1160, since fixing those doesn't help if the field can't reach them in the first place.
2. `src/mcp/server.ts` `memory_save` switch branch (L168-196) builds the `mem::remember` payload from `content`/`type`/`concepts`/`files`/`project` only — no `agentId`, unlike the same file's `memory_recall` branch (L116-129) which already forwards it (comment references #817). This is the `/agentmemory/mcp/call` MCP-call endpoint, a separate code path from `api::remember` (#1159's REST shim).
3. `src/mcp/standalone.ts` (the `@agentmemory/mcp` npm package — the stdio bridge most MCP hosts actually run) has the same gap in its own pipeline: `Validated` interface (L96-107), `validate()`'s `memory_save` case (L116-121), and `handleProxy()`'s `memory_save` case (L175-183) parse/forward `content`/`type`/`concepts`/`files` only. Notably this drops `project` too, even though the tool schema already exposes `project` (L78-85) — so `project` scoping is also silently broken specifically through the standalone stdio package.

Net effect: even after #1159/#1160 land, an MCP client calling `memory_save` still can't scope a memory to an agent (or, via the stdio package, to a project) — the schema never offers the parameter, and two independent forwarding hops drop it if a client patches around the schema gap.

## Reproduction

1. `tools/list` (or `getAllTools()`) → inspect `memory_save.inputSchema.properties` → no `agentId` key.
2. Patch a client to send `agentId` anyway via `tools/call` on `memory_save` → payload at `sdk.trigger({function_id:"mem::remember", ...})` in `server.ts` has no `agentId` field.
[...]

Issue #1159, opened 2026-08-06
REST shims silently drop agentId on /agentmemory/remember and /agentmemory/observe — per-request multi-agent scoping impossible over REST


benjaml4:
# REST API silently drops `agentId` on /agentmemory/remember and /agentmemory/observe — per-request multi-agent scoping impossible over REST

**Version:** @agentmemory/agentmemory 0.9.28 (npm), iii-engine 0.11.2, Linux x86_64, Node 22.23.2

## Summary

The internal functions `mem::remember` and `mem::observe` both accept a per-call
`agentId` (and `mem::smart-search` filters on it), but the HTTP shims
(`api::remember`, `api::observe`) build an explicit field whitelist from
`req.body` that does **not** include `agentId`. The field is silently discarded,
so every memory/observation stored over REST falls back to the env identity
(`getAgentId()`), or is stored unscoped when `AGENT_ID` is unset.

With `AGENTMEMORY_AGENT_SCOPE=isolated` this breaks scoping invisibly: writes
claim to succeed (201), but rows carry the wrong identity and scoped recalls
return 0 results with no error anywhere.

## Reproduction

```bash
# server: AGENT_ID=global AGENTMEMORY_AGENT_SCOPE=isolated
curl -s -X POST localhost:3111/agentmemory/remember \
  -H 'Content-Type: application/json' \
  -d '{"content":"test row","agentId":"health-agent"}'
# response: {"memory":{... "agentId":"global" ...}}   <-- env identity, not the payload's

curl -s -X POST localhost:3111/agentmemory/smart-search \
  -H 'Content-Type: application/json' \
  -d '{"query":"test row","agentId":"health-agent"}'
# 0 results — smart-search DOES forward agentId, so read and write paths disagree
```

## Root cause

`api::remember`'s payload construction spreads a fixed list
(`content`, `tags`, `concepts`, `files`, `type`, `ttlDays`,
`sourceObservationIds`, `project`) and omits `agentId`. Same for `api::observe`.
`api::smart-search` forwards it, making the asymmetry surprising.

## Suggested fix

Forward `agentId` in both shims, e.g.:

```js
...req.body.agentId !== void 0 && { agentId: req.body.agentId },
```

Use case: one agentmemory server behind a multi-agent gateway (OpenClaw with 7
logical agents) where a single process-wide `AGENT_ID` cannot represent the
caller; per-request identity is the only workable model.

kylen1211:
[...]

PR #1198, opened 2026-08-14
fix(state): memoryToObservation transparently passes through agentId and project

### LINKED ISSUE #1160: memoryToObservation() drops agentId — memories invisible to every agent-scoped search (residual hole in the #817 fix)
# memoryToObservation() drops `agentId` — memories are invisible to every agent-scoped search

**Version:** @agentmemory/agentmemory 0.9.28 (npm), iii-engine 0.11.2, Linux x86_64, Node 22.23.2

## Summary

Scoped search filters use strict equality on the hydrated result:

```js
// smart-search / hybrid search post-filter
if (filterAgentId !== undefined && obs.agentId !== filterAgentId) continue
```

Observations hydrate fine, but memories reach the filter through
`memoryToObservation(mem)` (used both at index time and in the KV-fallback
hydration path for `mem_*` ids), and that converter does not copy `agentId`
even though the `Memory` record stores it (`remember` writes
`...(callAgentId ? { agentId: callAgentId } : {})`).

Net effect: **a memory stored with `agentId: "X"` can never be found by a
search filtered to `agentId: "X"`** — only unfiltered / wildcard (`"*"`)
searches see it. Combined with `AGENTMEMORY_AGENT_SCOPE=isolated`, memories
effectively vanish for the very agents that own them.

## Reproduction

```bash
curl -s -X POST localhost:3111/agentmemory/remember \
  -H 'Content-Type: application/json' \
  -d '{"content":"the wombat clinic visit is on Mondays","agentId":"health-agent"}'
# (requires issue #1 fixed, or store via MCP so agentId reaches mem::remember)

curl -s -X POST localhost:3111/agentmemory/smart-search \
  -H 'Content-Type: application/json' \
  -d '{"query":"wombat clinic","agentId":"health-agent"}'   # -> 0 results
curl -s -X POST localhost:3111/agentmemory/smart-search \
  -H 'Content-Type: application/json' \
  -d '{"query":"wombat clinic","agentId":"*"}'              # -> 1 result
```

## Suggested fix

```js
function memoryToObservation(memory) {
  return {
    // ...existing fields...
    importance: memory.strength,
    ...(memory.agentId ? { agentId: memory.agentId } : {}),
  };
}
```

Verified locally: with this one-line change (plus issue #1), the full isolation
matrix behaves — owner sees its row, other agents get 0 hits, `"*"` sees all.

Related: consolidation-derived memories appear to lose `agentId` as well (the
consolidate path never threads it), which reopens the same hole for derived
rows; worth covering in the same fix.

kylen1211:
## What

`memoryToObservation()` (`src/state/memory-utils.ts`) now copies `agentId` and `project` from the underlying `Memory` record onto the `Observation` it returns. Also adds the (previously missing) `project` field to `CompressedObservation` in `src/types.ts` so the type accepts what this fix now sets.

## Why

Scoped search filters apply strict equality on the hydrated result (`obs.agentId !== filterAgentId`), but memories reach that filter through `memoryToObservation()` — used both at index time and in the KV-fallback hydration path for `mem_*` ids — and that converter dropped `agentId` even though `remember()` already stores it on the `Memory` record. Net effect: a memory stored with `agentId: "X"` could never be found by a search filtered to `agentId: "X"`, only unfiltered/wildcard searches saw it. Combined with `AGENTMEMORY_AGENT_SCOPE=isolated`, memories effectively vanished for the very agents that owned them.

This is a small, standalone fix on the read side only — it doesn't touch or depend on any of the write-side gaps (#1159, #1197) also open right now. It closes the "written with a scope tag but unfindable by that tag" symptom for any memory that already has `agentId`/`project` set on it (e.g. rows written directly against `mem::remember`, or once the write-side issues land).

Fixes #1160

## How to verify

```
npm test -- test/memory-to-observation-agent-scope.test.ts
```

Both new cases pass: the passthrough case (`agentId`/`project` set → preserved) and the boundary case (both `undefined` → stays `undefined`, no throw, no default fill-in). Full suite (`npm test`) passes unchanged otherwise — 1598 passed / 1 skipped.

## Summary by CodeRabbit

- **New Features**
  - Compressed observations now retain project information and agent context when available.
  - Missing project or agent details remain optional without causing errors.

- **Tests**
  - Added coverage to verify context is preserved and undefined values are handled correctly.

### DIFF
--- a/src/state/memory-utils.ts
+++ b/src/state/memory-utils.ts
@@ -20,5 +20,7 @@ export function memoryToObservation(memory: Memory): CompressedObservation {
     concepts: memory.concepts,
     files: memory.files,
     importance: memory.strength,
+    agentId: memory.agentId,
+    project: memory.project,
   };
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -61,6 +61,7 @@ export interface CompressedObservation {
   imageDescription?: string;
   modality?: "text" | "image" | "mixed";
   agentId?: string;
+  project?: string;
 }
 
 export type ObservationType =
--- a/test/memory-to-observation-agent-scope.test.ts
+++ b/test/memory-to-observation-agent-scope.test.ts
@@ -1,0 +1,39 @@
+import { describe, it, expect } from "vitest";
+
+import { memoryToObservation } from "../src/state/memory-utils.js";
+import type { Memory } from "../src/types.js";
+
+function makeMemory(overrides: Partial<Memory> = {}): Memory {
+  return {
+    id: "mem_test_1",
+    createdAt: "2026-08-14T00:00:00.000Z",
+    updatedAt: "2026-08-14T00:00:00.000Z",
+    type: "fact",
+    title: "test memory",
+    content: "test content",
+    concepts: [],
+    files: [],
+    sessionIds: [],
+    strength: 1,
+    version: 1,
+    isLatest: true,
+    ...overrides,
+  };
+}
+
+describe("memoryToObservation agentId/project passthrough", () => {
+  it("transparently passes through agentId and project", () => {
+    const memory = makeMemory({ agentId: "pm-agent", project: "worker-config" });
+    const observation = memoryToObservation(memory);
+    expect(observation.agentId).toBe("pm-agent");
+    expect(observation.project).toBe("worker-config");
+  });
+
+  it("leaves agentId/project undefined without defaults or errors", () => {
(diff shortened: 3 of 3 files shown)


kylen1211:
For context, this is one of a 5-PR set closing the full memory_save agentId chain — full picture and cross-links in #1197. This particular PR (read-side memoryToObservation) is independent of the other four; it doesn't need any of them merged to be correct on its own.

PR #1222, opened 2026-08-17
feat: klaatcode support (cli adapter, native hooks)

shivaylamba:
Adds [Klaat Code](https://github.com/KlaatAI/klaatcode) (npm `klaatai`) as the 22nd supported harness, following the shape of the Devin adapter in #1214.

Klaat Code is a Bun/TypeScript terminal coding agent with a full MCP client (stdio + Streamable HTTP + OAuth 2.1), a lifecycle hooks system, and a skills loader — so it can use all three agentmemory surfaces.

## What this adds

**MCP** — `agentmemory connect klaatcode` merges the standard block into `~/.klaatai/mcp.json`. Klaat Code keys its servers under `servers` rather than `mcpServers`, so the adapter reuses the same `wrapperKey` escape hatch Zed already needs for `context_servers`. Klaat Code supports `${VAR:-default}` expansion, so `AGENTMEMORY_MCP_BLOCK` goes in verbatim.

**Hooks** — `--with-hooks` installs six native auto-capture hooks into `~/.klaatai/hooks.json`.

| Klaat Code event | script |
|---|---|
| `session_start` | `session-start.mjs` |
| `before_message` | `prompt-submit.mjs` |
| `before_tool` | `pre-tool-use.mjs` |
| `after_tool` | `post-tool-use.mjs` |
| `after_message` | `stop.mjs` |
| `session_end` | `session-end.mjs` |

Klaat Code's hooks config is **flat** — event name straight to a list of entries, where an entry is a bare shell string (v1) or `{ command, matcher?, timeout? }` (v2) — and its event names are snake_case. `buildMergedHooks()` in `codex-hooks.ts` assumes the nested Claude shape, so it does not apply here. `klaatcode-hooks.ts` is a small sibling that keeps the same two guarantees:

- `${CLAUDE_PLUGIN_ROOT}` resolved to the absolute bundled plugin path, so the written file needs no env expansion
- re-installs strip our previous entries first, so upgrading never duplicates and never touches the user's own hooks (verified for both v1 string and v2 object entries)

**Hook payload compat** — Klaat Code names the project path `project_root` and exports `KLAATAI_PROJECT_ROOT`, neither of which `hookCwd()` knew about, so observations would have been attached to the wrong project. Both are now in the resolution chain alongside the existing Devin and Claude Code fallbacks. `plugin/scripts/*.mjs` are rebuilt accordingly (the only diff in those files is this one change).

## Deliberately not done

- **No context injection.** Klaat Code discards hook stdout except for `before_tool` block decisions, so there is no return channel. `contextPayload()` in `session-start.ts` is untouched — and `AGENTMEMORY_INJECT_CONTEXT` is default-off since #143 anyway. Recall runs through the MCP tools and the skills.
- **No guideline entry.** `guidelineTargets()` is for hook-less agents; Klaat Code has hooks.
- **No skills wiring needed.** Klaat Code already auto-loads `~/.claude/skills` and `.claude/skills`, so `npx skills add rohitg00/agentmemory` covers all 17 with no extra step.
- **No website grid entry** — I don't have a logo asset for Klaat Code. Happy to add one if you'd like to point me at it.

## Known limitation

Hooks and MCP are wired into Klaat Code's interactive TUI only — `klaatai run` (headless), `klaatai serve`, and ACP sessions are not captured. This is upstream's architecture, not something the adapter can work around. It's called out in the README row and printed after install.

## Testing

`npm run build && npm run skills:check && npm test` — green (1665 passed, 1 skipped).

New `test/klaatcode-connect-hooks.test.ts` covers placeholder rewriting, the six-event set, the snake_case tool matcher (and that it rejects Claude's `Edit`/`Write`/`Bash`), user-hook preservation in both entry forms, re-install idempotency, `project_root` / `KLAATAI_PROJECT_ROOT` resolution, and fail-open when the server is unreachable. `test/cli-connect.test.ts` bumped to 22 adapters.

Manually verified against a temp `HOME`:

- `--dry-run --with-hooks` reports the merge and writes nothing
- real install merges `servers.agentmemory` alongside a pre-existing `github` server, and appends hooks alongside a pre-existing user `after_message` string hook and `before_tool` guard script
- three consecutive `--force` re-installs produce a byte-identical `hooks.json`, 6 agentmemory entries total, no `${CLAUDE_PLUGIN_ROOT}` leakage

End-to-end against a running server: feeding Klaat-Code-shaped payloads through `session-start` → `post-tool-use` → `session-end` produced a session with `project: klaat-e2e`, `cwd: /tmp/klaat-e2e`, `status: completed` and the observation captured with `channel: tool`, `detail: edit_file`. The MCP shim handshakes at protocol `2025-06-18` and lists all 54 tools.

## Related

Pairs with KlaatAI/klaatcode#57, which asks for this integration from the Klaat Code side. A companion PR there adds an `agentmemory` MCP preset so `/mcp enable agentmemory` works without leaving the TUI.

## Summary by CodeRabbit

* **New Features**
  * Added Klaat Code integration through MCP.
  * Added optional hooks for session, message, and tool activity.
  * Added automatic project-root detection and agentmemory skill discovery.
  * Added onboarding support, setup documentation, verification steps, and limitations.
  * Added guideline generation for Klaat Code projects.
  * Captures interactive TUI sessions through supported hooks.

* **Bug Fixes**
  * Improved hook reliability across project-directory configurations.
  * Preserved custom hooks during installation and reinstallation.
  * Prevented duplicate bundled hooks during repeated setup.


shivaylamba:
Two updates: the logo is in, and I've now verified this end-to-end against a **real Klaat Code session** rather than synthetic payloads.

### Logo

Added `website/public/klaatcode.png` plus the `website/components/Agents.tsx` entry and a README grid tile, so the "no logo asset" caveat in the description is resolved.

### End-to-end, driven through the actual TUI

Previously I'd only fed Klaat-Code-shaped payloads to the hook scripts directly. I've now wired a temp `HOME` with `agentmemory connect klaatcode --with-hooks`, launched the real `klaatcode` TUI in a sized pty, and had it run a live task (read a file, write a file).

The sidebar showed `MCP Servers 1 — agentmemory`, and agentmemory captured the session unprompted:

```
id:      2026-08-17T21-23-52-a6r6      <- Klaat Code's own session id format
project: klaat-tui
cwd:     /private/tmp/klaat-tui        <- resolved from `project_root`
status:  completed                     <- session_end fired
observations: 3
```

with the observations proving three distinct hook events fired natively:

| channel | tool | type | hook |
|---|---|---|---|
| `user` | — | `conversation` | `before_message` |
| `tool` | `read_file` | `file_read` | `after_tool` |
| `tool` | `write_file` | `file_write` | `after_tool` |

That covers `session_start` (session registered), `before_message`, `after_tool`, and `session_end` (status `completed`) through Klaat Code's real dispatcher — and confirms the `project_root` fix in `_project.ts` is what makes the project resolve correctly.

Separately, driving Klaat Code's own `MCPManager` against the config this adapter writes connects in **12.1s and exposes all 54 tools** as `mcp__agentmemory__*`.

### One finding worth flagging

The first connect in a cold environment hit Klaat Code's 30s stdio init timeout while `npx -y @agentmemory/mcp` downloaded. It recovers on reconnect and affects every npx-based MCP server equally, so nothing to change here — but it's the reason a first-run user might see the server as disconnected before it settles.

I also found and fixed a related bug on the Klaat Code side (KlaatAI/klaatcode#58): native `.klaatai/mcp.json` entries weren't expanding `${VAR}` refs, so a user's custom `AGENTMEMORY_URL` was silently ignored. Our shim's literal-placeholder guard in `rest-proxy.ts:41-46` masked it for the default case, which is exactly what that guard is for.

Full gate still green: `npm run build && npm run skills:check && npm test` — 1665 passed.

shivaylamba:
Thanks — went through the three review findings. One was real and is now fixed; two I've skipped with reasons.

**1. First non-blank project dir env var — fixed** (`0178ec9`)

Valid catch, and it was mine: extending the `||` chain to a third variable meant a whitespace-only `DEVIN_PROJECT_DIR` would short-circuit the chain, fail the subsequent `.trim()` check, and leave the project unresolved instead of falling through to `KLAATAI_PROJECT_ROOT`. Now selects the first non-blank var. Regression test added, and the bundled `plugin/scripts/*.mjs` are rebuilt from it.

**2. Remove explanatory comments — skipping**

These explain *why*, not *what*: which host quirk each branch exists for. The direct sibling of the new file, `src/cli/connect/codex-hooks.ts`, opens with a ~30-line block comment of exactly this kind documenting the Codex plugin-dispatch workaround, and `klaatcode-hooks.ts` exists for the same category of reason (Klaat Code's flat, snake_case hook shape). Matching the established house style seemed better than diverging from it in one new file. Happy to trim if you'd rather they go.

**3. `__dirname` in an ESM test — not applicable**

Vitest provides `__dirname` in its module transform, so this works as written; `test/devin-connect-hooks.test.ts` already uses the identical `resolve(__dirname, "..", "plugin")` line. The suite passes (1666 tests), which wouldn't be the case if `__dirname` were undefined here.

Full gate re-run after the fix: `npm run build && npm run skills:check && npm test` — green, 1666 passed.

shivaylamba:
Correcting my previous comment: I was wrong to push back on the comment findings.

I justified keeping them by pointing at `codex-hooks.ts` as precedent, but I should have checked the written rule first. `AGENTS.md:102` says "No code comments explaining WHAT — use clear naming instead", and `CONTRIBUTING.md:59` says the same, allowing a comment only where the *why* is non-obvious. `CONTRIBUTING.md:41` also makes CodeRabbit green a merge gate, so leaving those open was blocking the PR on my own preference. Existing code that predates the rule isn't licence to add more.

Trimmed in the latest commit:

- `src/cli/connect/klaatcode.ts` — removed the config-layout block entirely; the constant names carry it.
- `src/cli/connect/klaatcode-hooks.ts` — 26-line header down to 8, keeping only the two things not visible from the code: why this file exists separately from `codex-hooks.ts`, and why a `${CLAUDE_PLUGIN_ROOT}` token appears for a host that never sets it.
- `src/hooks/_project.ts` — dropped both the `project_root` comment and the first-non-blank rationale.

Full gate re-run: `npm run build && npm run skills:check && npm test` — green, 1666 passed.

The `__dirname` finding I'm still leaving as-is: Vitest provides it in its transform, `test/devin-connect-hooks.test.ts` uses the identical line, and the suite passes. Happy to switch it to `fileURLToPath` anyway if you'd prefer consistency over precedent — just say the word.

shivaylamba:
I went back and tested the thing that actually matters — does a memory saved in one Klaat Code session come back in the next — rather than just proving hooks fire. It surfaced a gap in this PR, now fixed, and one limitation worth your call.

### The gap: recall was inert

I originally skipped `guidelineTargets()` for Klaat Code, reasoning that it has hooks so it's in the same bucket as Claude Code and Codex. That was wrong. Those two can inject recalled context back through their hooks; **Klaat Code discards hook stdout** except for `before_tool` block decisions. So capture worked fine, but nothing ever told the model to call `memory_recall` — the half users actually notice was doing nothing.

Fixed by adding a `klaatcode` target writing to `~/.klaatai/rules.md`, which `system-prompt.ts` loads globally ahead of per-project rules, in the same marked-block format as the other shared instruction files.

### Verified end to end, in the real TUI

Session 1 — model calls `mcp__agentmemory__memory_save`:
```
⏺ Mcp:agentmemory memory_save
  ⎿ {"memory":{"concepts":["billing","stripe","payment-intents","checkout","split-payouts"], …
```

Session 2 — brand new session, same repo, only the guideline in context:
```
⏺ Mcp:agentmemory memory_recall         "billing integration change"
⏺ Mcp:agentmemory memory_smart_search   "billing integration change reason"
  ⎿ "Moved billing from Stripe Checkout to Stripe Payment Intents …"  score 1
```

Unprompted recall of a fact from a previous session. That's the feature working. Also confirms the 54 tools connect in the real TUI (`● agentmemory 54t`) and that Klaat Code picks up the skills — it read `skills/remember/SKILL.md` off `~/.agents/skills` with no extra wiring.

### The limitation — your call on whether it blocks

Project scoping does not survive the MCP path, and this isn't Klaat Code-specific. The hook path resolves correctly (server logged `{"project":"memproj"}`), but `memory_save`'s schema says the project "must match the value used when the session was started" while **nothing in the MCP surface tells the model what that value is**. `resolveProject` is only wired into `src/hooks/*`. In my run the model filled the field with a plausible-looking but wrong slug, and the memory was filed under an unrelated project.

Recall still found it — because the model doesn't scope its queries either, so search runs unscoped. Correct behaviour today, by accident. On a machine with many projects it would surface as cross-project bleed.

I've left it alone here because a fix belongs in the MCP layer for every host, not in one adapter — but say the word and I'll take it in a follow-up.

Gate green after the change: `npm run build && npm run skills:check && npm test` — 1667 passed.

One environmental note: roughly half my TUI runs died on `KlaatAI is temporarily unavailable` mid-request. Unrelated to agentmemory — a plain headless prompt hit the same API fine.

PR #1231, opened 2026-08-20
fix(mcp): pin matching core package version

### LINKED ISSUE #668: @agentmemory/mcp does not pin the matching core version, allowing fixed MCP behavior to regress
## Summary

`@agentmemory/mcp@0.9.22` declares `@agentmemory/agentmemory` as `~0.9.0`, which allows the MCP shim to run against an older core package such as `0.9.21`.

This matters because MCP behavior fixes may live in the core package. In this case, the `AGENTMEMORY_URL` placeholder handling fixed in `0.9.22` was not active when `mcp@0.9.22` resolved `agentmemory@0.9.21`, so the MCP shim fell back to empty local storage even though a healthy `0.9.22` backend had sessions.

Related context: #510 documents the placeholder handling issue that was fixed in `0.9.22`. This report is about the package version relationship that can prevent that fix from being used.

## Environment

- OS: macOS 26.5
- Node: v26.0.0
- `@agentmemory/mcp`: `0.9.22`
- Backend `@agentmemory/agentmemory`: `0.9.22`

Verified public npm metadata:

```json
{
  "name": "@agentmemory/mcp",
  "version": "0.9.22",
  "dependencies": {
    "@agentmemory/agentmemory": "~0.9.0"
  }
}
```

## Steps to reproduce

1. Start an `@agentmemory/agentmemory@0.9.22` server with existing sessions.
2. Run the plugin-provided MCP shim through `npx -y @agentmemory/mcp`.
3. Observe an installation/cache state where:
   - `@agentmemory/mcp` is `0.9.22`
   - its installed `@agentmemory/agentmemory` dependency is `0.9.21`
4. Call `memory_sessions` through MCP.
5. Remove the stale `npx` install and allow it to reinstall so both packages resolve to `0.9.22`.
6. Call `memory_sessions` again.

## Actual behavior

With this version combination:

```text
@agentmemory/mcp             0.9.22
@agentmemory/agentmemory     0.9.21
```

`memory_sessions` returned an empty sessions array, while the healthy backend HTTP endpoint returned existing sessions.

After reinstalling so that the MCP shim loaded:

```text
@agentmemory/mcp             0.9.22
@agentmemory/agentmemory     0.9.22
```

`memory_sessions` immediately returned the backend session list successfully.

## Expected behavior

A released MCP shim should reliably execute against the matching core implementation containing the fixes for that release. Installing or reusing `@agentmemory/mcp@0.9.22` should not produce behavior from `@agentmemory/agentmemory@0.9.21`.

## Evidence

In `@agentmemory/agentmemory@0.9.22`, the MCP proxy now handles unexpanded environment placeholders by treating `${...}` as unset and falling back to the default URL:

```js
function resolveEnvOrEmpty(name) {
  const raw = process.env[name];
  if (!raw) return "";
  if (raw.startsWith("${") && raw.endsWith("}")) return "";
  return raw;
}
```

That fix is not present when the MCP shim loads core `0.9.21`.

## Suggested fix

Pin the MCP package's core dependency to its matching release version rather than a compatible range:

```json
{
  "version": "0.9.22",
  "dependencies": {
    "@agentmemory/agentmemory": "0.9.22"
  }
}
```

It would also be useful to enforce this relationship in the release process or CI, so `@agentmemory/mcp@X.Y.Z` cannot publish unless it depends on `@agentmemory/agentmemory@X.Y.Z`.

For plugin manifests that launch the MCP package through `npx`, consider using an explicitly coordinated MCP version generated during the plugin release, for example:

```json
"args": ["-y", "@agentmemory/mcp@0.9.22"]
```

This would make an installed plugin release reproducible instead of depending on whichever MCP/core combination is currently resolved at runtime.

cmondragon023:
## Summary

- pin `@agentmemory/mcp@0.9.29` to the exact matching `@agentmemory/agentmemory@0.9.29` core package
- extend the existing package consistency test so future release bumps fail unless the shim version, root version, and core dependency stay synchronized

Closes #668

## Testing

- `node_modules/.bin/vitest run test/consistency.test.ts` — 9 tests passed
- `node_modules/.bin/vitest run --exclude test/integration.test.ts` — 151 test files passed, 1 skipped; 1,657 tests passed, 1 skipped
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved package version consistency by pinning the MCP package dependency to the exact matching release.
* **Tests**
  * Added validation to ensure the MCP dependency version stays aligned with the root package version.

### DIFF
--- a/packages/mcp/package.json
+++ b/packages/mcp/package.json
@@ -28,7 +28,7 @@
   "homepage": "https://github.com/rohitg00/agentmemory#readme",
   "bugs": "https://github.com/rohitg00/agentmemory/issues",
   "dependencies": {
-    "@agentmemory/agentmemory": "~0.9.0"
+    "@agentmemory/agentmemory": "0.9.29"
   },
   "publishConfig": {
     "access": "public",
--- a/test/consistency.test.ts
+++ b/test/consistency.test.ts
@@ -35,13 +35,14 @@ describe("Consistency checks", () => {
     expect(plugin.version).toBe(pkg.version);
   });
 
-  it("packages/mcp version matches package.json", () => {
+  it("packages/mcp version and core dependency match package.json", () => {
     // The mcp package publishes in lockstep with the main package but its
     // version lives in its own manifest; without this guard a release bump
     // can silently ship a stale @agentmemory/mcp (it slipped in 0.9.29).
     const pkg = JSON.parse(readText("package.json"));
     const mcp = JSON.parse(readText("packages/mcp/package.json"));
     expect(mcp.version).toBe(pkg.version);
+    expect(mcp.dependencies["@agentmemory/agentmemory"]).toBe(pkg.version);
   });
 
   it("export-import.ts supports current version", () => {


cmondragon023:
The dependency mismatch remains on current `main`, which still uses `~0.9.0`. The branch is mergeable and CodeRabbit reported no actionable findings; the only red status is Vercel deployment authorization. Could a maintainer review the change and authorize the deployment when convenient?

PR #1250, opened 2026-08-26
fix(clip): use CLIP text tower for text embeddings

### LINKED ISSUE #1249: vision-search queryText always fails: CLIP text branch uses feature-extraction, which demands pixel_values
`vision-search` with `queryText` always fails on 0.9.29 once image embeddings are enabled. The image path works, so the feature looks half-broken: you can embed and search images, but you cannot search them by text — which is the main reason to enable it.

```console
$ curl -s -X POST -H 'Content-Type: application/json' \
    -d '{"queryText":"docker compose diagram","limit":3}' \
    http://127.0.0.1:3111/agentmemory/vision-search
{"error":"query embed failed: An error occurred during model execution: \"Missing the following inputs: pixel_values.","success":false}
```

For contrast, on the same store, same process, the image branch is fine:

```console
$ curl -s -X POST -H 'Content-Type: application/json' \
    -d '{"imageRef":"<managed>/<sha256>.png"}' \
    http://127.0.0.1:3111/agentmemory/vision-embed
{"dimensions":512,"imageRef":"...","success":true}

$ curl -s -X POST -H 'Content-Type: application/json' \
    -d '{"queryImageRef":"<managed>/<sha256>.png","limit":3}' \
    http://127.0.0.1:3111/agentmemory/vision-search
{"results":[{"imageRef":"...","score":1,"updatedAt":"..."}],"success":true,"total":1}
```

## Cause

`ClipEmbeddingProvider` builds **both** extractors from the same model id, but the text one goes through the generic `feature-extraction` pipeline (`src/providers/embedding/clip.ts`):

```js
const DEFAULT_MODEL = "Xenova/clip-vit-base-patch32";

async getTextExtractor() {
  if (this.textExtractor) return this.textExtractor;
  const t = await loadTransformers();
  this.textExtractor = await t.pipeline("feature-extraction", this.modelId, { dtype: "q8" });
  return this.textExtractor;
}

async getImageExtractor() {
  if (this.imageExtractor) return this.imageExtractor;
  const t = await loadTransformers();
  this.imageExtractor = await t.pipeline("image-feature-extraction", this.modelId, { dtype: "q8" });
  return this.imageExtractor;
}
```

`feature-extraction` on a CLIP repo instantiates the **full dual-encoder** `CLIPModel`, whose `forward` requires `input_ids` *and* `pixel_values`. The text-only call supplies just `input_ids`, so ONNX rejects it. `image-feature-extraction` maps to the vision tower alone, which is why only that branch works.

Because `embed()` delegates to `embedBatch()` → `getTextExtractor()`, every text-side entry point fails: `vision-search` with `queryText`, and any `queryImageBase64`/`queryImageRef` path that also embeds text.

## Minimal repro

Standalone, no agentmemory involved — run from a dir with `@huggingface/transformers` installed:

```js
const MODEL = "Xenova/clip-vit-base-patch32";
const t = await import("@huggingface/transformers");

// current code
try {
  const ex = await t.pipeline("feature-extraction", MODEL, { dtype: "q8" });
  await ex(["a docker compose diagram"], { pooling: "mean", normalize: true });
  console.log("[1] OK");
} catch (e) { console.log("[1] FAIL:", e.message); }

// proposed fix
const tok = await t.AutoTokenizer.from_pretrained(MODEL);
const textModel = await t.CLIPTextModelWithProjection.from_pretrained(MODEL, { dtype: "q8" });
const { text_embeds } = await textModel(tok(["a docker compose diagram"], { padding: true, truncation: true }));
console.log("[3] dims =", text_embeds.dims);
```

Output on Node 24.19.0 / `@huggingface/transformers` 4.2.0:

```console
[1] pipeline('feature-extraction', CLIP)  <- current code
    FAIL: An error occurred during model execution: "Missing the following inputs: pixel_values.

[2] pipeline('image-feature-extraction', CLIP)  <- image branch
    OK: image pipeline constructed

[3] AutoTokenizer + CLIPTextModelWithProjection  <- proposed fix
    OK, dims = [ 1, 512 ] -> matches image dim 512: true
```

## Suggested fix

Use the projection head for text instead of the generic pipeline, so both towers land in the same 512-d joint space:

```js
async getTextExtractor() {
  if (this.textExtractor) return this.textExtractor;
  const t = await loadTransformers();
  const tokenizer = await t.AutoTokenizer.from_pretrained(this.modelId);
  const model = await t.CLIPTextModelWithProjection.from_pretrained(this.modelId, { dtype: "q8" });
  this.textExtractor = async (texts) => {
    const { text_embeds } = await model(tokenizer(texts, { padding: true, truncation: true }));
    return text_embeds;
  };
  return this.textExtractor;
}
```

`embedBatch()` then normalizes as it does today. Verified above that `text_embeds` is `[n, 512]`, matching `dimensions = 512` and the image vectors already written by `mem::vision-embed`, so existing stored embeddings stay comparable — no re-embed needed.

Worth noting `pooling: "mean"` over CLIP's last hidden state would *not* be equivalent even if it ran: CLIP's contrastive space is defined by the projection head plus EOS-token pooling, so mean-pooled hidden states would embed into a different space and silently return meaningless cosine scores against the image vectors. The projection head avoids that.

## Environment

| | |
|---|---|
| `@agentmemory/agentmemory` | 0.9.29 |
| `@huggingface/transformers` | 4.2.0 |
| engine | `iiidev/iii:0.11.2` (Docker), worker on host |
| Node | v24.19.0 |
| OS | Windows 11 |
| flags | `AGENTMEMORY_IMAGE_EMBEDDINGS=true`, `EMBEDDING_PROVIDER=local` |

Happy to send a PR with the `CLIPTextModelWithProjection` change if the approach looks right.

anhtahaylove:
Fixes #1249.

`ClipEmbeddingProvider.getTextExtractor()` builds the text encoder from the generic `feature-extraction` pipeline. On a CLIP repo that instantiates the **full dual-encoder** `CLIPModel`, whose forward pass requires `input_ids` *and* `pixel_values`. The text-only call supplies just `input_ids`, so ONNX rejects it and every text entry point into the provider fails:

```console
$ curl -s -X POST -H 'Content-Type: application/json' \
    -d '{"queryText":"docker compose diagram","limit":3}' \
    http://127.0.0.1:3111/agentmemory/vision-search
{"error":"query embed failed: An error occurred during model execution: \"Missing the following inputs: pixel_values.","success":false}
```

Image-to-image search on the same store works, because `image-feature-extraction` maps to the vision tower alone — so the feature looks half-broken rather than obviously broken.

## What this does

Loads the text tower with its projection head (`AutoTokenizer` + `CLIPTextModelWithProjection`) instead of the generic pipeline. That is the encoder whose 512-d output shares the contrastive space with the image embeddings already written by `mem::vision-embed`, so **stored vectors stay comparable and no re-embedding is needed**.

Worth recording why the obvious smaller fix is wrong: `pooling: "mean"` over the last hidden state would not be equivalent even if it ran. CLIP's joint space is defined by the projection head, so mean-pooled hidden states embed into a *different* space and would return meaningless cosine scores against the image vectors — silently, with no error.

## How to verify

Against `Xenova/clip-vit-base-patch32` (dtype `q8`), embedding text and one image fixture through the patched provider:

```console
provider: clip dims: 512

[text] embedBatch -> 3 vectors, dim = 512
[text] L2 norm = 1.000000 (want ~1.0)

[image] embedImage -> dim = 512 | dims match: true

[cross-modal cosine] text vs that image:
   0.3229  "a red square"
   0.2539  "a blue circle"
   0.2364  "a docker compose diagram"
```

The fixture is a red square, and it ranks the three captions in the right order — the text and image vectors are in the same space, not merely the same length.

## Tests

`test/clip-embedding-provider.test.ts` mocked `pipeline("feature-extraction")`, so it asserted exactly the call that cannot work at runtime. Updated to mock the text tower, plus a regression guard:

```js
// Regression guard for #1249: the generic feature-extraction pipeline
// instantiates the full dual-encoder and demands pixel_values.
expect(pipeline).not.toHaveBeenCalledWith(
  "feature-extraction", expect.anything(), expect.anything(),
);
```

The batch test now also exercises a real 2-item batch, which the previous fixed-length mock would have passed regardless.

```console
$ npx vitest run test/clip-embedding-provider.test.ts
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

`npm run build` is clean. On the full suite this branch is **30 failed | 1665 passed**, identical to `main` at e04ba88 on the same machine — the pre-existing failures (`cli-lifecycle-safety`, `obsidian-export`, `compress-file`, and friends) are unrelated to this change and are unaffected by it.

Environment: Node v24.19.0, `@huggingface/transformers` 4.2.0, Windows 11.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLIP text embedding generation for more reliable text-only processing.
  * Text embeddings are now consistently normalized and returned in a predictable format.
  * Improved support for batched text inputs, including uneven batch sizes.
  * Enhanced compatibility between text and image embeddings in the shared vector space.
  * Improved handling of custom model configurations and model-loading errors.
  * Reused loaded text-embedding resources for more efficient repeated requests.

### DIFF
--- a/src/providers/embedding/clip.ts
+++ b/src/providers/embedding/clip.ts
@@ -7,13 +7,14 @@ type ClipPipeline = (
   input: string[] | RawImage | RawImage[],
   options?: { pooling?: string; normalize?: boolean },
 ) => Promise<{ tolist: () => number[][]; data: Float32Array }>;
+type ClipTextEncoder = (texts: string[]) => Promise<Float32Array[]>;
 
 const DEFAULT_MODEL = "Xenova/clip-vit-base-patch32";
 
 export class ClipEmbeddingProvider implements EmbeddingProvider {
   readonly name = "clip";
   readonly dimensions = 512;
-  private textExtractor: ClipPipeline | null = null;
+  private textExtractor: ClipTextEncoder | null = null;
   private imageExtractor: ClipPipeline | null = null;
   private readonly modelId: string;
 
@@ -27,9 +28,8 @@ export class ClipEmbeddingProvider implements EmbeddingProvider {
   }
 
   async embedBatch(texts: string[]): Promise<Float32Array[]> {
-    const extractor = await this.getTextExtractor();
-    const output = await extractor(texts, { pooling: "mean", normalize: true });
-    return output.tolist().map((v) => new Float32Array(v));
+    const encode = await this.getTextExtractor();
+    return encode(texts);
   }
 
   async embedImage(src: string): Promise<Float32Array> {
@@ -41,10 +41,24 @@ export class ClipEmbeddingProvider implements EmbeddingProvider {
     return normalize(vec);
   }
 
-  private async getTextExtractor(): Promise<ClipPipeline> {
+  private async getTextExtractor(): Promise<ClipTextEncoder> {
     if (this.textExtractor) return this.textExtractor;
     const t = await loadTransformers();
-    this.textExtractor = (await t.pipeline("feature-extraction", this.modelId, { dtype: "q8" })) as ClipPipeline;
+    // CLIP repos expose a dual-encoder. The generic "feature-extraction"
+    // pipeline instantiates the full CLIPModel, whose forward pass requires
+    // both input_ids and pixel_values, so a text-only call fails with
+    // "Missing the following inputs: pixel_values". Load the text tower with
+    // its projection head instead: that is the encoder whose output shares
+    // the 512-d contrastive space with the image embeddings.
+    const tokenizer = await t.AutoTokenizer.from_pretrained(this.modelId);
+    const model = await t.CLIPTextModelWithProjection.from_pretrained(this.modelId, {
+      dtype: "q8",
+    });
+    this.textExtractor = async (texts: string[]) => {
+      const inputs = tokenizer(texts, { padding: true, truncation: true });
+      const { text_embeds } = await model(inputs);
+      return text_embeds.tolist().map((v: number[]) => normalize(new Float32Array(v)));
+    };
     return this.textExtractor;
   }
 
--- a/src/providers/embedding/clip.ts
+++ b/src/providers/embedding/clip.ts
@@ -44,12 +44,9 @@ export class ClipEmbeddingProvider implements EmbeddingProvider {
   private async getTextExtractor(): Promise<ClipTextEncoder> {
     if (this.textExtractor) return this.textExtractor;
     const t = await loadTransformers();
-    // CLIP repos expose a dual-encoder. The generic "feature-extraction"
-    // pipeline instantiates the full CLIPModel, whose forward pass requires
-    // both input_ids and pixel_values, so a text-only call fails with
-    // "Missing the following inputs: pixel_values". Load the text tower with
-    // its projection head instead: that is the encoder whose output shares
-    // the 512-d contrastive space with the image embeddings.
+    // The generic "feature-extraction" pipeline loads the full dual-encoder,
+    // which demands pixel_values and breaks text-only calls (#1249). The
+    // projection head is also what puts text in the image vectors' 512-d space.
     const tokenizer = await t.AutoTokenizer.from_pretrained(this.modelId);
     const model = await t.CLIPTextModelWithProjection.from_pretrained(this.modelId, {
       dtype: "q8",
--- a/test/clip-embedding-provider.test.ts
+++ b/test/clip-embedding-provider.test.ts
@@ -119,4 +119,83 @@ describe("ClipEmbeddingProvider (with loaded pipeline)", () => {
       dtype: "q8",
     });
   });
+
+  it("caches the text tower across calls instead of reloading it", async () => {
+    const { fromPretrainedText, fromPretrainedTokenizer } = mockSuccessModule();
+    const { ClipEmbeddingProvider: Fresh } = await import(
+      "../src/providers/embedding/clip.js"
+    );
+    const provider = new Fresh();
+    await provider.embed("hello");
+    await provider.embed("world");
+    await provider.embedBatch(["a", "b"]);
+
+    expect(fromPretrainedTokenizer).toHaveBeenCalledTimes(1);
+    expect(fromPretrainedText).toHaveBeenCalledTimes(1);
+  });
+
+  it("tokenizes with padding so a batch of uneven texts stays one tensor", async () => {
+    const { tokenizer } = mockSuccessModule();
+    const { ClipEmbeddingProvider: Fresh } = await import(
+      "../src/providers/embedding/clip.js"
+    );
+    await new Fresh().embedBatch(["hi", "a much longer caption"]);
+
+    expect(tokenizer).toHaveBeenCalledWith(
+      ["hi", "a much longer caption"],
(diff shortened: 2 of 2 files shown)


anhtahaylove:
Correcting one line in my PR description: I wrote "`npm run build` is clean." That is wrong, and I should have checked the exit code instead of eyeballing the log.

`npm run build` exits **1** on Windows — on this branch *and* on clean `main` at e04ba88, identically. The compile itself is fine; the failure is in the asset-copy tail of the script:

```json
"build": "tsdown && (cp iii-config.yaml dist/ 2>/dev/null || true) && ... && mkdir -p dist/viewer && cp src/viewer/index.html dist/viewer/ && ..."
```

npm runs scripts through `cmd.exe` on Windows, which has no `cp` or `mkdir -p`, so the tail dies with "The system cannot find the path specified." / "The syntax of the command is incorrect." and `dist/viewer/` is never created. The `|| true` guards cover the first four copies but not `mkdir -p dist/viewer` or the two `cp` calls after it.

What actually holds for this PR:

```console
$ npx tsdown
exit 0

$ npx tsc --noEmit -p tsconfig.json
30 pre-existing errors, none in src/providers/embedding/clip.ts
  (api.ts, events.ts, ... — identical on main)
```

So the TypeScript half compiles clean and my change adds no type errors; the exit=1 is a pre-existing Windows portability issue in the build script, unrelated to this PR. Happy to send a separate PR swapping that tail for something cross-platform (`cpy-cli`/`shx`, or a small Node script) if that is wanted — it did not seem right to fold it into a one-file provider fix.

anhtahaylove:
Partially addressed in 89d19f2.

The guideline is "no code comments explaining WHAT — use clear naming instead" (AGENTS.md:102), and the middle of that block was indeed describing mechanics. Trimmed it from six lines to three.

I kept the *why*, because naming cannot carry it: nothing in `AutoTokenizer` / `CLIPTextModelWithProjection` tells the next reader that the obvious alternative — `pipeline("feature-extraction", ...)`, which is what this code did until this PR — is broken rather than merely different. Without that note the change looks like a stylistic preference and is a plausible "simplification" for someone to revert straight back into the bug.

That reading matches how the repo uses comments elsewhere: `src/cli.ts:525` explains why iii is installed privately, `src/cli/splash.ts:24` explains why colour 208 over 209, `src/hooks/antigravity-bridge.ts:6` explains why the bridge exists. All rationale, none of them narrating what the code does.

anhtahaylove:
@rohitg00 — this one is ready for review when you have a moment.

Fixes #1249. `CLIPEmbeddingProvider.embedText()` fed raw text through the image processor, so every text embedding threw `pixel_values` and cross-modal search was dead on arrival. The fix routes text through `AutoTokenizer` + `CLIPTextModelWithProjection` (the text tower), which is what CLIP expects.

Verified on a real 24-image store after the fix: text->image retrieval returns sane scores (0.3243 for "a red square" against the matching image), image->image self-match is 1.0000, and the MiniLM text index is untouched.

Tests: 5/5 in `test/clip-embedding-provider.test.ts`. CodeRabbit's note on the comment wording is addressed in 89d19f2. The Vercel check needs a Team member to authorize the preview deploy for outside contributors — nothing to do with the code.

PR #1256, opened 2026-08-26
fix(state): reclaim index generations the manifest can no longer name

### LINKED ISSUE #1115: Orphaned index generations accumulate when the previous manifest read fails — store grows until the server pegs a core while idle
**Package:** `@agentmemory/agentmemory` **0.9.28**
**iii runtime:** 0.11.2
**Host:** macOS 26.5.2, arm64, Node v24.16.0, launchd-managed (not Docker)
**Store:** file-based KV, `~/data/state_store.db`, ~3,000 observations

## Summary

`IndexPersistence.save()` cleans up the *previous* generation's shards only when it successfully read the previous manifest. When that read fails — which happens exactly when the store is under load — the cleanup is skipped entirely and **that generation's shards are orphaned forever**: every later `save()` only ever looks at *its own* `previous`, so nothing ever revisits older orphans.

Orphans are monotonically accumulating, and they feed back into the failure that creates them.

In my install this reached **8 index generations / 140 MB of shards**, of which only 2 generations (37 MB) were live. The server then sat at **110–150% CPU while completely idle** with every HTTP route timing out.

## Evidence

Storage breakdown before cleanup (338 MB total):

| Key prefix | Size | Kind |
|---|---|---|
| `mem:index:*` | **140 MB, 77 files, 8 generations** | derived (only 2 gens live) |
| `mem:graph:*` | 85 MB | derived |
| `mem:audit` | 42 MB | log |
| `mem:obs:*` (2,888 observations) | 66 MB | **real data** |

So **79% of the store was derived data**, most of it unreachable.

The manifest names the live generations explicitly:

```json
{"data:manifest":    {"generation": "idx_mrzxkql0_d41f7623ee4f", "shards": [...]},
 "vectors:manifest": {"generation": "idx_mrzxkrwf_8fe2c54a3dc8", "shards": [...]}}
```

The other six `idx_*` generations were referenced by nothing.

Effect of deleting the orphans + graph + audit (observations untouched):

| Metric | Before | After |
|---|---|---|
| Idle CPU (no requests in flight) | **110–150%** | **0–3%** |
| Wrapper RSS | 1.4–2.7 GB, oscillating | **93–250 MB** |
| `GET /agentmemory/health` | 8–25 s, often timeout | **6 ms** |
| `POST /agentmemory/search` | 20 s timeout | **14 ms** |
| `POST /agentmemory/smart-search` | **44.5 s** | **0.82 s** |
| Store size | 338 MB | 109 MB |

The server has now been up **45+ hours** with no restarts and no degradation.

## Root cause

`dist/index.mjs`, end of `IndexPersistence.save()`:

```js
await this.deleteKey(KV.bm25Index, legacyKey, "legacy_cleanup");
if (previous?.v === 1 && Array.isArray(previous.shards)) {
  const currentShardIds = new Set(shards.map((shard) => `${shard.scope}\0${shard.key}`));
  for (const shard of previous.shards) {
    if (currentShardIds.has(`${shard.scope}\0${shard.key}`)) continue;
    await this.deleteShards([shard], "previous_generation_cleanup");
  }
}
```

`previous` comes from reading the manifest at the start of `save()`. That read fails under load — from my logs:

```
[agentmemory] warn index persistence: BM25 manifest read failed
  {"scope":"mem:index:bm25","key":"data:manifest","message":"Invocation timeout after 180000ms: state::get"}
[agentmemory] warn index persistence: vector manifest read failed
  {"scope":"mem:index:bm25","key":"vectors:manifest","message":"Invocation timeout after 180000ms: state::get"}
```

When that happens `previous` is nullish, the `if` is skipped, and the shards written by the *preceding* generation are never deleted by anyone. There is no reconciliation pass that would find them later.

I found **zero** `previous_generation_cleanup` audit records against 3 logged manifest-read failures, which is consistent.

## Why it compounds

1. Store grows → `state::get` gets slower
2. Manifest read hits the 180 s invocation timeout
3. Cleanup skipped → one more generation orphaned permanently
4. Store grows further → go to 1

Each failure is permanent, so the floor only ever rises. This also explains why the symptom appears suddenly after weeks of fine operation: it is not linear, it is ratcheting.

A restart makes it *worse*, because a fresh `save()` mints yet another generation — so debugging by restarting is actively counterproductive. During one incident I restarted eight times and ended up with eight generations.

## Reproduction

Any deployment where a manifest read times out at least once. To force it:

1. Grow the store until `state::get` on the manifest key approaches the invocation timeout (~3,000 observations with vectors + graph did it here).
2. Trigger an index save.
3. Observe `manifest read failed` in the log.
4. `ls ~/data/state_store.db | grep 'mem%3Aindex%3A'` — the previous generation's shards are still present and are not referenced by `mem%3Aindex%3Abm25.bin`.

## Suggested fixes

1. **Reconcile instead of relying on `previous`.** When the manifest read fails, don't silently skip. Enumerate the `mem:index:bm25` scope and delete any `idx_*` shard whose generation is not named by the freshly published manifest. The manifest is the authority and it is written *before* this step, so a scan-based sweep is safe and self-correcting.
2. **Add a startup reconciliation pass.** One sweep at boot would bound the damage to a single generation regardless of how many past failures occurred.
3. **Reconsider the 180 s invocation timeout for manifest reads.** Three minutes of a blocked read on the single-threaded runtime is itself an availability problem; a short timeout plus explicit retry would fail faster and more visibly.
4. **Surface the skip.** `manifest read failed` is logged at `warn`, but the consequence (cleanup skipped, generation orphaned) isn't stated. Making it explicit would have saved a lot of diagnosis time.

Happy to open a PR for (1) + (2) if the approach sounds right.

## Workaround for anyone hitting this

Stop the server, back the store up, then keep only the generations the manifest names:

```bash
# 1. read the live generations
python3 -c "
import re;print('\n'.join(sorted(set(re.findall(r'idx_[a-z0-9]+_[a-f0-9]+',
  open('~/data/state_store.db/mem%3Aindex%3Abm25.bin',encoding='utf8',errors='replace').read())))))"
# 2. delete every mem%3Aindex%3A* shard whose generation is not in that list
```

⚠️ Do **not** delete `mem%3Aindex%3A*` wholesale — the index is **not** rebuilt automatically. I tried; the server came up healthy and fast, and every query returned 0 results while all 2,888 observations sat there intact. Keeping the manifest plus the two live generations restored search completely.

`mem:graph:*` and `mem:audit` *are* regenerated and can be dropped safely (dropping the 85 MB graph is what took `smart-search` from 44.5 s to 0.82 s).

## Related

- #309 (closed) — in-memory BM25/graph architecture; this is a consequence of the sharded-persistence design that came out of it
- #762 (closed) — introduced the sharded manifest/generation scheme this issue is about
- #1093 (open) — idle CPU growth over ~40 h. Looks similar from the outside but is a different layer: that one is `iii` worker threads busy-waiting and clears on restart, whereas this one is the Node wrapper and *survives* restarts because the orphans are on disk.

inix-x:
Closes #1115.

## The bug

`IndexPersistence.saveShardedIndex` writes each new index generation's shards, publishes a manifest naming it, then deletes the previous generation. That last step only runs when the previous manifest read succeeded:

```ts
const previous = await this.kv
  .get<IndexShardManifest>(KV.bm25Index, manifestKey)
  .catch(() => null);
```

That read is the first thing the function does, and it fails under exactly the load that makes saves slow. `previous` becomes `null`, the cleanup guard is skipped, and the whole preceding generation is stranded. Nothing ever revisits it: a later save only inspects its own predecessor, `loadShardedData` reads only the shards the current manifest lists, and `StateKV` can list keys within a scope but cannot enumerate scopes by prefix. An unrecorded generation is unfindable.

Two more paths reach the same state. The manifest `catch` block ends in a `throw`, so a write that threw after committing skips the cleanup below it. And a kill between the first shard write and the cleanup leaves a generation nothing recorded.

## Evidence

From a deployed store, counted rather than inferred:

| | files | MB |
|---|---|---|
| vectors, live | 140 | 266.6 |
| bm25, live | 78 | 164.4 |
| **bm25, 8 dead generations** | **162** | **339.4** |
| observations (the actual data) | 1314 | 35.4 |

339.4 MB of orphans, 33.5% of a 1013 MB store, against 35 MB of real data. Generation ids are `Date.now().toString(36)`, so they decode to timestamps: the oldest orphan was 13 days old, and every one of them lands on a restart or an incident.

On a container with a memory cap this escalates. The store grows, the engine's memory floor rises with it, and the engine is eventually killed. Each kill orphans another generation, so the next kill arrives sooner. Observed cadence went from about 9 hours to under 3 minutes.

This matches the two reports already in #1115 (macOS/launchd and Linux/systemd), which independently measured 79% derived data on their own stores.

## The fix

Record each generation in a ledger stored beside its manifest, written **before** the first shard write, and reclaim from it after publish and again on load.

**Saves are serialized.** `scheduleSave` fires `save()` unawaited from a timer, `flushIndexSave` awaits `save()` on every delete path, and `stop()` clears the timer without awaiting a save already running, so saves overlap by construction. Two of them read-modify-writing one ledger drop each other's entries, and the one that publishes second reclaims the shards the first is still writing. The queue never coalesces, because `flushIndexSave` awaits `save()` to make a delete durable and returning an in-flight promise that started before the delete would report success for a snapshot without it. The load-time reclaim joins the same queue.

**Degradation is deliberate.** An unreadable ledger skips that cycle's tracking and reclaim rather than failing the save, and falls back to the pre-ledger cleanup, so it stays at parity with current behaviour instead of regressing. A ledger whose shape is not recognised is left untouched. Reclaim keeps malformed entries rather than iterating them, so a GC step can never fail a load, and it emits one audit row per sweep rather than one per shard, per the policy in `src/functions/audit.ts`.

Each index also saves independently now. One `try` around both let a BM25 failure stop the vector index persisting.

That last clause used to end "and a lost vector index is never rebuilt because both rebuild triggers key on the BM25 size." That is no longer true as of #1258, which adds a vector-aware rebuild trigger. See the composition note below.

## Known limitations

Three things a reviewer should weigh, stated up front rather than discovered later.

1. **The ledger cannot discover generations orphaned before it existed.** It stops further accumulation and reclaims everything from the upgrade forward, but the 339 MB already on disk needs a one-time reclaim. Nothing can enumerate shard scopes, so that cannot be done from inside this abstraction.

2. **`state::delete` on a missing key is benign. Verified, so this is no longer a limitation.** It is load-bearing in two places, the rollback path and the reclaim retry, so it was worth settling rather than assuming. Four independent confirmations at the pinned engine v0.11.2: `engine/src/workers/state/state.rs:290-329` returns `Success(None)` for an absent key, and its only failure paths are store-level and equally reachable on a *present* key; the configured `kv` / `file_based` adapter (`adapters/kv_store.rs:50-59`) wraps an infallible store, so even those branches are unreachable for this deployment; the engine's own unit test `test_delete_nonexistent_key` asserts it; and a probe driven through `iii-sdk` against the real 0.11.2 binary resolved every case, including the exact rollback double-delete shape, with no rejection and no log line at any level.

3. **The gc ledger's `generations` array is unbounded while every shard delete keeps failing.** Each debounce appends a generation carrying its full shard list, and `reclaimGenerations` only rewrites the ledger when something was actually reclaimed, so nothing drains it until one delete succeeds. Reaching that state takes a total delete outage. A length cap is more than a one-line change and is not attempted here.

   Measured rather than argued: under a differential brownout where the 2 MB shard `state::set` times out while small ledger writes keep working, 20 rollback cycles leave the current code flat at 0 ledger entries, and a variant with the rollback bookkeeping removed at 20, growing linearly. At `DEBOUNCE_MS = 5000` that is 12 entries per minute on the save hot path.

## Composes with #1258

Both PRs rewrite `deleteKey` to return a boolean and both restructure `save()` in `src/state/index-persistence.ts`, so they conflict textually and semantically. #1258 splits `save()` with vectors first and calls `saveVectorBuckets`, where this PR still calls `saveVectorIndex(this.vector.serialize())`.

**#1258 should land first.** It replaces the vector save path this PR reasons about, and it closes the "a lost vector index is never rebuilt" hole quoted above. Landing this one first means #1258 rewrites the same functions again.

## Testing

Thirteen new tests in `test/index-persistence.test.ts`, each verified to fail without its corresponding fix (mutation-checked, not assumed):

- previous generation reclaimed when the previous-manifest read fails
- same for the vector path
- generation stranded by a failed cleanup reclaimed on the next load
- published manifest stays whole when two saves overlap
- index still persists when the gc ledger is unreadable
- index still reclaims when the ledger holds a malformed entry
- unrecognised ledger is never overwritten
- fallback cleanup runs when the ledger is unusable
- ledger entry survives a failed rollback delete

`npm test`: 1721 passing. `tsc --noEmit` is unchanged from base (same 30 pre-existing errors).

Three subprocess tests failed on one full run and were confirmed as the suite's known non-determinism rather than assumed: the failure set *changed between two runs of an identical tree*, every failure was a 5000 ms wall-clock timeout rather than an assertion, and all pass in isolation. A static code change cannot produce different failures across two runs of the same tree.

Measured against base on the two regression-prone paths: with the gc ledger unreadable across 10 saves, both base and this branch leave 0 orphans. With 5 rollback cycles where every delete fails, base leaves 10 orphans permanently while this branch drains to 0 after one clean save.

### DIFF
--- a/src/state/index-persistence.ts
+++ b/src/state/index-persistence.ts
@@ -90,15 +90,7 @@ export class IndexPersistence {
   private timer: ReturnType<typeof setTimeout> | null = null;
   private lastFailureLogAt = new Map<string, number>();
   private queue: Promise<void> = Promise.resolve();
-  /**
-   * A save that is queued but has not started running yet.
-   *
-   * It will serialise the index as it stands at the moment it runs, so it
-   * already covers every caller that arrives before then. Handing them this
-   * promise instead of queueing another identical save is what keeps a delete
-   * from waiting out an unbounded line of saves ahead of it.
-   */
-  private pendingSave: Promise<void> | null = null;
+  private unstartedSave: Promise<void> | null = null;
 
   constructor(
     private kv: StateKV,
@@ -123,20 +115,13 @@ export class IndexPersistence {
       clearTimeout(this.timer);
       this.timer = null;
     }
-    // Coalesce onto a save that is queued but has not started. It has not taken
-    // its snapshot yet, so it will include whatever this caller just changed.
-    // Only a save that is ALREADY RUNNING must not be handed back: its snapshot
-    // predates this caller's mutation, and reporting success for it would tell
-    // a delete it was durable when it was not.
-    if (this.pendingSave) return this.pendingSave;
+    if (this.unstartedSave) return this.unstartedSave;
 
     const pending = this.enqueue(() => {
-      // Running now. Its snapshot is fixed from here, so the next caller needs
-      // a save of its own.
-      if (this.pendingSave === pending) this.pendingSave = null;
+      if (this.unstartedSave === pending) this.unstartedSave = null;
       return this.runSave();
     });
-    this.pendingSave = pending;
+    this.unstartedSave = pending;
     return pending;
   }
 
@@ -524,11 +509,6 @@ export class IndexPersistence {
       });
     }
 
-    // ponytail: unbounded while EVERY delete keeps failing — each debounce
-    // appends a generation carrying all its shard descriptors, and
-    // reclaimGenerations only rewrites when something was actually reclaimed.
-    // Add a length cap if that ever shows up in the wild; a cap is not one
-    // line, and this needs a total delete outage to reach.
     await this.kv.set<IndexGcLedger>(
       KV.bm25Index,
       this.gcKey(manifestKey),
--- a/src/state/index-persistence.ts
+++ b/src/state/index-persistence.ts
@@ -90,6 +90,15 @@ export class IndexPersistence {
   private timer: ReturnType<typeof setTimeout> | null = null;
   private lastFailureLogAt = new Map<string, number>();
   private queue: Promise<void> = Promise.resolve();
+  /**
+   * A save that is queued but has not started running yet.
+   *
+   * It will serialise the index as it stands at the moment it runs, so it
+   * already covers every caller that arrives before then. Handing them this
+   * promise instead of queueing another identical save is what keeps a delete
+   * from waiting out an unbounded line of saves ahead of it.
+   */
+  private pendingSave: Promise<void> | null = null;
 
   constructor(
     private kv: StateKV,
@@ -114,7 +123,21 @@ export class IndexPersistence {
       clearTimeout(this.timer);
       this.timer = null;
     }
-    return this.enqueue(() => this.runSave());
+    // Coalesce onto a save that is queued but has not started. It has not taken
+    // its snapshot yet, so it will include whatever this caller just changed.
+    // Only a save that is ALREADY RUNNING must not be handed back: its snapshot
+    // predates this caller's mutation, and reporting success for it would tell
+    // a delete it was durable when it was not.
+    if (this.pendingSave) return this.pendingSave;
+
+    const pending = this.enqueue(() => {
+      // Running now. Its snapshot is fixed from here, so the next caller needs
+      // a save of its own.
+      if (this.pendingSave === pending) this.pendingSave = null;
+      return this.runSave();
+    });
+    this.pendingSave = pending;
+    return pending;
   }
 
   /**
--- a/src/state/index-persistence.ts
+++ b/src/state/index-persistence.ts
@@ -24,6 +24,27 @@ type IndexShardManifest = {
   chars: number;
 };
 
+// Suffix for the reclaim ledger that sits beside each manifest.
+const GC_LEDGER_SUFFIX = ":gc";
+
+// Stands in for a manifest written before generations were recorded, whose
+// `generation` field is absent. Cannot collide with createIndexGeneration().
+const PRE_LEDGER_GENERATION = "pre-ledger";
+
+// Every generation whose shards may still be on disk, live one included. The
+// manifest alone cannot answer that: it names only the generation that is
+// current, so a generation stranded by a failed read, a throw after commit, or
+// a kill mid-cleanup becomes unreachable the moment the next manifest replaces
+// it. Nothing else enumerates shard scopes — StateKV lists keys within a scope,
+// not scopes by prefix — so what is not recorded here can never be found again.
+type IndexGcLedger = {
+  v: 1;
+  generations: Array<{
+    generation: string;
+    shards: Array<{ scope: string; key: string }>;
+  }>;
+};
+
 type IndexPersistenceOptions = {
   shardChars?: number;
   createGeneration?: () => string;
@@ -67,7 +88,8 @@ function isValidShardDescriptor(
 
 export class IndexPersistence {
   private timer: ReturnType<typeof setTimeout> | null = null;
-  private lastFailureLogAt = 0;
+  private lastFailureLogAt = new Map<string, number>();
+  private queue: Promise<void> = Promise.resolve();
 
   constructor(
     private kv: StateKV,
@@ -83,7 +105,7 @@ export class IndexPersistence {
     // under sustained iii-engine write timeouts (issue #204). Funnel
     // rejections through logFailure() instead.
     this.timer = setTimeout(() => {
-      this.save().catch((err) => this.logFailure(err));
+      this.save().catch((err) => this.logFailure("index", err));
     }, DEBOUNCE_MS);
   }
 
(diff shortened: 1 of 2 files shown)


inix-x:
Follow-up commit: the save queue this PR adds caused a regression I need to flag rather than let a reviewer find.

`flushIndexSave` is awaited on five delete paths (`governance.ts:33,139`, `remember.ts:321`, `retention.ts:392`, `auto-forget.ts:194`). Once saves are serialised rather than overlapping, a delete waits for every save ahead of it. That is fine when a save takes a second. When saves are timing out at 180s it turns a delete into an unbounded wait and the request dies.

Measured on a deployed instance running this change: no index save completed in a 29 minute window, and 5xx rates rose sharply from the point this commit was deployed.

My original note in this PR — "the cost is real: a delete-path flush waits out the save ahead of it" — was correct and I underestimated it.

The reasoning against coalescing was right about one case and too broad about the other. A save that is **already running** must never be handed back: its snapshot predates the caller's mutation, so reporting success would tell a delete it was durable when it was not. But a save that is **queued and has not started** has not taken its snapshot yet, so it will serialise whatever the caller just changed. Those callers can safely share it.

`c6851e7` tracks the queued-but-not-started save and hands it to arriving callers, clearing it the moment it starts running so the next caller queues its own. Six concurrent delete flushes now cost at most two saves instead of six, and the durability guarantee is unchanged.

The test was confirmed to fail against the unfixed code first: it measured 6 saves for 6 callers, and now measures 2.

PR #1258, opened 2026-08-26
perf(state): write only the vector buckets that changed

inix-x:
## The bug

`IndexPersistence` serialises the whole index to one string and cuts it into shards by **character offset**:

```ts
for (let offset = 0; offset < serialized.length; offset += chunkChars) {
  const chunk = serialized.slice(offset, offset + chunkChars);
```

A single new observation shifts every downstream byte, so every shard differs and every shard is rewritten. That is O(entire corpus) per change, and it compounds: more memories means a bigger index means bigger rewrites.

## Evidence

Read off a deployed store rather than inferred. Generation ids are `Date.now().toString(36)`, so they decode to the moment that generation was published:

| | |
|---|---|
| **vector index, last completed save** | **`2026-08-26T06:21:43Z`** |
| bm25 index, last completed save | `2026-08-26T11:03:31Z` |
| time of reading | `2026-08-26T11:13:57Z` |
| **elapsed with the vector index unable to persist** | **4h 52m** |
| `mem:sessions` mtime | `11:12` (actively written) |
| shards | 78 bm25 + 140 vector |
| store size vs real data | 753 MB over ~35 MB of observations |

The store is written continuously and BM25 does get through, while the vector index has not persisted for nearly five hours. That split is the whole point, and it follows directly from `runSave`:

```ts
await this.saveBm25Index(this.bm25.serialize());
if (this.vector) {
  await this.saveVectorIndex(this.vector.serialize());
}
```

BM25 goes first at ~164 MB and completes. The vector index goes second at ~266 MB and does not. The failure is not random: it lands on the larger of the two writes, every time. This is not a stalled process — it is one specific write that has outgrown what a full rewrite can push through.

With no orphaned generations present, that 753 MB is the *live* index. Write amplification is the whole story of the size, not garbage.

It is also not slow storage. A 64 MB `dd` with `conv=fsync` on that volume runs at **65.5 MB/s**, so the disk would absorb the entire 431 MB index in under seven seconds — against an invocation timeout of 180. Whatever the write costs, it is not being spent on the device.

The consequence that matters most: `flushIndexSave` is awaited on every delete path (`governance.ts:33,139`, `retention.ts:392`, `auto-forget.ts:194`, `remember.ts:321`) precisely to make a delete durable. While no save completes, **no delete is durable** — forgotten memories return on restart.

## The change

Partition the vector index into a fixed number of buckets by `hash(obsId) % N`. Each save serialises a bucket, hashes it, and writes it only when the hash differs from the one the manifest recorded.

Bucket keys are deterministic and reused in place, so there is no generation to strand and **orphaned shards are structurally impossible** — which is a different and stronger property than reclaiming them after the fact.

**Hashes live in the manifest, not in process memory.** A restart-heavy service with an in-memory cache would rewrite every bucket after every restart, which is exactly the cost being removed.

**Buckets are still chunked to `shardChars`.** Bucket size grows with the corpus and the engine rejects an oversized `state::set` payload. Chunking bounds the payload; bucketing bounds how much gets rewritten. They solve different problems and both are needed — there is an existing regression test for the payload bound.

**Nothing is deleted until the manifest naming its replacement is live.** Deleting first destroys data the still-current manifest points at when a publish fails, since the old manifest survives the failure.

**The manifest records the addressing scheme, not just the content.** This one is worth calling out, because it is the subtle failure of the whole approach and I got it wrong first: a content hash cannot see an addressing change. The bytes are identical, they just belong somewhere else now. So `layout`, `buckets`, and `chunkChars` are all compared before any write is skipped, and a mismatch forces a full rewrite plus a reclaim of every key the old manifest named.

Without that, changing the bucket count stranded every old bucket permanently — the exact orphan class this format was supposed to make impossible — and changing the chunk size skipped the write while recording a new chunk count, so load read the wrong number of keys and dropped the bucket as corrupt. `layout` is versioned separately from `v` on purpose: an older manifest must stay parseable for reclaim, and bumping `v` would make it unreadable and strand the very keys that need deleting.

## The trap in this design, and why keys carry a hash

The obvious way to build this is to overwrite each bucket in place under a stable key, publishing the manifest last. I built it that way first, and it is wrong in a way that is worth spelling out, because it looks safe.

A save that dies between the chunk writes and the manifest publish leaves a bucket that no longer matches the hash the **still-current** manifest recorded. Load drops it. Nothing restores it: rebuild is gated on the BM25 index being empty (`src/index.ts`, `src/functions/search.ts`), never on the vector index, so with BM25 populated no re-embed ever runs. The next save then serialises the index without those vectors and overwrites the bucket. **Silent, permanent loss** — and the generation swap this replaces lost nothing in the same situation, so it was a straight durability regression.

So chunk keys carry the bucket's content hash. The old bucket stays readable until the new manifest names the new one, a save that dies partway loses nothing, and reclaim stays exact because the old manifest still names the old keys.

Two consequences that are easy to miss and are covered by tests:

- **Never delete a key the new manifest names.** Old and new keys can legitimately collide: identical content re-chunked keeps the same bucket *and* hash, so chunk zero's key is unchanged even when the old entry listed more chunks.
- **Unfinished reclaim is carried in the manifest.** Deletes run after the publish, so dying partway would strand the remainder with nothing able to name them. A delete that fails stays on the list rather than being reported done.

Cross-bucket atomicity is still deliberately given up, and that part buys nothing today because the atomic save never completes.

**On overlapping saves.** `scheduleSave` fires `save()` unawaited from the debounce timer while five delete paths (`remember.ts:321`, `auto-forget.ts:194`, `governance.ts:33`, `governance.ts:139`, `retention.ts:392`) await it, so two saves can overlap on `main` today. Content-addressed keys make that far less dangerous than it would otherwise be: two saves write different content, so they write different keys and cannot clobber each other's bytes. Whichever publishes last wins, and every key its manifest names was written by that same save, so the published state is readable. The loser's keys leak rather than corrupting anything. That composes with the save serialisation in #1256, which removes the leak entirely.

## A reviewer found a second way to lose the same data, and was right

Review raised that changing `contentHash()` would make `loadVectorBuckets()` reject every existing bucket, and that bumping `VECTOR_LAYOUT` alone would not be enough because rejected buckets cannot be rewritten. I drafted a rejection. Tracing it end to end showed the stronger form of the claim is literally correct about this code:

`isVectorBucketManifest` does not check `layout`, so the manifest parses. Chunk keys derive from the **stored** hash, so the chunks are found. Every `contentHash(body) !== entry.hash` comparison then fails, `corrupt` reaches the entry count, and `loadVectorBuckets` returns an **empty `VectorIndex` rather than null** — so `load()` never falls through to the v1 reader. On the next save `serializeBuckets` yields nothing, the `prior` loop re-names every entry, and the early return fires. No rebuild runs, because `src/index.ts` keyed `needsRebuild` on the BM25 size and never on the vector index.

Bytes preserved, manifest preserved, vectors permanently unreadable, no automatic recovery. The reproduction rewrites a saved store under an alien hash and drives `save()` from the `load()` that rejected it. Against unmodified source it fails with `corrupt: 20 / total: 20` and a reopened index of 0.

Three commits close it:

- **`5a1edc7`** adds a precise recovery signal. Returning null was traced and recovers nothing: `load()` falls to `loadVectorData`, which fails `isVectorBucketManifest` and returns null anyway, and `src/index.ts` treats null and empty identically. Keying the rebuild on `vectorIndex.size === 0` would false-positive on a store whose vectors were never written, and `rebuildIndex` calls `idx.clear()`, so a false positive wipes a healthy BM25 index. The signal is therefore narrow on purpose: set only when a load read entries and rejected **all** of them. A failed *read* may succeed next boot and the existing preserve-and-wait policy already covers it. A failed *content check* fails identically forever.
- **`ad38fb6`** accepts a bucket hashed by any scheme this format has published, so a build one step behind can still read a store the newer one wrote. Sequenced before the hash change so no commit in history leaves a store recoverable only by a full re-embed.
- **`bf21c2d`** moves published hashes to sha256 and `VECTOR_LAYOUT` to 3.

**On sha256 rather than `fingerprintId`.** Review suggested `fingerprintId`. Measured, it is the wrong tool here: `vectorChunkKey` takes `hash.slice(0, 12)`, and `fingerprintId("vec", ...)` spends four of those twelve characters on a constant `vec_` prefix, leaving 8 hex characters. That is **32 bits against raw sha256's 48**, a 65536x reduction in exactly the content addressing that makes a torn write survivable. Plain sha256 clears the CWE-328 warning without paying that.

## The layout bump is not what carries this migration

Worth stating because it is the same class of error the finding was about, and mutation testing caught me assuming it.

Reverting `VECTOR_LAYOUT` from 3 back to 2 leaves **48 of 48 tests green**. A stored sha1 hash cannot equal a computed sha256 one, so the per-bucket comparison already forces the full rewrite on its own. The bump is kept for the constant's documented contract and as a manifest-readable marker of the scheme, not because it is load-bearing here.

It *was* load-bearing for the earlier chunk-key change described under Testing, where every hash matched on both sides and only the layout comparison could see the addressing move.

## Compatibility

v1 manifests still load, and the first save migrates and reclaims the shards they name. A v1 manifest names its own shards, so migration needs no enumeration — which matters, because `StateKV.list` returns values rather than keys and cannot be used to walk a shard scope without pulling every body into memory.

**One caveat worth stating plainly:** the migrating save is not cheap. With no previous v2 manifest, every bucket counts as changed, so that first save writes the whole index once — the same size as the save that is currently failing. The change only takes effect after one full vector save lands.

That is acceptable because it needs a single success rather than a sustained one, and saves clearly do land in quieter windows (BM25 completed at 11:03, and the vector index itself completed at 06:21). Every save after that one is incremental. But it does mean the fix is not in force the moment it deploys, and it should be verified by checking that `vectors:manifest` has flipped to `v: 2`, rather than assumed.

I considered publishing a partial v2 manifest to make the migration resumable and rejected it: it overwrites the v1 manifest, which both strands the v1 shards and leaves a partial index behind. All-or-nothing is the safer trade.

## Scope

BM25 is untouched. Its `inverted` map is term to docIds, so one new document dirties the bucket of every term it contains, and `totalDocLength` is a global scalar with no natural shard home. That is a genuinely different design problem and bundling it here would have made this unreviewable.

## What this does not claim

This fixes saves not completing, delete durability, and store growth. **It is not a fix for the engine being killed under memory pressure**, and I want to be explicit about that rather than overclaim: a previous change on this store reclaimed 339 MB (memory went 98% to 46% of cap) and the engine deaths continued unchanged. Engine growth is load-proportional and was still climbing during a window in which no save completed, so it is not proven to be driven by these writes.

## Testing

Twenty new tests in `test/index-persistence.test.ts`. Each was run against the unmodified source first and confirmed to fail, then mutation-checked against the finished code. The last seven mutations cover the recovery work added after review:

- removing the skip-unchanged guard kills 4 tests
- deleting stale buckets before the manifest publish kills 2
- removing load-time hash verification kills 1
- forcing a single chunk per bucket kills 3
- removing the v1 migration reclaim kills 1
- skipping on the content hash alone, ignoring layout, kills 1
- reclaiming only layout-matched buckets kills 1
- disabling tail-chunk reclaim kills 1
- reverting to in-place bucket keys kills 5
- clearing the reclaim list when a delete failed kills 1
- dropping the layout-version comparison kills 1
- skipping every write whenever the layout matches kills 3
- forcing the all-rejected signal off kills the recovery reproduction
- marking only missing entries incomplete kills it with the buckets destroyed
- narrowing the accepted-hash set to the published one alone kills the sha256 test
- making the hash comparison always succeed kills 2, including the pre-existing torn-write test
- reverting the published hash to sha1 kills 2
- dropping sha1 from the accepted set kills 2, including migration
- reverting `VECTOR_LAYOUT` 3 to 2 kills **nothing**, which is why the section above exists

That second-to-last one is worth calling out, because it caught a bug I had already shipped into this branch. Changing the chunk key format is an addressing change, and I made it without moving `VECTOR_LAYOUT`. On an upgrade the layout, bucket count, chunk size, and every content hash all matched, so every write was skipped and load then looked for hash-bearing keys the store did not have: 30 vectors to 0.

No test could see it, because every test in the file writes *and* reads through the same build, which keeps the suite self-consistent under any addressing change. Deleting the layout comparison outright left the whole suite green. There is now one cross-version fixture that stands in for a store written by an older layout, and it is the only thing in the file that pins that guard.

## Known limitations

Stated here rather than left in the diff to be found:

- **Recovery runs `rebuildIndex`, which clears BM25 first.** On the exact store this targets, healthy BM25 with rejected vectors, boot now wipes a fully restored BM25 index and rebuilds both over what `src/index.ts` calls hours. The right trade against permanent vector loss, but real and new.
- **`vectorLoadIncomplete` is sticky for the process.** If the corpus shrank between the bad save and the rebuild, uncovered buckets are re-named into the new manifest still carrying alien hashes. The next load then sees `corrupt > 0` but not all-corrupt, so no rebuild fires and those entries stay stale. The follow-up shape is to preserve on `missing` and reclaim on `corrupt`, which contradicts an existing design comment and is not attempted here.
- **The `MAX_BUCKET_CHUNKS` guard throws mid-loop**, so buckets written earlier in the same pass are left behind. If the corpus has not moved by the next successful save the identical content-addressed keys are reused; if it has, they are named by no manifest and leak. Same shape as a crash mid-save, which this format already accepts. Note it is unreachable at the shipped configuration: the only production constructor passes no options, so `shardChars` is always 2,000,000 and the throw needs a 20 GB single bucket.
- **The manifest itself is never chunked.** Both payload guards test `typeof data === "string"`, and the manifest goes through `kv.set` as an object, so it bypasses them. It measures ~18 KB at the default 256 buckets and grows linearly with `vectorBuckets` — comfortably under the frame cap today, but unbounded in principle.
- **The bucket-count clause of the layout comparison is not independently pinned.** Removing it alone keeps the suite green. That is defensible, since a bucket membership change also changes the body hash, but it means the test named for it is really pinning something else.
- **Overlapping `save()` has no re-entrancy guard on `main`.** `stop()` clears a pending timer but cannot cancel a save already awaiting a `state::set`. I was unable to build a deterministic reproduction, so I am reporting it as an unverified gap rather than a bug.

One of these initially passed against broken code: the torn-bucket test used an invalid-JSON body, which the row parser discards whether or not the hash is verified. It was rewritten to use a **valid** body with content that disagrees with the recorded hash, which is the only shape that can show the check doing work.

`npm test`: 1720 passing, 0 failing. `tsc --noEmit` is unchanged from base (same 30 pre-existing errors, confirmed by stashing).

## Summary by CodeRabbit

- **New Features**
  - Improved vector index persistence with stable, content-based storage and configurable bucket sizing.
  - Added incremental saves that update only changed data.
  - Added support for migrating existing persisted indexes.
  - Improved loading behavior by preserving valid entries when data is missing or corrupted.
  - Added automatic index rebuilding when persisted vector data cannot be recovered.

- **Bug Fixes**
  - Improved cleanup of obsolete persisted data and recovery from interrupted saves.
  - Delete operations now clearly report whether they succeeded.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -464,7 +464,13 @@ async function main() {
     }
   }
 
-  const needsRebuild = bm25Index.size === 0;
+  // A vector load that rejected every bucket it read is the one failure BM25
+  // cannot stand in for: BM25 restores fine, so the store looks healthy, while
+  // the vectors sit on disk unreadable and nothing re-reads them. Rebuilding is
+  // the only exit. Kept off `vectorIndex.size === 0`, which is also the ordinary
+  // state of a store whose vectors were simply never written.
+  const needsRebuild =
+    bm25Index.size === 0 || (vectorIndex !== null && loaded?.vectorRejected === true);
 
   if (needsRebuild) {
     // Fire-and-forget. rebuildIndex iterates every observation across
--- a/src/state/index-persistence.ts
+++ b/src/state/index-persistence.ts
@@ -32,28 +32,47 @@ type IndexShardManifest = {
 /**
  * Bucketed vector manifest.
  *
- * `shards` maps bucket key -> that bucket's content hash, chunk count, and
- * length, and holds only non-empty buckets.
+ * `shards` maps bucket key -> content hash and chunk count, non-empty buckets
+ * only.
  *
- * Keeping the hashes *in the manifest* rather than in process memory is what
- * makes this work in production: the service restarts often, and an in-memory
- * cache would force a full rewrite of every bucket after each restart, which
- * is the cost this change exists to remove.
- *
- * A bucket is still chunked to `shardChars`, because bucket size is unbounded
- * — it grows with the corpus — and the engine rejects an oversized
- * `state::set` payload. Chunking is what bounds the payload; bucketing is what
- * bounds how much gets rewritten. They solve different problems and both are
- * needed.
+ * Hashes live in the manifest, not process memory: restarts must not force a
+ * full rewrite. Buckets bound how much is rewritten; chunks bound the payload
+ * size, since a bucket grows with the corpus and the engine rejects an
+ * oversized `state::set`.
  */
-type VectorBucketEntry = { hash: string; chunks: number; chars: number };
+// No `chars`: the sha1 already proves both content and length. v1 needed a
+// length check only because it had no hash to check against.
+type VectorBucketEntry = { hash: string; chunks: number };
+
+/**
+ * Version of the *addressing scheme* — which bucket an obsId lands in and which
+ * keys that bucket occupies. Bump it whenever that mapping changes, including a
+ * change to the bucket hash function or the key format.
+ *
+ * This exists because a content hash cannot see an addressing change: the bytes
+ * are identical, they just belong somewhere else now. Without this, such a
+ * change silently skips every write while the manifest claims new locations,
+ * and the vectors are lost on the next load.
+ *
+ * It is deliberately separate from `v`. `v` describes the manifest's shape, and
+ * keeping it stable is what lets a newer build still parse an older manifest
+ * well enough to reclaim the keys it named. Bumping `v` instead would make the
+ * old manifest unreadable and strand every key in it.
+ */
+const VECTOR_LAYOUT = 1;
 
 type VectorBucketManifest = {
   v: 2;
+  layout: number;
   buckets: number;
+  chunkChars: number;
   shards: Record<string, VectorBucketEntry>;
 };
 
+// `layout` and `chunkChars` are not required here on purpose. A manifest
+// missing them simply fails the layout comparison, which triggers a full
+// rewrite and a full reclaim — the correct, self-healing response — whereas
+// rejecting the manifest outright would strand the keys it names.
 function isVectorBucketManifest(value: unknown): value is VectorBucketManifest {
   if (!value || typeof value !== "object") return false;
   const candidate = value as Partial<VectorBucketManifest>;
@@ -69,13 +88,13 @@ function isVectorBucketManifest(value: unknown): value is VectorBucketManifest {
 function isValidBucketEntry(value: unknown): value is VectorBucketEntry {
   if (!value || typeof value !== "object") return false;
   const candidate = value as Partial<VectorBucketEntry>;
+  // The chunks bound matters beyond shape: it is the loop bound for both the
+  // reclaim walks below, so a poisoned value would spin a garbage-length loop.
   return (
     typeof candidate.hash === "string" &&
     candidate.hash.length > 0 &&
     Number.isInteger(candidate.chunks) &&
-    (candidate.chunks as number) >= 1 &&
-    Number.isInteger(candidate.chars) &&
-    (candidate.chars as number) >= 0
+    (candidate.chunks as number) >= 1
   );
 }
 
@@ -247,11 +266,8 @@ export class IndexPersistence {
    * Persist the vector index as fixed, independently-addressed buckets,
    * writing only the buckets whose contents actually changed.
    *
-   * The offset-chunked format this replaces was O(entire corpus) per change:
-   * shard boundaries are character offsets into one serialised blob, so a
-   * single inserted vector shifts every downstream byte and rewrites every
-   * shard. In production that was 431 MB per save against a corpus of 35 MB of
-   * observations, and no save had completed in over four hours.
+   * Replaces an offset-chunked format that was O(entire corpus) per change;
+   * see the commit message for the production measurements.
    *
    * Cross-bucket atomicity is deliberately given up here. It buys nothing
    * today, because the atomic save never completes. Recovery is convergent
@@ -267,11 +283,22 @@ export class IndexPersistence {
     const buckets = vectorBucketCount(this.options);
     const chunkChars = shardChars(this.options);
 
-    // Only trust previous hashes when the bucket count matches. A different
-    // count remaps every obsId, so every stored bucket is meaningless.
     const previousV2 = isVectorBucketManifest(previous) ? previous : null;
-    const previousShards =
-      previousV2 && previousV2.buckets === buckets ? previousV2.shards : {};
+
+    // Every input that decides *where* a vector's bytes live. If any of them
+    // moved, a matching content hash proves nothing — the bytes are the same
+    // but their home is not — so nothing on disk may be skipped.
+    const layoutMatches =
+      !!previousV2 &&
+      previousV2.layout === VECTOR_LAYOUT &&
+      previousV2.buckets === buckets &&
+      previousV2.chunkChars === chunkChars;
+
+    // One map, used for two things: comparison (only when the layout matches)
+    // and reclaim (always). Reclaim must see the previous manifest even on a
+    // layout change, because those keys are exactly the ones about to become
+    // unreachable, and nothing can enumerate them afterwards.
+    const prior = previousV2?.shards ?? {};
 
     const nextShards: Record<string, VectorBucketEntry> = {};
     // Nothing is deleted until the manifest naming its replacement is live.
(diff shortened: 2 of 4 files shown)


inix-x:
Closing this one. None of the 10 commits here are in production (`inix-x/agentmemory` branch `production`, deployed on Railway at `fc218d1`). Production's `src/state/index-persistence.ts` has zero `bucket` references, and runs generation-based vector persistence instead.

The bucketing approach has not been run against the 6.9 GB production data volume, so there is no measurement to back the perf claim in the title. Upstream `main` has not moved since 2026-08-23 (`e04ba88`). The branch `fix/vector-index-bucketing` stays on `inix-x`. If a sandbox hold on the real volume shows a win, this comes back as a fresh PR with the numbers.

PR #1261, opened 2026-08-26
fix(test): point USERPROFILE at the temp home on Windows

anhtahaylove:
Two tests in `test/slots-flag-gate.test.ts` fail on Windows for anyone whose own `~/.agentmemory/.env` sets `AGENTMEMORY_SLOTS` or `AGENTMEMORY_REFLECT` to `false`.

## Cause

`beforeEach` creates a temp home and points `HOME` at it, but `src/functions/slots.js` resolves the config path through `os.homedir()` — and on Windows that reads `USERPROFILE`, ignoring `HOME` entirely:

```
> process.env.HOME = 'C:/tmp/fake-home'
> os.homedir()
'C:\Users\Administrator'
```

So the flag is read from the developer's real `.env` rather than the temp file the test just wrote. CI is green because `HOME` and `USERPROFILE` agree on Linux.

## Fix

Set `USERPROFILE` alongside `HOME` in both `describe` blocks and restore it in `afterEach`.

## Verification

On a clean checkout of `main`, Windows:

```
before: Tests  2 failed | 2 passed (4)
after:  Tests  4 passed (4)
```

`slots-flag-gate` is the only test file that overrides `HOME` without `USERPROFILE` — the other nine already set both, so this brings it in line with them.

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform test reliability by ensuring temporary home-directory settings work correctly on Windows.
  * Added cleanup to restore the original environment after each test.

### DIFF
--- a/test/slots-flag-gate.test.ts
+++ b/test/slots-flag-gate.test.ts
@@ -13,20 +13,27 @@ import { join } from "node:path";
 describe("isSlotsEnabled — reads merged env (#678)", () => {
   let home: string;
   let ORIG_HOME: string | undefined;
+  let ORIG_USERPROFILE: string | undefined;
   let ORIG_FLAG: string | undefined;
 
   beforeEach(() => {
     home = mkdtempSync(join(tmpdir(), "am-slots-flag-"));
     mkdirSync(join(home, ".agentmemory"), { recursive: true });
     ORIG_HOME = process.env["HOME"];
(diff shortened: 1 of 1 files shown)


anhtahaylove:
Context for this one: #1264 audits the whole Windows test picture — 30 failures, what causes each group, and which PR covers it.

PR #1262, opened 2026-08-27
fix(test): normalise fs mock paths so they match on Windows

anhtahaylove:
`test/compress-file.test.ts` and `test/obsidian-export.test.ts` fail on Windows — 13 tests between them — because their in-memory `node:fs/promises` mocks are keyed on the raw path string while the functions under test resolve the path first.

## Cause

```js
resolve("/tmp/notes.md")   // Linux:   /tmp/notes.md
                           // Windows: C:\tmp\notes.md
```

`compress-file` calls `resolve(data.filePath)` before its first `lstat`. The mock's `fileStore` has `/tmp/notes.md`, the lookup asks for `C:\tmp\notes.md`, `lstat` throws ENOENT, and the function returns `file not found` — for a fixture the test wrote two lines earlier. `obsidian-export` has the same mismatch through `join()`.

That is why the failures read like unrelated assertion noise (`expected 'file not found' to contain 'symlink'`, `expected undefined to be defined`) rather than a path problem.

## Fix

A small `key()` normaliser applied where paths enter the mock maps, plus three assertions that compare a returned path or error message against a POSIX literal. No production code changes.

## Verification

Windows, clean checkout of `main`:

```
compress-file    before: 5 failed | 2 passed (7)     after: 7 passed (7)
obsidian-export  before: 8 failed | 8 passed (16)    after: 16 passed (16)

full suite       before: 30 failed | 1682 passed     after: 17 failed | 1695 passed
```

Linux is unaffected: `key()` is an identity transform on paths that are already POSIX.

The remaining 17 are separate Windows issues (process spawning, file locking, `cli-lifecycle-safety` fixtures) — happy to look at those in follow-ups if useful.

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform test coverage for file compression and export workflows.
  * Added consistent path handling across Windows and POSIX-style environments.
  * Strengthened verification of backup paths, generated files and directories, and export-related error messages.
  * Reduced false failures caused by differences in path separators and drive-letter formatting.

### DIFF
--- a/test/compress-file.test.ts
+++ b/test/compress-file.test.ts
@@ -8,12 +8,19 @@ const fileStore = new Map<string, string>();
 const symlinkPaths = new Set<string>();
 const openEloopPaths = new Set<string>();
 
+// The function resolves filePath before touching the fs, and resolve() turns
+// a POSIX fixture path into a drive-rooted one on Windows ("/tmp/notes.md" ->
+// "C:\tmp\notes.md"), so raw-path keys never match there.
+function key(path: string): string {
+  return path.replace(/\\/g, "/").replace(/^[A-Za-z]:/, "");
+}
+
 vi.mock("node:fs/promises", () => ({
   lstat: vi.fn(async (path: string) => {
-    if (symlinkPaths.has(path)) {
+    if (symlinkPaths.has(key(path))) {
       return { isSymbolicLink: () => true };
     }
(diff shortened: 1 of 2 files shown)


anhtahaylove:
Context for this one: #1264 audits the whole Windows test picture — 30 failures, what causes each group, and which PR covers it.

Issue #1203, opened 2026-08-14
mem::summarize is dispatched twice per Stop hook


inix-x:
`plugin/scripts/stop.mjs` fires two requests on every Stop hook, and both reach `mem::summarize`.

```js
// plugin/scripts/stop.mjs
line 27   fetch(`${REST_URL}/agentmemory/summarize`, ...)
line 33   fetch(`${REST_URL}/agentmemory/session/end`, ...)
```

The second one is not independent. It fans out:

```
POST /agentmemory/session/end
  -> api::session::end                 (src/triggers/api.ts:646)
  -> sdk.trigger event::session::stopped, TriggerAction.Void()   (api.ts:661)
  -> event::session::stopped handler   (src/triggers/events.ts:97)
  -> sdk.trigger mem::summarize        (events.ts:98)
```

So one Stop hook produces two concurrent `mem::summarize` runs on the same `sessionId`, a few milliseconds apart. Neither takes a lease. `mem::lease-acquire` is registered, and `src/functions/summarize.ts` never calls it.

## Observed on a self-hosted 0.9.28 deployment

`GET /agentmemory/health` `functionMetrics`, over roughly 21 hours of uptime:

| functionId | totalCalls | failureCount | avgLatencyMs |
|---|---|---|---|
| `mem::summarize` | 236 | 52 (22%) | 13,006 |
| `mem::compress` | 1,051 | 34 (3%) | 7,150 |

The duplication is visible in the log. Two chunking lines a fraction of a millisecond apart, then two completions with different titles, because both runs did their own LLM pass over the same observations:

```
21:51:07.160774  Summarize chunking session {"sessionId":"<redacted>","chunks":2,"chunkSize":400,"concurrency":6,"totalObservations":492}
21:51:07.160782  Summarize chunking session {"sessionId":"<redacted>","chunks":2,"chunkSize":400,"concurrency":6,"totalObservations":492}
21:51:47.162719  Session summarized {"sessionId":"<redacted>","title":"AgentMemory migration complete: ...","decisions":13,"qualityScore":100}
21:51:47.162724  Session summarized {"sessionId":"<redacted>","title":"Claude-Mem to AgentMemory Migration: ...","decisions":14,"qualityScore":100}
```

## The multiplier

[...]

Issue #1127, opened 2026-07-30
mem::graph-extract hangs to its own 180s invocation timeout on every batch, eventually stalling state::get and crashing the worker


deemaxx:
## Setup

- agentmemory: `@agentmemory/agentmemory@0.9.27` at time of incident (since upgraded to 0.9.28 during investigation — hang persisted after upgrade based on log/behavior review, not independently re-verified with a fresh 7-day soak)
- Runtime: macOS (darwin arm64), launched via LaunchAgent (`com.agentmemory.server.plist`)
- LLM provider: local Ollama via OpenAI-compatible endpoint (`OPENAI_BASE_URL=http://localhost:11434/v1`, `OPENAI_MODEL=qwen2.5-coder:7b`)
- `AGENTMEMORY_LLM_TIMEOUT_MS=120000`
- `GRAPH_EXTRACTION_ENABLED=true`
- Hourly cron (`com.agentmemory.graph-build.plist`, `StartInterval: 3600`) calling `POST /agentmemory/graph/build` with `{"batchSize":10}`

## Symptom

Starting ~2026-07-23T19:57Z, every hourly `graph/build` cron invocation returned an empty response body (`curl -s -m 3300 ...` timed out or returned nothing) — this repeated roughly every hour for **7 straight days** with zero successful runs, versus clean `{"batches":N,"edges":N,"nodes":N,"sessions":N,"success":true}` results before that point.

Server-side log (`~/Library/Logs/agentmemory-server.log`) shows the actual failure, repeated across many sessions and restarts:

```
[agentmemory] warn graph-build batch failed {"sessionId":"<id>","batchIndex":N,"error":"Invocation timeout after 180000ms: mem::graph-extract"}
```

This fires for many `batchIndex` values per session, across multiple distinct sessions, across at least 9 separate worker restarts recorded in the log. Confirmed live: `POST /agentmemory/graph/build` from a plain `curl -v` hangs and returns **zero bytes** even after 20s (well under both the 120s LLM timeout and the 180s function-invocation timeout, so this isn't just "slow LLM").

At least once, the accumulated hang escalated further and stalled the graph snapshot read itself:

```
[agentmemory] warn Graph snapshot read failed {"error":"Invocation timeout after 180000ms: state::get"}
```

This appears to correlate with full worker crashes/restarts (`KeepAlive.SuccessfulExit=false` in the plist relaunches the process, which then repeats the same pattern).

[...]

PR #1269, opened 2026-08-27
perf(summarize): memoize per-chunk partials and skip unchanged sessions

### LINKED ISSUE #1131: Session summarization is quadratic: the Stop hook re-summarizes the entire session every assistant turn
# Session summarization is quadratic: the `Stop` hook re-summarizes the entire session every assistant turn

**Version:** 0.9.27 (npm `@agentmemory/agentmemory`)
**Provider:** Ollama (llama3.2), local
**Platform:** Linux, Claude Code plugin hooks

## Summary

`plugin/scripts/stop.mjs` runs on Claude Code's `Stop` hook, which fires at the end of **every assistant turn** — not at session end. It POSTs the session ID to `/agentmemory/summarize`, and the server re-summarizes the **entire session from scratch** each time.

Because observation count grows monotonically through a session, and chunk count grows with it, total LLM calls over a session's life are **O(N²/chunkSize)** rather than O(N). Long sessions saturate the local LLM, trip the circuit breaker, and then every chunk fails instantly.

## Why it's quadratic

`produceSummaryXml(provider, compressed, sessionId, project)` chunks the full `compressed` observation array on every invocation:

```js
const chunks = [];
for (let i = 0; i < compressed.length; i += chunkSize) chunks.push(compressed.slice(i, i + chunkSize));
```

There's no memo of previously-summarized observations and no partial reuse — chunk *k* is recomputed identically on every turn for the rest of the session. Only the final merge differs.

## Observed

One session on 2026-07-28, logged over ~40 minutes. Each line is a separate turn triggering a full re-summarization:

```
chunks=20  totalObservations=229
chunks=21  totalObservations=241
chunks=21  totalObservations=251
chunks=24  totalObservations=279
chunks=28  totalObservations=327
chunks=28  totalObservations=331
...
chunks=40  totalObservations=480
```

That session eventually saturated Ollama. Once the circuit breaker opened, every chunk failed immediately:

```
warn  Summarize chunk LLM call failed {"chunk":"36/40","attempt":2,"error":"circuit_breaker_open"}
warn  Summarize chunk LLM call failed {"chunk":"37/40","attempt":1,"error":"circuit_breaker_open"}
...
error Summarize failed {"error":"too_many_chunks_skipped: 40/40 chunks failed to parse after retry"}
```

**2,976 failure records in one night**, versus 0–14/day on normal days.

A smaller, current example — 32 chunk-LLM calls to summarize a session holding 47 observations, where incremental would need ~4:

```
06:50  chunks=2  obs=16
07:29  chunks=3  obs=31
07:32  chunks=3  obs=35
07:33  chunks=4  obs=38
07:36  chunks=4  obs=47
```

## Interaction with `SUMMARIZE_CHUNK_SIZE`

Lowering `SUMMARIZE_CHUNK_SIZE` from the default 400 is currently the recommended fix for XML parse failures on small-context local models (400 observations builds a prompt far past llama3.2's default Ollama context, truncating the format instructions). That fix works — but it **amplifies this bug**, because smaller chunks mean proportionally more LLM calls per re-summarization. At `chunkSize=12`, a 480-observation session is 40 chunks per turn.

So the two knobs pull against each other: the setting that makes each call *succeed* makes the redundant work *worse*.

## Secondary: each summarize appears to run twice

Every `Summarize chunking session` line is logged exactly twice with an identical timestamp:

```
07:29:12  Summarize chunking session {"chunks":3,...,"totalObservations":31}
07:29:12  Summarize chunking session {"chunks":3,...,"totalObservations":31}
```

`stop.mjs` is the only script that POSTs to `/summarize`, so I couldn't account for the 2×. It may be duplicate logging rather than duplicate work, but if it's the latter it doubles the cost above. Happy to dig further if useful.

## Suggested fixes

In rough order of preference:

1. **Incremental summarization** — persist per-chunk partials keyed by observation range/hash and only summarize chunks containing new observations, merging against the stored partials. This is the real fix and turns the cost linear.
2. **Debounce server-side** — skip a summarize request if one completed for that session within the last *N* seconds (configurable). Cheap to implement, doesn't fix the asymptotics but caps the damage.
3. **Decouple from `Stop`** — `Stop` firing per-turn seems like an unintended reading of the hook's semantics. If per-turn freshness isn't required, `SessionEnd` alone would be dramatically cheaper. Note `session-end.mjs` currently does **not** call `/summarize`, so moving it there would need that added.

## Workaround in use

A wrapper on the `Stop` hook that lets the full path through at most once per 10 minutes per session, falls back to the cheap `/session/end` ping otherwise, and forces a final summarize on `SessionEnd` so the session tail isn't lost. Happy to share if it's useful to others hitting this, though it's obviously a band-aid over (1).

## Environment notes

- `SUMMARIZE_CHUNK_SIZE=12`, `SUMMARIZE_CHUNK_CONCURRENCY=2`, `OPENAI_TIMEOUT_MS=120000`
- These are read from `process.env` only — worth documenting, since setting them in `~/.agentmemory/.env` has no effect (that file is loaded for provider config and never merged into `process.env`). On systemd they must go in a unit drop-in.

DanielCarmingham:
## What

`mem::summarize` re-summarizes a session from scratch on every run. The Claude Code Stop hook fires on **every assistant turn**, not at session end, so a session that reaches N observations does O(N²) summarization work over its lifetime — and for sessions past the chunking threshold, every turn re-runs *every* chunk's LLM call, not just the part that changed.

This adds two layers of change detection:

1. **Session-level skip** — hash the compressed observation set (`sourceFingerprint`) and return the stored summary unchanged when the input is byte-identical to what produced it.
2. **Chunk-level memoization** — cache each chunk's summary partial under a content-addressed key (`chunkFingerprint`), so appending one observation to a 400-observation session re-runs one chunk instead of all of them.

Plus the cache-reclamation paths those two require, and a per-session lock so concurrent runs don't both do the full work.

## What the fingerprints hash

Both keys hash exactly the fields `buildSummaryPrompt` renders — `id`, `type`, `title`, `narrative`, `facts`, `files` — so a digest match means the LLM would receive a byte-identical prompt. `src/prompts/summary.ts` is the contract: a field that reaches the prompt without reaching the hash is a silent stale-summary bug, and there is a comment on `hashObservationSet` saying so. `concepts` is deliberately excluded — `buildSummaryPrompt` declares it but its template never renders it, so hashing it would force re-summarization that cannot change the output.

Fields are length-prefixed before hashing, so `facts: ["a", "bc"]` and `["ab", "c"]` do not collide.

An id-only key would be unsound: a same-id rewrite (bulk `import-jsonl` with `strategy !== "skip"`, or `replay` over an edited transcript) preserves every id while changing the text. Three tests cover the cases a weaker key would miss — a title-and-narrative rewrite, a facts-only change, and a files-only change.

An `observationCount` comparison is also unsound as a change signal: observations are **not** append-only. Four paths delete from live sessions — `evict.ts` per-observation eviction (age + importance, on a timer), `evict.ts` project-scoped eviction, `mem::forget` with explicit `observationIds`, and `export-import.ts`. A delete plus an equal-size append leaves the count unchanged and the content entirely different.

## Ordering

`kv.list` is not order-stable — the `file_based` store is a hash map. `sourceFingerprint` sorts defensively, but chunk *boundaries* must be positionally stable across turns too, or every chunk key changes when the list order shifts and the cache never hits. The handler therefore sorts by id once before both fingerprinting and chunk slicing.

That sort doubles as a correctness fix: `generateId`'s base36 timestamp prefix is a fixed 8 chars until 2059, so id order is chronological order. Observations now reach the LLM — and the reduce step's `obsRangeStart`/`obsRangeEnd` labels — in chronological order instead of arbitrary hash order. `test/summarize.test.ts` covers a permuted `kv.list` between turns.

## Cache reclamation

`KV.summaryChunks` is introduced here, so this change owns its cleanup. Entries are reclaimed on:

- whole-session delete (`mem::forget`, `remember.ts`)
- stale-session and per-observation eviction (`evict.ts`) — per-observation eviction shifts every downstream chunk boundary, and whole-session delete never fires for a live session, so nothing else would reclaim these
- explicit-ids `mem::forget`, same reason
- every summarize run, which prunes entries whose keys are no longer live chunks

Without the last one the trailing chunk's key changes on each append and the cache grows unbounded.

## On #1203

The per-session `withKeyedLock` here does not remove the duplicate dispatch — `api::summarize` stays publicly reachable, so two concurrent full-history passes over identical input are still possible. It serializes them, so the second one reaches the fingerprint skip instead of repeating the LLM work. Removing the duplicate trigger itself is a separate change.

## How to verify

```bash
npx vitest run test/summarize.test.ts test/evict.test.ts test/remember-forget-audit.test.ts
npm test        # full suite: 159 files, 1725 pass, 1 skip
npx tsc --noEmit
```

`tsc` reports 30 pre-existing errors on `main` and the same 30 here — none in any file this branch touches.

New coverage: concurrent-run serialization; unchanged-session skip; re-summarize on append; re-summarize on a same-id rewrite, on a facts-only change, and on a files-only change; chunk reuse on append; chunk reuse under permuted `kv.list`; dead-entry pruning; and cache reclamation from each eviction and forget path.

Fixes #1131.

## Summary by CodeRabbit

* **New Features**
  * Improved summary generation with per-chunk caching and change detection.
  * Added support for forcing summary regeneration.
  * Summaries now process observations chronologically for more consistent results.

* **Bug Fixes**
  * Prevented duplicate summary work during concurrent session processing.
  * Prevented deleted sessions from receiving orphaned summaries.
  * Summary caches are cleared when sessions or observations are removed, including imports and automatic cleanup.
  * Unchanged summaries are preserved while outdated cached chunks are cleaned up.
  * Dry-run evictions no longer modify summary caches.

### DIFF
--- a/src/functions/auto-forget.ts
+++ b/src/functions/auto-forget.ts
@@ -5,6 +5,7 @@ import { StateKV } from "../state/kv.js";
 import { recordAudit } from "./audit.js";
 import { deleteAccessLog } from "./access-tracker.js";
 import { getSearchIndex, vectorIndexRemove, flushIndexSave } from "./search.js";
+import { deleteSummaryChunks } from "./summarize.js";
 import { logger } from "../logger.js";
 
 const MS_PER_DAY = 24 * 60 * 60 * 1000;
@@ -157,6 +158,8 @@ export function registerAutoForgetFunction(sdk: ISdk, kv: StateKV): void {
         );
         obsPerSession.push(...results);
       }
+      const touchedSessionIds = new Set<string>();
+
       for (let i = 0; i < sessions.length; i++) {
         for (const obs of obsPerSession[i]) {
           if (!obs.timestamp) continue;
@@ -172,6 +175,7 @@ export function registerAutoForgetFunction(sdk: ISdk, kv: StateKV): void {
                 deletedOk = false;
               }
               if (deletedOk) {
+                touchedSessionIds.add(sessions[i].id);
                 if (obs.imageData) await decrementImageRef(kv, sdk, obs.imageData);
                 if (obs.imageRef && obs.imageRef !== obs.imageData) {
                   await decrementImageRef(kv, sdk, obs.imageRef);
@@ -188,6 +192,10 @@ export function registerAutoForgetFunction(sdk: ISdk, kv: StateKV): void {
             }
           }
         }
+      }
+
+      for (const sessionId of touchedSessionIds) {
+        await deleteSummaryChunks(kv, sessionId);
       }
 
       if (!dryRun && (result.ttlExpired.length > 0 || result.lowValueObs.length > 0)) {
--- a/src/functions/evict.ts
+++ b/src/functions/evict.ts
@@ -13,6 +13,7 @@ import { recordAudit } from "./audit.js";
 import { deleteAccessLog } from "./access-tracker.js";
 import { logger } from "../logger.js";
 import { deleteSummaryChunks } from "./summarize.js";
+import { withKeyedLock, sessionWriteLockKey } from "../state/keyed-mutex.js";
 
 interface EvictionConfig {
   staleSessionDays: number;
@@ -171,7 +172,19 @@ export function registerEvictFunction(sdk: ISdk, kv: StateKV): void {
             }
 
             try {
-              await kv.delete(KV.sessions, session.id);
+              // Shared with mem::summarize's write step: a run that
+              // already passed its session-exists check would otherwise
+              // write these rows back with nothing left to reclaim them.
+              // The KV.summaries delete covers the ordering where that run
+              // wins the lock — this branch is otherwise reached only for
+              // sessions with no summary, though they can still hold chunk
+              // partials from a run whose final reduce failed after the
+              // per-chunk calls succeeded.
+              await withKeyedLock(sessionWriteLockKey(session.id), async () => {
+                await kv.delete(KV.sessions, session.id);
+                await kv.delete(KV.summaries, session.id);
+                await deleteSummaryChunks(kv, session.id);
+              });
               stats.staleSessions++;
             } catch (err) {
               logger.warn("Eviction delete failed", {
@@ -181,10 +194,6 @@ export function registerEvictFunction(sdk: ISdk, kv: StateKV): void {
               });
               continue;
             }
-            // Reached only for sessions with no KV.summaries entry, which
-            // can still hold chunk partials from a run whose final reduce
-            // failed after the per-chunk calls succeeded.
-            await deleteSummaryChunks(kv, session.id);
             await recordAudit(kv, "delete", "mem::evict", [session.id], {
               resource: "session",
               reason: recovered
--- a/src/functions/evict.ts
+++ b/src/functions/evict.ts
@@ -172,14 +172,9 @@ export function registerEvictFunction(sdk: ISdk, kv: StateKV): void {
             }
 
             try {
-              // Shared with mem::summarize's write step: a run that
-              // already passed its session-exists check would otherwise
-              // write these rows back with nothing left to reclaim them.
-              // The KV.summaries delete covers the ordering where that run
-              // wins the lock — this branch is otherwise reached only for
-              // sessions with no summary, though they can still hold chunk
-              // partials from a run whose final reduce failed after the
-              // per-chunk calls succeeded.
+              // The KV.summaries delete is not dead code: this branch is
+              // reached only for sessions with no summary, but a summarize
+              // run that wins the lock leaves one behind.
               await withKeyedLock(sessionWriteLockKey(session.id), async () => {
                 await kv.delete(KV.sessions, session.id);
                 await kv.delete(KV.summaries, session.id);
--- a/src/functions/evict.ts
+++ b/src/functions/evict.ts
@@ -12,6 +12,7 @@ import { isConsolidationEnabled } from "../config.js";
 import { recordAudit } from "./audit.js";
 import { deleteAccessLog } from "./access-tracker.js";
 import { logger } from "../logger.js";
+import { deleteSummaryChunks } from "./summarize.js";
 
 interface EvictionConfig {
   staleSessionDays: number;
@@ -180,6 +181,10 @@ export function registerEvictFunction(sdk: ISdk, kv: StateKV): void {
               });
               continue;
             }
+            // Reached only for sessions with no KV.summaries entry, which
+            // can still hold chunk partials from a run whose final reduce
+            // failed after the per-chunk calls succeeded.
+            await deleteSummaryChunks(kv, session.id);
             await recordAudit(kv, "delete", "mem::evict", [session.id], {
               resource: "session",
               reason: recovered
@@ -194,6 +199,12 @@ export function registerEvictFunction(sdk: ISdk, kv: StateKV): void {
       if (!dryRun && recoveredStaleSessions > 0) {
         await runRecoveredSessionConsolidation(sdk);
       }
+
+      // Evicting one observation shifts every downstream chunk boundary,
+      // orphaning the session's whole chunk cache. Collect sessions here
+      // and flush each once after both loops, rather than per evicted
+      // observation.
+      const touchedSessionIds = new Set<string>();
 
       const projectObs = new Map<string, CompressedObservation[]>();
       for (const session of sessions) {
@@ -225,6 +236,7 @@ export function registerEvictFunction(sdk: ISdk, kv: StateKV): void {
                 });
                 continue;
               }
+              touchedSessionIds.add(session.id);
               if (o.imageData) await decrementImageRef(kv, sdk, o.imageData);
               if (o.imageRef && o.imageRef !== o.imageData) await decrementImageRef(kv, sdk, o.imageRef);
               await recordAudit(kv, "delete", "mem::evict", [o.id], {
@@ -268,6 +280,7 @@ export function registerEvictFunction(sdk: ISdk, kv: StateKV): void {
                 });
                 continue;
               }
+              touchedSessionIds.add(o.sessionId);
               if (o.imageData) await decrementImageRef(kv, sdk, o.imageData);
               if (o.imageRef && o.imageRef !== o.imageData) await decrementImageRef(kv, sdk, o.imageRef);
               await recordAudit(kv, "delete", "mem::evict", [o.id], {
@@ -279,6 +292,14 @@ export function registerEvictFunction(sdk: ISdk, kv: StateKV): void {
             }
           }
         }
+      }
+
+      // Empty on a dryRun pass, so this needs no separate dryRun guard.
+      // Kept sequential rather than Promise.all: a sweep can touch
+      // hundreds of sessions, and the file-based KV adapter stalls under
+      // that much concurrent fan-out (upstream #1127).
+      for (const sessionId of touchedSessionIds) {
+        await deleteSummaryChunks(kv, sessionId);
       }
 
       const memories = await kv.list<Memory>(KV.memories).catch(() => []);
--- a/src/functions/export-import.ts
+++ b/src/functions/export-import.ts
@@ -33,6 +33,7 @@ import { VERSION } from "../version.js";
 import { recordAudit } from "./audit.js";
 import { indexRecords } from "./search.js";
 import { deleteSummaryChunks } from "./summarize.js";
+import { withKeyedLock, sessionWriteLockKey } from "../state/keyed-mutex.js";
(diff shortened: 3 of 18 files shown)


DanielCarmingham:
Thanks — the fingerprint finding was correct and I've pushed a fix. Summary of what I did with each point:

**Fingerprint completeness — fixed.** This was worse than described. `buildSummaryPrompt` renders six fields (`type`, `title`, `narrative`, `facts`, `files`, plus identity); the hash covered `id`, `title`, and `narrative.length`. So `type`, `facts` and `files` were not merely length-approximated, they were unhashed entirely — any change to them false-skipped at both the session and chunk level.

`hashObservationSet` now hashes exactly what the prompt renders, length-prefixed so `["a","bc"]` and `["ab","c"]` don't collide. `concepts` stays out on purpose: `buildSummaryPrompt` declares it but the template never renders it, so hashing it would force re-summarization that cannot change the output. There's a comment on the function naming `src/prompts/summary.ts` as the contract to keep in sync.

Two tests added — a facts-only change and a files-only change, both with `id`, `title` and `narrative` byte-identical. I verified both fail against the previous hash and pass against the new one.

I did not switch to `fingerprintId()`; this needs incremental `createHash().update()` per observation to avoid materializing a multi-megabyte string per turn on the largest sessions, which is on the Stop-hook hot path.

**Parallel deletes in `summarize.ts` — applied.** Both the prune loop and `deleteSummaryChunks` now use `Promise.all`. These are bounded by chunk count for a single session, so the fan-out is small.

**Parallel cleanup in `evict.ts` — declined, with a comment explaining why.** That loop is different in kind: it fans out across every session an eviction pass touched, which can be hundreds, each doing its own `kv.list` plus deletes. Unbounded concurrency of that size against the `file_based` KV adapter is the failure mode described in #1127, and every one of those calls inherits the 180s worker-level `invocationTimeoutMs` (#1128). The linear wait is the better trade until per-call KV timeouts land. Left sequential with a comment recording the reasoning.

**Explanatory comments — trimmed.** Fair, and I'd overshot badly: comments were 56% of added source lines against a ~9% baseline in this file. Now at 29%, with the remainder limited to non-obvious invariants — why observations are sorted before chunk slicing, why the cache owns its own reclamation, and the prompt/hash contract above.

Also squashed the branch from eight commits to two.

DanielCarmingham:
The `remember.ts` finding was correct, and chasing it turned up two more instances of the same gap. Pushed as `18238f3`.

**Missed reclamation paths — fixed, three of them.** Any path that deletes observations while leaving the session alive has to reclaim the chunk cache, because removing one observation shifts every downstream chunk boundary. Three did not:

- `remember.ts` — explicit-ids `mem::forget`. The session survives, so it never reaches the whole-session branch that already reclaims.
- `export-import.ts` — replace-strategy import. Wipes the session rows the cache is keyed against, so nothing afterwards can find those entries to reclaim.
- `auto-forget.ts` — hourly low-value observation deletion. Same shape as the eviction paths, and the most costly of the three because it recurs.

A test per path, each verified to fail before the fix. Full suite 1728 pass / 1 skip; `tsc` unchanged at the 30 pre-existing errors.

The PR description already listed explicit-ids `mem::forget` as a reclamation site, so that was a description/behaviour mismatch on my side rather than a difference of opinion. Thanks for catching it.

**`fingerprintId()` — still declining, same reason as before.** Its signature is `fingerprintId(prefix: string, content: string)`: it takes one complete string. Both fingerprints here hash incrementally via `createHash().update()` per observation precisely so the full serialized content is never materialized — the largest session in the corpus I tested against is ~28.8k observations, which is several megabytes per call on a path that runs every assistant turn. Adopting `fingerprintId()` would mean building that string to hash it once and discard it. The prefixes (`sfp_`, `chk_`) and the 16-hex-char truncation already match its output format, so the keys are indistinguishable from `fingerprintId()`-produced ones; only the construction differs.

Happy to switch if there's a repo convention I'm missing that outweighs the allocation.

DanielCarmingham:
The finding was correct for the paths that delete a whole session, and I've pushed a fix as `b7e084c`. I did not take the approach suggested, though — using `mem:summarize:${sessionId}` itself would mean a user-initiated `mem::forget` blocks behind an LLM round trip, up to the 180s worker timeout. So it is a generation check *and* a narrower lock:

- **`mem::summarize` re-checks the session before writing.** After the provider call it confirms `KV.sessions` still holds the row, and discards the summary plus any chunk partials written during the call when it does not.
- **Deletion paths take a new key, `sessionWriteLockKey()`,** covering only the summary-row write. The re-check and the write are separate awaits, so the check alone leaves a window a delete can still land in; the shared key closes it without anyone waiting on a provider call.

Lock order is `mem:summarize:S` → `mem:session-write:S` in the summarizer and `mem:session-write:S` alone everywhere else, so there is no cycle.

**On the four sites listed.** Two of them are the permanent case and are fixed — `remember.ts#L318-320` (whole-session `mem::forget`) and `export-import.ts#L321-322` (replace-strategy import). The list also missed one of the same kind: `evict.ts` stale-session eviction deletes `KV.sessions`, so it had the identical exposure. Fixed too, and it now deletes `KV.summaries` as well, since in the ordering where the summarizer wins the lock there is a summary to remove.

The other two — `remember.ts#L294-296` and `auto-forget.ts#L197-199` — I'm declining, because they leave the session alive and self-heal. The resurrected chunk keys are content-addressed on the observations that were just deleted, so they are no longer live chunks and the next run's prune reclaims them; the stored `sourceFingerprint` no longer matches the live set, so that same run regenerates the summary. The whole-session case is different in kind precisely because nothing self-heals: every later `mem::summarize` bails at `session_not_found` before reaching any cleanup, so the orphaned rows are permanent.

**Overlap coverage — four tests, each verified to fail first.** One for the coarse window (delete landing during the provider call) and three pinning the delete to the window between the existence re-check and the write, one per fixed path. The three gate on the handler's second `KV.sessions` read rather than counting event-loop ticks, so the interleaving is deterministic rather than timing-dependent — my first attempt passed without the fix because it was not actually reaching that window.

Also spread the real module in the five test files that stub `keyed-mutex`, so a new export does not silently become `undefined`.

**On the comment nitpick from the follow-up review — partly taken, in `e31843b`.** I don't agree the whole set explains what the code does; four of the five said *why* the design is not the obvious one, which is the case the guideline exempts. But the density was 37% of added source lines against a ~9% baseline for these files, and the cause was repetition: the same lock rationale restated at four call sites. It now lives once at `sessionWriteLockKey`, where the key is defined. The call sites in `remember.ts` and `summarize.ts` carry nothing beyond the name.

Three site-specific ones stay, because each guards a wrong fix: that `evict.ts`'s new `KV.summaries` delete is not dead code, that `export-import.ts`'s absent one is covered by the sweep that lists after it, and that the summarizer's existence check is not redundant with the one on entry. Now 24%.

Full suite 1732 pass / 1 skip; `tsc` unchanged at the 30 pre-existing errors.

DanielCarmingham:
On the summary-cache invalidation race — the dangerous half is now closed by `b7e084c`, and I want to be precise about what remains rather than claim the whole thing is fixed.

`sessionWriteLockKey(sessionId)` is a second lock that both the summary write and the whole-session deletion paths take. The summary write re-checks the session still exists under that lock, and discards the summary (reclaiming its chunks) if it does not:

- `summarize.ts:514` — the write
- `remember.ts:319` — whole-session `mem::forget`
- `evict.ts:178` — stale-session eviction
- `export-import.ts:318` — replace-strategy import

That covers the case that actually corrupted state: a whole-session delete landing during a multi-second provider call used to be silently undone by the write that followed it, and nothing reclaimed the result, because every later run bailed at `session_not_found` before reaching cleanup.

Three cleanup sites are still outside that lock, all of them **partial** deletions where the session survives — `auto-forget.ts:198`, `evict.ts:306`, `remember.ts:296`. I left those deliberately. The exposure there is bounded to one cycle and self-corrects:

- Chunks are contiguous slices from index 0, so deleting an observation only shifts boundaries at and after its own chunk. Entries written back for earlier chunks are still live and still correct.
- Entries written back for shifted chunks carry keys that are no longer live, so they can never be *hit* — `chunkFingerprint` is content-addressed.
- The prune at `summarize.ts:178-188` deletes every stored entry outside the current run's live key set, so those dead entries are gone on the next run.
- The session-level `sourceFingerprint` differs after any deletion, so the next turn re-summarizes rather than serving the pre-deletion summary.

So the worst case for the unlocked paths is a handful of unreachable cache entries surviving until the next summarize, not a stale summary or an undone delete. Extending the write lock to cover partial deletions would serialize the hourly `auto-forget` sweep and the eviction pass against every in-flight summarize on the same session, which costs more than it buys.

Happy to take the stricter version if you'd rather have the invariant be unconditional.

PR #1275, opened 2026-08-28
fix(openai): retry with max_completion_tokens when the API rejects max_tokens

### LINKED ISSUE #1219: Bug: v0.9.29 default OpenAI model (`gpt-5.6-luna`) is a reasoning model, but `OpenAIProvider` sends `max_tokens` → every `mem::compress` / `mem::summarize` 400s
# Bug: v0.9.29 default OpenAI model (`gpt-5.6-luna`) is a reasoning model, but `OpenAIProvider` sends `max_tokens` → every `mem::compress` / `mem::summarize` 400s

## Summary

v0.9.29 changed the default `OPENAI_MODEL` from `gpt-4o-mini` to `gpt-5.6-luna`. `OpenAIProvider.call()` unconditionally sends `max_tokens`, which OpenAI's reasoning models reject. The result is that **every** LLM call fails with HTTP 400 for any user on the OpenAI provider who has not explicitly set `OPENAI_MODEL`, which repeatedly trips the circuit breaker.

The `max_tokens` usage is not itself new — it was in 0.9.28 too — but it was latent, because the old default model accepted the parameter. The default-model bump is what turned it into a hard failure.

## Impact

Affects any OpenAI-provider user on default config after upgrading to 0.9.29. `AGENTMEMORY_AUTO_COMPRESS=true` makes it continuous, since compression runs on every observation batch.

Memory capture keeps working (synthetic compression still indexes observations), so the failure is silent from the user's perspective — it degrades LLM-backed compression and summarization without any visible error.

## Reproduction

Default config, `OPENAI_API_KEY` set, `OPENAI_MODEL` unset. Reduced to a direct API call:

```console
$ curl -s https://api.openai.com/v1/chat/completions \
    -H "Authorization: Bearer $OPENAI_API_KEY" -H "Content-Type: application/json" \
    -d '{"model":"gpt-5.6-luna","messages":[{"role":"user","content":"hi"}],"max_tokens":4096}'
{
  "error": {
    "message": "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.",
    "type": "invalid_request_error",
    "param": "max_tokens",
    "code": "unsupported_parameter"
  }
}
```

Same request with `max_completion_tokens` returns 200.

## Observed behaviour

From `GET /agentmemory/health` on an affected install, ~1h after upgrading:

`successCount` was frozen at the pre-upgrade value while `failureCount` climbed monotonically — every call after the upgrade failed. The circuit breaker cycled open → half-open → retry → 400 → reopen, roughly every 5 minutes.

After pinning `OPENAI_MODEL=gpt-4.1-mini` and restarting: circuit `closed`, `failures: 0`, `successCount` climbing again with `failureCount` flat.

## Environment

- agentmemory 0.9.29 (upgraded from 0.9.28)
- iii-engine 0.11.2 (pinned), iii-sdk 0.11.2
- Node 24.13.0, macOS 15 (arm64)
- Provider: OpenAI, `OPENAI_MODEL` unset (default), `MAX_TOKENS` unset (default 4096)
- `AGENTMEMORY_AUTO_COMPRESS=true`, `AGENTMEMORY_INJECT_CONTEXT=true`, `GRAPH_EXTRACTION_ENABLED=true`

dKaulig:
Fixes #1219.

## What

`OpenAIProvider.call()` sends `max_tokens` unconditionally. Reasoning models — the gpt-5 family and the o-series — reject it:

```
400 Unsupported parameter: 'max_tokens' is not supported with this model.
    Use 'max_completion_tokens' instead.
```

Since 0.9.29 defaults `OPENAI_MODEL` to `gpt-5.6-luna`, that is every LLM call on default config, as #1219 documents. It is not Azure-specific, but Azure makes it easy to hit even with an explicit model: deployment names are user-chosen, so nothing about the configured string reveals which spelling the endpoint wants.

## Why detection rather than a config flag or a model allowlist

- A flag pushes a compatibility detail onto users who cannot reasonably know the answer in advance, and it would need setting again for every new deployment.
- A model-name allowlist is guesswork on Azure (`OPENAI_MODEL` is the deployment name) and goes stale with every model release.
- The API already states which parameter it wants. Reading that is exact and needs no maintenance.

So: send `max_tokens`; if a 400 names `max_completion_tokens`, switch and retry once. The provider keeps the learned spelling for its lifetime, so the extra round trip happens once per process rather than once per request. Endpoints that accept `max_tokens` never return that 400 and never see a second request — the existing path is unchanged.

## How to verify

`test/openai-max-completion-tokens.test.ts` covers four behaviours:

- `max_tokens` is sent by default
- a 400 naming `max_completion_tokens` triggers exactly one retry with the other spelling
- the learned spelling persists, so a later call costs one request, not two
- unrelated 400s (e.g. `context_length_exceeded`) do not retry

Also exercised end to end against an Azure OpenAI `gpt-5.4-mini` deployment: before the change every `mem::compress` failed with the 400 above; after it, compression returns normally with no other request-body change.

```
npm test          # 1715 passing (1711 before, +4 here)
npm run build     # clean
```

## Note on scope

#1219 also raises whether `gpt-5.6-luna` is the right default. That is a separate decision and is deliberately untouched here — this PR makes the provider work with whatever model is configured, including the current default.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility with APIs that require the `max_completion_tokens` parameter.
  - Automatically retries requests when the initial token-limit parameter is rejected.
  - Remembers the accepted parameter format to avoid unnecessary retries on future requests.
  - Preserves existing behavior for unrelated request errors.
  - Ensures retries share the original request’s timeout budget.

- **Tests**
  - Added coverage for fallback behavior, remembered configuration, timeout handling, and error handling.

### DIFF
--- a/src/providers/openai.ts
+++ b/src/providers/openai.ts
@@ -54,6 +54,13 @@ export class OpenAIProvider implements MemoryProvider {
   private timeoutMs: number;
   private isAzure: boolean;
   private azureApiVersion: string;
+  // Newer OpenAI models (gpt-5 family, o-series) reject `max_tokens` and
+  // require `max_completion_tokens` instead. Which spelling an endpoint wants
+  // cannot be derived from the model string: Azure deployment names are
+  // user-chosen and OpenAI-compatible proxies differ. So start with the widely
+  // supported spelling and latch onto the other one the first time the API
+  // asks for it — see the 400 handling in `call()`.
+  private tokenLimitParam: "max_tokens" | "max_completion_tokens" = "max_tokens";
 
   constructor(apiKey: string, model: string, maxTokens: number, baseURL?: string) {
     this.apiKey = apiKey;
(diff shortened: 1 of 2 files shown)


dKaulig:
Thanks — the timeout finding was correct and is fixed in fd3dce6.

**Shared budget.** The retry was getting a fresh `this.timeoutMs`, so a slow initial rejection could stretch one `call()` to nearly twice the configured bound. There is now a single deadline captured before the first send; each send gets only the remaining time, and no request goes out when the budget is spent.

I added a test that pins the contract rather than the shape: a 200ms rejection followed by a hanging retry must fail at roughly 300ms total. Measured against the previous per-request budget it came out at 518ms, so the test does catch the regression it describes.

**`response.clone()`.** Removed, and the fix fell out of the same change: the error body is now read once and serves both the retry decision and the error message. Nothing is teed, so there is no unread branch when the retry replaces the response.

**Mocking `iii-sdk` in the test file** — skipping this one. I could not find the convention it refers to: no test file in the repo mocks `iii-sdk` (0 of 162), including the existing provider tests `minimax-provider.test.ts`, `agent-sdk-provider.test.ts`, `local-embedding-provider.test.ts` and `clip-embedding-provider.test.ts`. `OpenAIProvider` does not touch the SDK or the KV — it is constructed directly and only `fetch` is stubbed, matching how `minimax-provider.test.ts` is written. Happy to add it if there is a guideline I have missed.

`npm test` 1716 passing, `npm run build` clean.

dKaulig:
Thanks — two of the three are in 504c727, the third I would rather split out. Taking them in order.

**Concurrent calls sharing the latch.** Correct, and reachable: the compression pipeline issues these concurrently, so this is not only a contrived-load case. Each call now captures what it sent and tests that. The regression test holds both requests until both are in flight, then releases them together; before the change the second one threw its 400 unretried.

I did not take the second half of the proposed fix, though — latching only after `response.ok`. The evidence that the endpoint wants the other spelling is the 400 naming it, and a retry that fails for an unrelated reason (429, a timeout) has not disproved that. Deferring the latch to a successful retry would make every subsequent call pay the extra round trip again, which is the cost the field exists to avoid. So the latch stays before the retry; only the condition moved to the local value. That is enough to fix the race, since the race was in the *test*, not the assignment.

**Comments.** Trimmed. The field and deadline comments keep only what is not derivable from the source — that the accepted spelling cannot be inferred from the model string, and that the deadline spans the retry — and the "read once" note went out with the clone it described. The two lines added for the concurrency point are why, not what.

**Error-body reads outside the timeout.** The finding is right: `fetchWithTimeout` clears its timer once headers arrive, so `response.text()` and the `response.json()` below it are unbounded, and a server that sends headers then stalls hangs the call.

I am not fixing it here. It predates this PR — the `json()` read has always been outside the budget — and the fix is not local: bounding body reads properly means threading the deadline through `_fetch.ts` for every raw-fetch provider that shares the helper (minimax, openrouter, gemini, openrouter-embed), otherwise only OpenAI gets the guarantee and the contract stays uneven. That is a change worth its own PR and its own tests, not a rider on a token-parameter fix.

Happy to open it if you want it tracked; say the word and I will file the issue.

PR #1277, opened 2026-08-28
fix(resilient): content filter rejections must not open the circuit breaker

### LINKED ISSUE #1276: Bug: Azure content filter rejections open the circuit breaker, failing unrelated compressions
## Summary

`ResilientProvider` counts every thrown error as a provider failure, including content filter rejections. Azure OpenAI content filters reject individual prompts, so three filtered observations inside the 60s failure window open the circuit breaker — and then every *other* compression fails with `circuit_breaker_open` for the 30s recovery timeout. A single awkward file takes an unrelated batch of observations down with it.

## Why this is not a rare edge case

Azure's Prompt Shields flags input that *looks like* a jailbreak attempt. A memory system that compresses arbitrary tool output feeds it exactly that kind of material without any of it being an attack.

Measured against a `gpt-5.4-mini` deployment with the default content filter, using the real prompt shape that `buildCompressionPrompt()` produces:

| Payload | Result |
|---|---|
| 6 KB of real source code | 200 — protected-material for code does not fire |
| error log containing `kill` / `fatal` / `attack` | 200 |
| credentials in tool output | 200 |
| tool output containing injection-style text | **400 `content_filter`** |
| a `SECURITY.md` that merely *describes* prompt injection | **400 `content_filter`** |

The last row is the point: legitimate documentation is enough. Any repo with a page about prompt security produces these.

The rejection is unambiguous — every severity category is `safe`, only the jailbreak shield fires:

```json
{"error":{"code":"content_filter","status":400,
  "innererror":{"code":"ResponsibleAIPolicyViolation",
    "content_filter_result":{
      "hate":{"filtered":false,"severity":"safe"},
      "jailbreak":{"detected":true,"filtered":true},
      "self_harm":{"filtered":false,"severity":"safe"},
      "sexual":{"filtered":false,"severity":"safe"},
      "violence":{"filtered":false,"severity":"safe"}}}}}
```

## Reproduction

With `AGENTMEMORY_AUTO_COMPRESS=true` against an Azure OpenAI deployment with default content filtering, POST six observations to `/agentmemory/observe` — five whose `tool_output` contains injection-style text, then one harmless.

Observed in `stderr.log`:

```
Compression failed {"obsId":"…","error":"OpenAI API error (400): …content_filter…"}   ×3
Compression failed {"obsId":"…","error":"circuit_breaker_open"}                       ×2
Compression failed {"obsId":"…","error":"circuit_breaker_open"}   ← the harmless one
```

Expected: five failures and one successful compression. The sixth observation has nothing wrong with it.

Two things worth knowing when reproducing this:

- The hook field is `tool_output`. A payload using a different key is silently dropped, the prompt never reaches the provider, and the run looks like a pass.
- Observation dedup hashes `(sessionId, toolName, toolInput)` in memory, so a second run with the same command returns `{"deduplicated": true}` without calling the provider. Vary the session id and the command per run.

## Expected behaviour

A content filter rejection describes the payload, not the health of the provider. The call should fail; the breaker should stay closed so unrelated observations keep compressing.

## Environment

- agentmemory 0.9.29
- Node 24.19.0, macOS 26.6 (arm64)
- Provider: Azure OpenAI (`gpt-5.4-mini` deployment), default content filter configuration
- `AGENTMEMORY_AUTO_COMPRESS=true`, `GRAPH_EXTRACTION_ENABLED=true`, `CONSOLIDATION_ENABLED=true`

Happy to send a PR — I have the fix and tests ready.

dKaulig:
Fixes #1276.

## What

`ResilientProvider.call()` calls `recordFailure()` on every throw. Azure content filter rejections are throws, so three of them inside the failure window open the breaker and the next 30s of compressions fail with `circuit_breaker_open` — including observations that have nothing wrong with them.

A content filter rejection is a property of one payload, not a health signal about the provider. This skips `recordFailure()` for it. The call still fails; only the blast radius changes.

## Why this fires often in practice

Azure's Prompt Shields flags input that looks like a jailbreak attempt, and a memory system compresses arbitrary tool output. Measured against a `gpt-5.4-mini` deployment with the default filter, in the prompt shape `buildCompressionPrompt()` actually produces: source code, error logs and credentials all pass, but injection-style text is rejected — and so is a `SECURITY.md` that merely *describes* prompt injection. Ordinary documentation is enough to trigger it. Details and the full `content_filter_result` are in #1276.

## How to verify

`test/resilient-content-filter.test.ts`:

- `isPayloadRejection()` recognises `content_filter` / `ResponsibleAIPolicyViolation` and does not misclassify 503s, timeouts or socket errors
- five consecutive filter rejections leave the breaker `closed`
- an unrelated call after three filtered ones succeeds — this is the regression; it previously threw `circuit_breaker_open`
- three genuine provider failures still open the breaker, and the next call is short-circuited

Confirmed end to end as well: five filtered observations followed by a harmless one now yield five `content_filter` failures and one successful compression, with no `circuit_breaker_open` in between. Before the change the harmless one failed too.

```
npm test          # 1716 passing (1711 before, +5 here)
npm run build     # clean
```

## Scope

Deliberately narrow. There is a reasonable argument that no 4xx should count toward the breaker, since none of them indicate the provider is unhealthy — but that is a larger behavioural change, and the filter case is the one I have evidence for. `isPayloadRejection()` is exported so that rule has somewhere to grow if you want it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Azure OpenAI content-filter rejections are now correctly distinguished from provider failures.
  * Repeated content-filter rejections no longer trigger the resilience circuit breaker.
  * Genuine provider failures, including timeouts, outages, and gateway errors, continue to activate failure protection.

### DIFF
--- a/src/providers/resilient.ts
+++ b/src/providers/resilient.ts
@@ -1,24 +1,9 @@
 import type { MemoryProvider, CircuitBreakerState } from "../types.js";
 import { CircuitBreaker } from "./circuit-breaker.js";
 
-/**
- * A rejection that means "this particular payload is not acceptable" rather
- * than "the provider is unhealthy".
- *
- * Azure OpenAI content filters are the motivating case. Prompt Shields flags
- * tool output that merely *looks* like a jailbreak — a README describing
- * prompt injection, a security test fixture, an error log quoting user input:
- *
- *   400 {"error":{"code":"content_filter", ...
- *        "innererror":{"code":"ResponsibleAIPolicyViolation",
- *        "content_filter_result":{"jailbreak":{"detected":true,"filtered":true}}}}}
- *
- * Counting those as provider failures means three filtered observations inside
- * the failure window trip the breaker, and every *other* compression then
- * fails with `circuit_breaker_open` for the recovery timeout. One awkward file
- * costs a batch of unrelated observations. The provider is answering fine, so
- * leave the breaker closed and let just that one call fail.
- */
+// Azure Prompt Shields rejects ordinary material — a SECURITY.md describing
+// prompt injection is enough — so these arrive often enough to trip the
+// breaker and take unrelated compressions down with them. See #1276.
 export function isPayloadRejection(err: unknown): boolean {
   const message = err instanceof Error ? err.message : String(err);
   return /content_filter|ResponsibleAIPolicyViolation/.test(message);
--- a/src/providers/resilient.ts
+++ b/src/providers/resilient.ts
@@ -4,9 +4,14 @@ import { CircuitBreaker } from "./circuit-breaker.js";
 // Azure Prompt Shields rejects ordinary material — a SECURITY.md describing
 // prompt injection is enough — so these arrive often enough to trip the
 // breaker and take unrelated compressions down with them. See #1276.
+const FILTER_CODE = /content_filter|ResponsibleAIPolicyViolation/;
+// The status the provider itself recorded, before any echoed upstream body:
+// a gateway that quotes a filter rejection is still a gateway failure.
+const REJECTED_STATUS = /^[^{]*\(400\)/;
+
 export function isPayloadRejection(err: unknown): boolean {
   const message = err instanceof Error ? err.message : String(err);
-  return /content_filter|ResponsibleAIPolicyViolation/.test(message);
+  return REJECTED_STATUS.test(message) && FILTER_CODE.test(message);
 }
 
 export class ResilientProvider implements MemoryProvider {
--- a/src/providers/resilient.ts
+++ b/src/providers/resilient.ts
@@ -1,6 +1,29 @@
 import type { MemoryProvider, CircuitBreakerState } from "../types.js";
 import { CircuitBreaker } from "./circuit-breaker.js";
 
+/**
+ * A rejection that means "this particular payload is not acceptable" rather
+ * than "the provider is unhealthy".
+ *
+ * Azure OpenAI content filters are the motivating case. Prompt Shields flags
+ * tool output that merely *looks* like a jailbreak — a README describing
+ * prompt injection, a security test fixture, an error log quoting user input:
+ *
+ *   400 {"error":{"code":"content_filter", ...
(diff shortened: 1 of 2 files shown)


dKaulig:
Trimmed in the follow-up commit — though not removed entirely, so here is the reasoning in case you want it gone completely.

CONTRIBUTING draws the line at "no comments that restate what the code does; only write a comment when the *why* is non-obvious." Most of that block was indeed restatement: the failure chain is in the PR body and in #1276, and `isPayloadRejection` names itself.

The part I kept is the bit a reader cannot derive from the source: that these rejections are *frequent*, not exceptional. Azure Prompt Shields fires on a `SECURITY.md` that merely describes prompt injection, so any repo documenting prompt safety produces them routinely — and that frequency is the entire justification for special-casing them instead of letting the breaker do its job. Without it the next reader sees a narrow regex around an edge case and may reasonably delete it.

Three lines now, pointing at #1276 for the measurements. Happy to drop them if you would rather keep the file comment-free.

dKaulig:
Taken — this was a real hole, and it was reachable. Fixed in 1305be9.

`OpenAIProvider` formats failures as `OpenAI API error (${status}): ${body}`, and gateways echo upstream bodies, so a 502 quoting a filter rejection matched the old text-only test and was excused from the breaker. Repeated gateway failures would therefore leave it closed — the exact outage it exists to catch.

The status now has to agree, matched before the JSON body starts so the provider's own status counts and an echoed one does not:

```ts
const REJECTED_STATUS = /^[^{]*\(400\)/;
```

Two regression tests: a 502 whose body quotes both filter tokens is not a payload rejection, and three of them open the breaker.

One note on the diff, since it changes an assertion rather than only adding: `isPayloadRejection(new Error("ResponsibleAIPolicyViolation"))` used to assert `true` — the code alone was sufficient. That is precisely the contract this narrows, so it now asserts `false`, and a 400-shaped message covers what it used to. Flagging it so it does not read as a test bent to fit the code.

PR #1280, opened 2026-08-29
feat(graph): skip mem::graph-extract when the observation set is unchanged

### LINKED ISSUE #978: Batch/throttle graph extraction instead of triggering an LLM call on every Stop hook
`mem::graph-extract` (and the LLM-backed `provider.compress()` call it makes) is only wired to `event::session::stopped`, which the Claude Code plugin fires via the `Stop` hook at the end of *every assistant turn* — not just when a logical session truly ends. With `GRAPH_EXTRACTION_ENABLED=true`, this means an Anthropic API call fires every few minutes during active use, even though the graph itself is merged incrementally and cheap.

Could graph extraction support a batched/throttled mode — e.g. a debounce/interval (only extract if N minutes have passed since the last extraction for a session) or an opt-in nightly cron-style batch via the existing `/agentmemory/graph/build` endpoint — so users don't pay one LLM call per turn? Right now the only way to avoid the per-turn cost is to disable graph extraction entirely, which seems like an unnecessarily blunt tradeoff between cost and graph coverage.

Noticed `GRAPH_EXTRACTION_BATCH_SIZE` already exists in `config.ts` but isn't read anywhere in the codebase — might be a leftover hook for exactly this kind of batching.

### LINKED ISSUE #1063: Session stop re-runs graph extraction over the full observation set even when nothing changed

DanielCarmingham:
## Problem

`event::session::stopped` fires on every assistant turn and hands `mem::graph-extract` the session's **entire** compressed observation set, with no chunking and no change detection. A long-running session therefore re-extracts an unchanged corpus once per turn: the full heuristic pass plus, when a provider is configured, a full LLM call over the whole set. This is the graph-side half of the token/latency burn tracked in #1063, and the repeated-work behavior described in #978.

## Fix

Gate `mem::graph-extract` on a fingerprint of the observation set, stored under a single fixed per-session key (`mem:graph:extract-marks:<sessionId>`). When the incoming set's fingerprint matches the stored mark, the function returns `{ skipped: "unchanged" }` before the heuristic/LLM pass — the cost saved is the whole pass, not just the LLM call.

Design decisions worth calling out:

- **Content-aware, not identity-only.** The fingerprint hashes id + title + narrative length per observation. An identity-only fingerprint (sorted ids) would false-skip forever after a same-id content rewrite — bulk `import-jsonl` with strategy ≠ `skip`, or replay over an edited transcript, both keep ids stable while changing content. Narrative *length* rather than the body keeps this cheap on the hot Stop path (sessions can run tens of thousands of observations).
- **Sorted by id before hashing.** The set originates from `kv.list`, whose order is not contractually stable; an unsorted fingerprint would churn on a permuted-but-unchanged set and never hit the gate.
- **The mark records whether the LLM pass ran.** A heuristic-only mark (the default keyless install) never suppresses a later run once an LLM becomes available; a mark written with `llm: true` is a ceiling, so a heuristic-only re-run against the same set can skip. Without this, sets extracted while `GRAPH_EXTRACTION_ENABLED` was off would stay permanently skipped after the user enables it — and a second `/graph/build` (the documented recovery for an empty graph) would hit its own marks and report success with zero nodes added.
- **The mark is written on clean attempt completion, not persist success.** A provider timeout/rate-limit on exactly the biggest sessions must not mark the set done and permanently drop that turn's LLM extraction.
- **A batch spanning multiple sessions skips the gate** rather than guessing a scope — degrading to "always extract", never to "never extract". No current caller sends one.

## Reclamation

`KV.graphExtractMarks` is a new per-session scope, so this PR also owns its cleanup (`deleteGraphExtractMarks`):

- **Whole-session deletion** — `mem::forget` (remember.ts), stale-session eviction (evict.ts), replace-strategy import (export-import.ts) — deletes the mark alongside the session; nothing can ever reach it afterwards.
- **Per-observation deletion** — evict.ts low-importance/project-cap paths, `mem::auto-forget`'s low-value sweep, `mem::forget` with explicit `observationIds` — flushes once per touched session, since removing an observation invalidates the fingerprint recorded against the old membership.

## Relationship to #1269

Independent — neither depends on the other. Both touch the deletion paths in `evict.ts`/`auto-forget.ts`, so whichever merges second will have a small (~3-line) conflict in those files; happy to rebase this one if #1269 lands first, or vice versa.

## Tests

17 new tests: 9 on the gate in `test/graph.test.ts` (skip on unchanged set, re-extract on new observation and on same-id content rewrite, permuted-set stability, no mark after LLM failure, zero-yield marking, heuristic-only ceiling semantics both directions, multi-session batch bypass), plus one reclamation test per deletion path across `test/evict.test.ts`, `test/remember-forget-audit.test.ts`, `test/auto-forget.test.ts`, `test/export-import.test.ts`, including dry-run-does-not-flush coverage.

Full suite: 1728 passed / 1 skipped. `tsc --noEmit` unchanged at the 30 pre-existing errors (none in touched files).

Closes #1063. Closes #978.

## Summary by CodeRabbit

* **Performance**
  * Graph extraction now skips unchanged observation sets, reducing unnecessary processing.
  * Updates to observation content or metadata correctly trigger re-extraction.

* **Bug Fixes**
  * Extraction records are cleaned up when observations or sessions are forgotten, evicted, automatically removed, or replaced during import.
  * Dry-run eviction leaves existing records unchanged.

* **Reliability**
  * Improved handling of failed extractions, empty results, and provider availability.

### DIFF
--- a/src/functions/auto-forget.ts
+++ b/src/functions/auto-forget.ts
@@ -6,6 +6,7 @@ import { recordAudit } from "./audit.js";
 import { deleteAccessLog } from "./access-tracker.js";
 import { getSearchIndex, vectorIndexRemove, flushIndexSave } from "./search.js";
 import { logger } from "../logger.js";
+import { deleteGraphExtractMarks } from "./graph.js";
 
 const MS_PER_DAY = 24 * 60 * 60 * 1000;
 const CONTRADICTION_THRESHOLD = 0.9;
@@ -144,6 +145,8 @@ export function registerAutoForgetFunction(sdk: ISdk, kv: StateKV): void {
         }
       }
 
+      const touchedSessionIds = new Set<string>();
+
       const sessions = await kv.list<Session>(KV.sessions);
       const obsPerSession: CompressedObservation[][] = [];
       for (let batch = 0; batch < sessions.length; batch += 10) {
@@ -172,6 +175,7 @@ export function registerAutoForgetFunction(sdk: ISdk, kv: StateKV): void {
                 deletedOk = false;
               }
               if (deletedOk) {
+                touchedSessionIds.add(sessions[i].id);
                 if (obs.imageData) await decrementImageRef(kv, sdk, obs.imageData);
                 if (obs.imageRef && obs.imageRef !== obs.imageData) {
                   await decrementImageRef(kv, sdk, obs.imageRef);
@@ -188,6 +192,11 @@ export function registerAutoForgetFunction(sdk: ISdk, kv: StateKV): void {
             }
           }
         }
+      }
+
+      // Empty on a dryRun pass, so no dryRun guard is needed.
+      for (const sessionId of touchedSessionIds) {
+        await deleteGraphExtractMarks(kv, sessionId);
       }
 
       if (!dryRun && (result.ttlExpired.length > 0 || result.lowValueObs.length > 0)) {
--- a/src/functions/auto-forget.ts
+++ b/src/functions/auto-forget.ts
@@ -194,7 +194,10 @@ export function registerAutoForgetFunction(sdk: ISdk, kv: StateKV): void {
         }
       }
 
-      // Empty on a dryRun pass, so no dryRun guard is needed.
+      // Empty on a dryRun pass, so no dryRun guard is needed. Kept
+      // sequential rather than Promise.all: this runs hourly across every
+      // session, and the file-based KV adapter stalls under that much
+      // concurrent fan-out (upstream #1127).
       for (const sessionId of touchedSessionIds) {
         await deleteGraphExtractMarks(kv, sessionId);
       }
--- a/src/functions/evict.ts
+++ b/src/functions/evict.ts
@@ -64,7 +64,7 @@ async function recoverStaleSession(
       function_id: "event::session::stopped",
       // Suppress the per-session consolidation fan-out: eviction runs a
       // single corpus-wide consolidation pass after all recoveries instead.
-      payload: { sessionId, skipConsolidation: true },
+      payload: { sessionId, skipConsolidation: true, awaitGraphExtract: true },
     });
     if (!isValidRecoveryResult(result)) {
       logger.warn("Stale session recovery failed", {
@@ -290,7 +290,10 @@ export function registerEvictFunction(sdk: ISdk, kv: StateKV): void {
         }
       }
 
-      // Empty on a dryRun pass, so no dryRun guard is needed.
+      // Empty on a dryRun pass, so no dryRun guard is needed. Kept
+      // sequential rather than Promise.all: a sweep can touch hundreds of
+      // sessions, and the file-based KV adapter stalls under that much
+      // concurrent fan-out (upstream #1127).
       for (const sessionId of touchedSessionIds) {
         await deleteGraphExtractMarks(kv, sessionId);
       }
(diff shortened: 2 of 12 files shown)


DanielCarmingham:
All four addressed in `4cb7a30`. Two were real bugs.

**Hash all extraction inputs — fixed, and it was the more serious of the two.** `hashObservationSet` covered `id`, `title` and `narrative.length`, so `type`, `concepts` and `files` were not merely approximated, they were unhashed entirely. Since `extractGraphHeuristics` builds its file and concept nodes straight from `obs.files`/`obs.concepts`, a retag or a file remap changed the graph while leaving the fingerprint identical — the gate returned `skipped: "unchanged"` over a graph that no longer matched its observations.

It now hashes the union of what both passes actually read: `extractGraphHeuristics` uses `id`/`files`/`concepts`, `buildGraphExtractionPrompt` renders `title`/`narrative`/`concepts`/`files`/`type`. Fields are length-prefixed so `["a","bc"]` and `["ab","c"]` cannot collide. Four regressions added — same-length narrative rewrite, concepts-only, files-only, type-only — each verified failing against the previous hash and passing after.

**Audit the zero-yield mark — fixed.** The branch wrote a mark and returned without `recordAudit`, even though that mark gates every later run. It now records the audit first and skips the mark entirely if the audit throws, so an unaudited session simply re-extracts. Test added, verified failing without the change.

**Stale-session recovery race — fixed.** You were right that `event::session::stopped` starts extraction with `TriggerAction.Void()` and does not await it. The payload now carries `awaitGraphExtract`, set only by eviction's recovery path, alongside the `skipConsolidation` flag that same caller already sends. Ordinary Stop events stay fire-and-forget.

**Parallel cleanup — declining, deliberately.** These loops fan out across every session an eviction or auto-forget pass touched, which can be hundreds, each doing its own KV round trip. Unbounded `Promise.all` at that width against the `file_based` adapter is the stall described in #1127, and every one of those calls inherits the 180s worker-level `invocationTimeoutMs` until #1128 lands. I declined the same suggestion on #1269 for the same reason and would rather be consistent than fast here. The reasoning is now recorded at both sites so the next reader does not have to rediscover it.

Also cut the added comments from 45% to 32% of added source lines.

Issue #931, opened 2026-06-14
Local embedding provider (@xenova/transformers) not detected in v0.9.27 — smart-search returns 0 semantic hits


ksteve0327:
## Summary

`@xenova/transformers` is installed in the server's own `node_modules`, but smart-search still returns **0 semantic hits** and the server log shows **no embedding-provider detection at all** (not even an error). It appears the local embedding provider is never activated, so search silently falls back to BM25-only.

## Environment

- agentmemory: **v0.9.27** (global npm install)
- iii-engine: v0.11.2
- Node.js: **v22.22.3** (Homebrew `node@22`)
- OS: macOS (darwin 25.3.0, arm64 / Apple Silicon)
- Server run via a launchd LaunchAgent (always-on), bound to localhost:3111

## What I did

1. `npm install -g @xenova/transformers` — installed fine.
2. Since the global install wasn't picked up, I also installed it **directly inside the server package folder**:
```
   cd "$(npm root -g)/@agentmemory/agentmemory"
   npm install @xenova/transformers
   ls node_modules/@xenova   # -> transformers  (confirmed present)
```
3. Restarted the server (launchctl unload/load) and waited ~30s.
4. Ran `agentmemory demo`.

## Expected

The demo query `"database performance optimization"` should match the seeded **N+1 query fix** via semantic search (this is the demo's own advertised behavior).

## Actual

```
"jwt auth middleware"               -> 5 hit(s)   (BM25 keyword match works)
"database performance optimization" -> 0 hit(s)   (no semantic match)
"rate limiting"                     -> 2 hit(s)   (BM25 keyword match works)
```

Keyword searches return hits, but the semantic-only query stays at 0 across every attempt.

## Server log

`grep -iE "embed|xenova|transformer|vector|provider" ~/.agentmemory/server.log` returns **nothing** — there is no log line about embedding provider detection, model loading, or any embedding-related error. The only provider-related lines are about the LLM summarizer:

```
No LLM provider key found (...). LLM-backed compression and summarization are DISABLED — using no-op provider.
```

(I understand that's the separate LLM summarize layer, not embeddings.)

## Question

With `@xenova/transformers` present in the server's `node_modules`, **what enables the local embedding provider in v0.9.27?** Is there an env var (e.g. something I should set in `~/.agentmemory/.env` or pass to the process) or a config flag required to activate vector search? The README mentions `npm install @xenova/transformers` as sufficient for local embeddings, but in my setup it never gets detected.

[...]

PR #1283, opened 2026-08-29
fix(embedding): fall back to the local provider instead of none

### LINKED ISSUE #395: Bug: Local embedding stuck at "none" (BM25-only) in v0.9.13 despite correct xenova config

DanielCarmingham:
## Problem

On a keyless install, `detectEmbeddingProvider` returns `null` unless `EMBEDDING_PROVIDER` is set explicitly. That leaves `currentEmbeddingProvider` unset, so `vectorIndexAddGuarded` and `indexRecords` silently no-op for **every** observation — semantic search returns zero hits forever, with no error surfaced anywhere. This is the root cause behind the zero-coverage reports in #395 and the missing-embeddings half of #931: the local provider existed but was an explicit opt-in nobody was told about, not the keyless default.

## Fix

- **Fall back to `local`** (Xenova/all-MiniLM-L6-v2, 384-dim, on-device, no key required) when no cloud key is configured, instead of returning `null`. The local provider's optional `@huggingface/transformers` dependency resolves lazily on the first `embed()` call, so a missing package surfaces as a per-write warning from the existing vector-index guards — not a boot failure.
- **`EMBEDDING_PROVIDER=none` becomes the explicit opt-out**, preserving a deliberate BM25-only mode.
- `.env.example` updated: the "without an embedding key, BM25-only" claim is stale under this change, and `none` joins the documented value list.

**Deliberate behavior change worth your sign-off:** a keyless install now downloads the local embedding model on the first embedding request (network access required at that moment) unless a cloud key is configured or `EMBEDDING_PROVIDER=none` opts out. If you'd rather keep opt-in semantics, say so and I'll rework this into a louder warning instead — but as-is, the default install ships with semantic search silently broken.

## Tests

`test/embedding-provider.test.ts` previously asserted the defect directly ("returns null when no API keys are set" for `createEmbeddingProvider()`); that test now expects a `LocalEmbeddingProvider` instance, with companion tests for the `none` opt-out and a new `detectEmbeddingProvider` block covering the fallback, explicit-provider precedence, cloud-key precedence, and opt-out.

Full suite: 1716 passed / 1 skipped. `tsc --noEmit` unchanged at the 30 pre-existing errors (none in touched files).

A follow-up PR adds boot-time provider visibility (probe + health reporting) and fixes the indexing gate for empty-narrative observations — kept separate so this behavior change can be discussed on its own.

Closes #395. Refs #931.

## Summary by CodeRabbit

* **New Features**
  * Local on-device embeddings are now enabled by default when no cloud provider is configured.
  * Semantic search and vector indexing remain available without external API keys.
  * Added an explicit `none` option to disable embeddings while retaining BM25 and structural graph search.

* **Documentation**
  * Updated environment configuration guidance to reflect provider detection order and available options.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -66,16 +66,17 @@
 # 2. Embedding provider — auto-detected, override via EMBEDDING_PROVIDER
 # -----------------------------------------------------------------------------
 #
-# Without an embedding provider, vector embeddings are disabled. `mem::search`
-# uses BM25; `mem::smart-search` can also fuse structural graph matches when
-# graph data exists. Remote
-# provider detection order: EMBEDDING_PROVIDER override → GEMINI_API_KEY →
-# OPENAI_API_KEY → VOYAGE_API_KEY → COHERE_API_KEY → OPENROUTER_API_KEY →
-# BM25-only. Local embeddings are an explicit opt-in, not the keyless default.
-# `EMBEDDING_PROVIDER=local` uses Xenova/all-MiniLM-L6-v2 (384-dim) on-device;
-# the first embedding request downloads the model and needs network access.
+# Without an embedding key, agentmemory falls back to the local on-device
+# provider by default instead of running BM25-only. Detection order:
+# EMBEDDING_PROVIDER override → GEMINI_API_KEY → OPENAI_API_KEY →
+# VOYAGE_API_KEY → COHERE_API_KEY → OPENROUTER_API_KEY → local
+# (Xenova/all-MiniLM-L6-v2, 384-dim, on-device; the first embedding
+# request downloads the model and needs network access).
+# EMBEDDING_PROVIDER=none disables embeddings explicitly: `mem::search`
+# uses BM25, and `mem::smart-search` can still fuse structural graph
+# matches when graph data exists.
 
-# EMBEDDING_PROVIDER=local                       # local | openai | voyage | cohere | gemini | openrouter
+# EMBEDDING_PROVIDER=local                       # none | local | openai | voyage | cohere | gemini | openrouter
 
 # VOYAGE_API_KEY=pa-...                          # Optimised for code embeddings
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -271,6 +271,9 @@ export function detectEmbeddingProvider(
 ): string | null {
   const source = env ?? getMergedEnv();
   const forced = source["EMBEDDING_PROVIDER"];
+  // #395: an explicit opt-out has to stay reachable now that local is
+  // the fallback rather than null.
+  if (forced === "none") return null;
   if (forced) return forced;
 
   if (source["GEMINI_API_KEY"]) return "gemini";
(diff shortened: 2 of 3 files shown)


DanielCarmingham:
Trimmed in `ebff5c9`. Kept the two things the code cannot say — why an explicit `none` opt-out has to stay reachable now that the fallback is a real provider, and why falling back to `local` is safe (no key required, optional dependency resolves lazily, so a missing package surfaces as a per-write warning rather than a boot failure). Dropped the rest.

PR #1284, opened 2026-08-29
fix(embedding): boot-time provider visibility and empty-narrative indexing

DanielCarmingham:
## Problem

#931's underlying complaint is that embedding failures are invisible. Two concrete holes:

1. **A missing or broken embedding runtime surfaced only as a per-write `logger.warn`** inside the vector-index guards. On a live deployment we observed, the corpus reached **201,102 observations at 1.1% vector coverage** before anyone noticed — the installed package was the legacy `@xenova/transformers` 2.x while the code imports the renamed `@huggingface/transformers`, and not one boot line said so.
2. **`indexRecords` silently dropped observations with an empty narrative** from both the BM25 and vector indexes. Synthetic compressions legitimately have an empty narrative when the hook carried no prompt/input/output (`compress-synthetic.ts`), and the live `observe.ts` path indexes those fine — only the rebuild path lost them.

## Fix

**Boot probe (commit 1).** Resolve and warm the embedding provider once at boot with a single `embed()` call, and report the outcome — provider name and dimensions on success, the underlying error plus the BM25-only degradation on failure. Details that matter:

- **Fire-and-forget.** At that point in `main()` the shutdown handlers aren't registered yet, and a cold local-model load can download tens of MB — awaiting the probe would open a window where a rolling deploy's SIGTERM kills the process before `indexPersistence.save()` runs, losing the persisted search index. Detached, a slow cold load simply reports later; no timeout needed.
- **Reported through `logger`, not just `bootLog`.** `bootLog` reaches stderr only under `--verbose`, which a daemon (launchd/systemd) start never passes — its buffer is otherwise never read. That's precisely why hole #1 stayed invisible on the deployment that mattered. `bootLog` is kept alongside for `--verbose` parity.
- **The disabled branch prints the resolved `EMBEDDING_PROVIDER` value** instead of hardcoding `none`, so a typo'd value isn't blamed on the deliberate opt-out.
- **The probe calls `embedBatch`, not `embed`, and verifies the returned shape** (one vector of exactly `dimensions` length) — `embedBatch` is what the indexing path (`vectorIndexAddBatchGuarded`) actually uses, and that guard drops any wrong-length vector, so a wrong-shape provider would pass a bare probe yet index nothing.
- **The local provider's install error now names `@xenova/transformers` 2.x as incompatible** — the exact trap from the observed deployment.
- **`LocalEmbeddingProvider` caches the in-flight extractor load.** The probe and a BM25 rebuild's embedding queue can both hit a cold provider at once; with only a post-`await` cache, each concurrent caller kicks off its own `pipeline()` initialization (duplicate model download/memory — verified against plain Node promise semantics; vitest's module runner serializes mocked dynamic imports, so the concurrency itself isn't black-box testable there). The cached promise is evicted on rejection so a transient download failure retries instead of latching until restart — and that eviction behavior is pinned by a test that fails against the naive `??=` implementation.

**Indexing gate (commit 2).** `indexRecords` now requires only a title, with title-only text for the embedding queue when the narrative is empty — matching what `observe.ts` already does.

## Tests

Probe success/failure/never-rejects via the exported `reportEmbeddingProbeResult` (exported precisely because the call site is fire-and-forget, so awaiting it in a test is the only way to observe settlement); structural checks for the two boot lines in `main()` (which can't be invoked from a unit test — importing `src/index.ts` starts a real worker); the rewritten install error naming both packages; a wrong-shape probe failure; retry-after-failed-init for the extractor cache; and an empty-narrative observation indexing to count 1.

Full suite: 1720 passed / 1 skipped. `tsc --noEmit` unchanged at the 30 pre-existing errors (none in touched files).

One reviewer-style suggestion I investigated and declined: replacing the factory-less `vi.doMock("@huggingface/transformers")` in the package-unavailable tests with a factory that raises `ERR_MODULE_NOT_FOUND`. Vitest wraps any factory throw/rejection in its own "error when mocking a module" error, discarding the `code` the provider's mapping keys on, so the explicit simulation is not expressible; the factory-less form (the repo's existing pattern for these tests) reliably produces the module-not-found path because the optional runtime is not in devDependencies. Documented in a comment at the test site.

Independent of #1283 (local-by-default detection): the probe reports whatever provider resolves under current detection rules, and each PR merges cleanly without the other. Together they close the loop on #931's "embeddings silently absent" story.

Refs #931.

## Summary by CodeRabbit

* **New Features**
  * Added startup diagnostics for embedding providers, including provider status, successful checks, and configuration details.
  * Embedding failures now clearly fall back to BM25-only search without preventing startup.
* **Bug Fixes**
  * Observations with titles but empty narratives are now searchable.
  * Improved local embedding initialization reliability by preventing duplicate loading and retrying after failures.
  * Updated installation guidance for supported embedding packages.

### DIFF
--- a/src/functions/search.ts
+++ b/src/functions/search.ts
@@ -294,12 +294,17 @@ export async function indexRecords(
     count++
   }
   for (const obs of observations) {
-    if (!obs.title || !obs.narrative) continue
+    // #931: a synthetic compression legitimately has an empty narrative
+    // when the hook carried no prompt, input, or output
+    // (compress-synthetic.ts). Requiring one here dropped those
+    // observations from BOTH the BM25 and vector indexes, while the live
+    // observe.ts path indexed them fine using title + narrative.
+    if (!obs.title) continue
     idx.add(obs)
     await enqueue({
       id: obs.id,
       sessionId: obs.sessionId,
-      text: obs.title + ' ' + obs.narrative,
+      text: obs.narrative ? obs.title + ' ' + obs.narrative : obs.title,
       context: { kind: "observation", logId: obs.id },
     })
     count++
--- a/src/index.ts
+++ b/src/index.ts
@@ -551,35 +551,15 @@ async function main() {
     config.restPort,
   );
 
-  // #931: a broken or absent embedding runtime previously surfaced only as
-  // a per-observation logger.warn inside vectorIndexAddGuarded, so a corpus
-  // could reach six figures at ~1% vector coverage with no visible signal.
-  // Resolve and warm the provider once at boot and say what happened.
-  //
-  // Fire-and-forget: process.on("SIGINT"/"SIGTERM", shutdown) is not
-  // registered until further down this function; awaiting the probe here
-  // would delay that registration by however long the first embed call
-  // takes. A cold local-model load can mean downloading tens of MB, so on
-  // a slow or rate-limited link that could be a long wait - if a rolling
-  // deploy's SIGTERM lands in that window, there is no shutdown handler
-  // installed yet and the process dies without healthMonitor.stop()/
-  // indexPersistence.save()/sdk.shutdown() running, losing the persisted
-  // search index. Detaching this also means there is no reason to bound it
-  // with a timeout: a slow cold load simply reports later instead of being
-  // killed off and logged as a spurious failure for a provider that would
-  // otherwise have worked.
+  // Deliberately not awaited: the SIGINT/SIGTERM handlers are registered
+  // further down this function, and a cold local-model load can take long
+  // enough to download tens of MB. A deploy signal arriving in that window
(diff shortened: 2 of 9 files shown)


DanielCarmingham:
Both addressed in `b23c250`.

**Assert index contents — fixed, and you were right that the count alone was hollow.** `expect(indexed).toBe(1)` passes on the return value regardless of whether `o1` reached the index, which is precisely the regression the test exists to catch. It now asserts `getSearchIndex().has("o1")` and that a search for the title returns it.

**Comments — trimmed**, from 44% to 27% of added source lines. Kept the reason the boot probe is deliberately not awaited (the SIGINT/SIGTERM handlers are registered later in that function, and a cold model load can outlast a deploy signal, which would lose the persisted index), and the reason the probe calls `embedBatch` rather than `embed` with a shape check. Dropped the rest.

PR #1285, opened 2026-08-29
fix(health): measure heap against heap_size_limit, not heapTotal

### LINKED ISSUE #1223: Health reports critical from heapUsed/heapTotal, so /agentmemory/health returns 503 on a healthy process
`evaluateHealth` measures memory severity against `heapTotal`, which is the heap V8 has already committed, not the ceiling it is allowed to grow to. V8 sizes `heapTotal` to demand, so a busy healthy process sits near 100% of it more or less permanently.

`src/health/thresholds.ts:62-65`

```ts
const memPercent =
  snapshot.memory.heapTotal > 0
    ? (snapshot.memory.heapUsed / snapshot.memory.heapTotal) * 100
    : 0;
```

Since `heapTotal <= heap_size_limit` always holds, this can only ever over-report severity, never under-report it.

## What I saw

On a live deployment:

| | bytes |
|-|-|
| `heapUsed` | 390,771,624 |
| `heapTotal` | 404,930,560 |
| `rss` | 584,327,168 |
| `v8.getHeapStatistics().heap_size_limit` | 6,492,782,592 (6192 MB) |

That is 6.0% of the heap the process may actually use. The alert read `memory_critical_97%_rss557mb` and status was `critical`.

`GET /agentmemory/health` then returned **HTTP 503**, because `src/triggers/api.ts:274` maps `critical` to 503:

```ts
const statusCode = status === "critical" ? 503 : 200;
```

`/agentmemory/livez` returned 200 at the same moment. So an uptime monitor pointed at `/health` reads the service as down while it is serving normally. At that point the container was using 3.0 GB of an 8 GB limit, the circuit breaker was closed, CPU was 3.4%, and event loop lag was 0.1 ms.

Worth noting for anyone deploying from this repo: `deploy/railway/railway.json` sets `healthcheckPath` to `/agentmemory/livez`, so Railway's own check keeps passing and the service never restart-loops. The shipped template dodges this by accident. It is the richer `/agentmemory/health` endpoint, the one you would point an external monitor at, that reports the false 503.

## What I expected

`healthy`, and HTTP 200. Six percent of the heap limit is not memory pressure.

## How this relates to #158

#158 reported the same formula misfiring at `heapUsed` ~45 MB, `heapTotal` ~46 MB, and rss ~120 MB. The fix that landed added `memoryRssFloorBytes` (512 MB, `src/health/thresholds.ts:20`) so the alert stays quiet on small processes. That covers the case as reported, but the ratio itself was never corrected, so the bug comes back for any process whose RSS clears the floor.

There is also no way to tune around it from the outside. `src/health/monitor.ts:87` calls `evaluateHealth(snapshot)` with no config argument, so `DEFAULTS` always wins and neither the percentages nor the RSS floor are reachable from the environment.

## Repro

1. Run agentmemory under a workload that keeps RSS above 512 MB.
2. `curl -si "$AGENTMEMORY_URL/agentmemory/health" | head -1`
3. You get 503, with `memory_critical_NN%` in `health.alerts`, while the process is using a small fraction of `v8.getHeapStatistics().heap_size_limit`.

The viewer shows the same thing. `src/viewer/index.html:1582` recomputes the identical `heapUsed / heapTotal` ratio client-side, so the HEAP gauge renders red on a healthy process.

## Environment

- agentmemory 0.9.28 (latest published: 0.9.29)
- Node v22.23.2
- Platform: Railway, self-hosted from this repo's own `deploy/railway` template
- Linux container, 8 GB memory limit, single replica
- Reproduced against the deployed service, not a local dev server

I have a fix with tests and can open a PR.

DanielCarmingham:
## Problem

The health monitor's memory status measures `heapUsed / heapTotal` (#1223). `heapTotal` is V8's *current allocation*, which V8 grows on demand — a healthy process routinely runs at 85–95% of `heapTotal` while sitting far below the actual ceiling. The ratio therefore reports "degraded"/"critical" on processes that are nowhere near out of memory, and it can never distinguish "V8 hasn't grown the heap yet" from "we're about to OOM".

## Fix

Measure against `getHeapStatistics().heap_size_limit` — the hard ceiling V8 will actually enforce — captured into the snapshot as `memory.heapLimit` (optional field on `HealthSnapshot`, so existing persisted snapshots stay valid; the thresholds fall back to the old ratio when it's absent).

## Tests

Threshold-side tests for the new ratio (including the fallback when `heapLimit` is absent), plus a producer-wiring test comparing the persisted snapshot's `memory.heapLimit` against the real unmocked `getHeapStatistics().heap_size_limit` — verified to fail (`undefined` vs. the real value) when the wiring line in `monitor.ts` is deleted, since nothing else pins it.

Full suite: 1714 passed / 1 skipped. `tsc --noEmit` unchanged at the 30 pre-existing errors (none in touched files).

**Known limitation, deliberately out of scope:** `heap_size_limit` excludes native allocations (buffers, stacks, code memory), so a process under *native*-memory pressure can show a healthy heap ratio. Watching for that properly means comparing RSS against the effective process/container memory budget, and the codebase has no budget source (cgroup awareness) today — the existing `memoryRssFloorBytes` config only gates alerts from below. Worth its own issue; this PR stays scoped to removing the false-positive ratio, which made /health return 503 on healthy processes. Note the old ratio did not cover native pressure either — its numerator was heap-only too.

One operational caveat documented in the commit rather than code: in a container, V8 sizes `heap_size_limit` from host memory, not the cgroup limit — `--max-old-space-size` should be set to the container's budget for the ratio to reflect a ceiling V8 will respect. (An `.env.example` note about this rides with the upcoming configurable-thresholds PR, which owns that doc section.)

Closes #1223.

## Summary by CodeRabbit

* **Bug Fixes**
  * Health monitoring now measures memory usage against the runtime’s maximum heap limit when available.
  * Snapshots without valid heap-limit data continue using total heap size.
  * Health statuses and alerts now more accurately reflect memory pressure.

* **Tests**
  * Added coverage confirming heap-limit data is captured in health snapshots.
  * Added threshold tests for both heap-limit calculations and fallback behavior.

### DIFF
--- a/src/health/monitor.ts
+++ b/src/health/monitor.ts
@@ -1,4 +1,5 @@
 import type { ISdk } from "iii-sdk";
+import { getHeapStatistics } from "node:v8";
 import type { HealthSnapshot } from "../types.js";
 import type { StateKV } from "../state/kv.js";
 import { KV } from "../state/schema.js";
@@ -69,6 +70,8 @@ export function registerHealthMonitor(
       memory: {
         heapUsed: mem.heapUsed,
         heapTotal: mem.heapTotal,
+        // #1223: hard ceiling, see thresholds.ts for why this beats heapTotal.
+        heapLimit: getHeapStatistics().heap_size_limit,
         rss: mem.rss,
         external: mem.external,
       },
--- a/src/health/monitor.ts
+++ b/src/health/monitor.ts
@@ -70,7 +70,6 @@ export function registerHealthMonitor(
       memory: {
         heapUsed: mem.heapUsed,
         heapTotal: mem.heapTotal,
-        // #1223: hard ceiling, see thresholds.ts for why this beats heapTotal.
         heapLimit: getHeapStatistics().heap_size_limit,
         rss: mem.rss,
         external: mem.external,
--- a/src/health/thresholds.ts
+++ b/src/health/thresholds.ts
@@ -59,9 +59,15 @@ export function evaluateHealth(
     degraded = true;
   }
 
+  // #1223: prefer the V8 hard ceiling. heapTotal is demand-sized, so the
+  // old ratio reported 'critical' on a process at 10% of its real limit.
+  const heapDenominator =
+    snapshot.memory.heapLimit && snapshot.memory.heapLimit > 0
+      ? snapshot.memory.heapLimit
+      : snapshot.memory.heapTotal;
   const memPercent =
-    snapshot.memory.heapTotal > 0
-      ? (snapshot.memory.heapUsed / snapshot.memory.heapTotal) * 100
+    heapDenominator > 0
+      ? (snapshot.memory.heapUsed / heapDenominator) * 100
       : 0;
   const rss = snapshot.memory.rss ?? 0;
   const rssAboveFloor = rss >= cfg.memoryRssFloorBytes;
--- a/src/health/thresholds.ts
+++ b/src/health/thresholds.ts
@@ -59,8 +59,8 @@ export function evaluateHealth(
     degraded = true;
   }
 
-  // #1223: prefer the V8 hard ceiling. heapTotal is demand-sized, so the
-  // old ratio reported 'critical' on a process at 10% of its real limit.
+  // heapTotal is demand-sized, so measuring against it reported critical
+  // on a process at 10% of its real ceiling.
   const heapDenominator =
     snapshot.memory.heapLimit && snapshot.memory.heapLimit > 0
       ? snapshot.memory.heapLimit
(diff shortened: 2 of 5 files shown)


DanielCarmingham:
Comments trimmed in `0870424` — the three added blocks are down to the rationale the code cannot carry (why `heapTotal` is the wrong denominator, why `heapLimit` is optional).

On the Vitest mock: declining, because the suggested pattern is not this repository's convention. Of 153 test files, **53 define a local `mockSdk`/`mockKV`** and **2** call `vi.mock("iii-sdk")` — and both of those use `importOriginal` to keep the real `TriggerAction` rather than replacing it. Local fakes are the established pattern here, and the test passes against the real module, which is a stronger guarantee than a mock would give.

Happy to switch if there's a written convention I've missed — I checked `AGENTS.md` and `CLAUDE.md` and found nothing on this.

PR #1287, opened 2026-08-29
fix(evict): schedule the eviction sweep at boot

DanielCarmingham:
## Problem

`mem::evict` is fully implemented — stale sessions, low-importance observations, the `maxObservationsPerProject` cap, expired/non-latest memories — but nothing ever runs it. The boot scheduler in `src/index.ts` registers auto-forget, lesson-decay, insight-decay, the recent-searches sweep, and consolidation, and never eviction; the only trigger is `POST /agentmemory/evict`, which no deployment calls on a schedule. On a live 201k-observation store, a dry run reported **55,716 evictions due against the 10,000-per-project default** — the cap has effectively never been enforced.

## Fix

Schedule the sweep at boot on the auto-forget pattern (`EVICTION_ENABLED` opt-out, `EVICTION_INTERVAL_MS` override, 6h default, `unref()`'d timer), with three hardenings this particular timer needs:

- **Outcome logging.** The sweep can run genuinely long — stale-session recovery fans out an LLM `mem::summarize` and `mem::graph-extract` per recovered session plus a corpus-wide consolidation pass — so the body logs completion (returned stats, elapsed ms) and failure explicitly rather than a bare `try {} catch {}` that would silently absorb a mid-sweep timeout and leave the cap unenforced with no trace.
- **Overlap guard.** If a slow sweep outlasts the interval, the next tick is dropped (with a warning) instead of overlapping: `decrementImageRef` is lock-protected per image path but not idempotent across separate passes, so two concurrent passes evicting observations that share an image could each decrement its refcount and delete an image a third observation still references.
- **Interval validation.** `parsePositiveIntervalMs` (in `config.ts`, exported so its rejection cases are unit-testable) accepts only a plain positive decimal integer within Node's 32-bit timer range. Everything it rejects would arm the timer far more often than configured: `parseInt("abc")` is `NaN` and `setInterval(fn, NaN)` fires on effectively every event-loop tick; `parseInt` truncates `"1e3"`/`"1.5"` to 1; and a value above 2147483647 overflows Node's timer delay and is coerced to **1ms** — for a data-deleting timer, each is a continuous destructive loop. The existing auto-forget and consolidation intervals get the same guard. Regression tests cover all three adversarial shapes, verified to fail against the naive `parseInt` version.
- **Visible arming.** The "eviction armed" confirmation reports through `logger` (reaching daemon logs) with `bootLog` kept for `--verbose` parity — an operator should get a log line saying a destructive 6-hourly sweep just started running.

## Tests

The scheduler lives inside `main()`, which connects to a real engine and opens listeners on import, so the registration is pinned structurally (source-regex on the actual call shape — the guarded `setInterval` triggering `mem::evict {dryRun:false}` timed by `evictionIntervalMs` and unref'd), matching the idiom `test/session-end-triggers-graph.test.ts` and `test/events-consolidation.test.ts` already use for boot-time wiring. Plus behavioral tests for the completion/failure logging shape, the overlap guard, the interval derivation, and `reportEvictionScheduled` reaching `logger.info`.

Full suite: 1720 passed / 1 skipped. `tsc --noEmit` unchanged at the 30 pre-existing errors (none in touched files).

Heads-up on interaction: first scheduled sweeps on long-lived stores may evict a large backlog (that's the point, but worth knowing before merging). `EVICTION_ENABLED=false` is the escape hatch.

## Summary by CodeRabbit

* **New Features**
  * Added an automatic eviction sweep that starts with the application.
  * Added configuration to enable or disable eviction and set its interval, defaulting to every six hours.
  * Added startup and runtime logging for scheduled, completed, and failed eviction sweeps.
  * Added safeguards to prevent overlapping eviction runs.

* **Bug Fixes**
  * Improved timer interval validation to reject invalid or unsafe values and use reliable defaults.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -200,3 +200,16 @@
 # -----------------------------------------------------------------------------
 
 # OBSIDIAN_AUTO_EXPORT=true                      # Auto-export memories to an Obsidian vault on every consolidation
+
+# -----------------------------------------------------------------------------
+# 11. Eviction sweep (boot scheduler)
+# -----------------------------------------------------------------------------
+#
+# mem::evict (stale sessions, low-importance observations, the
+# maxObservationsPerProject cap, expired/non-latest memories) was previously
+# reachable only from POST /agentmemory/evict, so the cap was never enforced
+# on a running deployment - a dry run against a live store reported 55,716
+# evictions due against the 10,000 default. Scheduled on boot like auto-forget.
+
+# EVICTION_ENABLED=true                          # Run the eviction sweep on a timer. Default on.
+# EVICTION_INTERVAL_MS=21600000                  # Milliseconds between sweeps (default 6h)
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,9 @@
 #
 # Every line is OFF by default — `agentmemory` runs out of the box with no
 # LLM key, no embedding key, and no API auth. Set keys here only when you
-# want to enable the corresponding feature.
+# want to enable the corresponding feature. The one exception is
+# `EVICTION_ENABLED`, which is on by default; set it to `false` to stop the
+# sweep.
 #
 # Run `npx -y @agentmemory/agentmemory@latest init` to copy this file into
 # place automatically. Run `npx -y @agentmemory/agentmemory@latest doctor`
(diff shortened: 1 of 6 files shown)


DanielCarmingham:
Both addressed in `ce2e393`, though one of them differently than suggested.

**`.env.example` default-state claim — fixed.** You're right, and it was a genuine trap: the overview said every line is off by default, while `EVICTION_ENABLED` reads `!== "false"` and this PR is what makes it actually run on a timer. Someone enabling the daemon could get a destructive sweep they never opted into. The exception is now stated in the overview.

**Duplicate `reportEvictionScheduled` — no duplicate exists.** I checked before changing anything: there is exactly one definition, `src/functions/evict.ts:117`, and it is exported. `src/index.ts` imports it at line 51 and calls it at line 617. `src/index.ts:117-121` is `writeWorkerPidfile`, an unrelated helper — I think the line range was carried across from `evict.ts`. So there is nothing to consolidate and no divergence risk. Happy to look again if you're seeing something different.

Also trimmed the added comments to the source-comment guideline, and dropped some over-specific numbers from a live deployment that didn't belong in upstream docs.

PR #1288, opened 2026-08-29
fix(session): gate terminal session-end write behind an explicit final flag

### LINKED ISSUE #745: `stop` hook prematurely calls `/agentmemory/session/end`, causing sessions to be marked as ended before user actually exits
## Bug Description

The `agentStop` hook (`plugin/scripts/stop.mjs` / `src/hooks/stop.ts`) sends a request to `/agentmemory/session/end` every time the agent pauses after completing a response. In **Claude Code CLI**, this causes agentmemory to mark the session as `completed` while the user is still actively in the conversation — the session should only end when the user actually exits the CLI (which is the `sessionEnd` hook's responsibility).

## Impact

- **Context loss / session fragmentation**: agentmemory believes the session is over and treats subsequent user messages as a new session.
- **Downstream pipeline breakage**: `session-end.mjs` is supposed to trigger `slot-reflect`, `graph-extract`, and consolidation **once** at true session exit. Because `stop.mjs` already ended the session, these lifecycle events either fire prematurely or against an already-completed session.
- **Log noise**: The old `triggerVoid failed` warning (pre-#731) was partly caused by `stop.mjs` repeatedly firing `session/end` and triggering `event::session::stopped` in a partially-teardown state.

## Root Cause

In `src/hooks/stop.ts`, the `agentStop` event handler performs two requests:

```typescript
fetch(`${REST_URL}/agentmemory/summarize`, ...)   // ✅ reasonable: summarize the turn
fetch(`${REST_URL}/agentmemory/session/end`, ...) // ❌ breaks Claude Code
```

The plugin's hook configuration (`plugin/hooks/hooks.copilot.json`) clearly separates the two events:

| Hook | Script | Intended semantics |
|------|--------|-------------------|
| `agentStop` | `stop.mjs` | Agent finished a response, pausing for next user input |
| `sessionEnd` | `session-end.mjs` | User exited the CLI, true end of session |

Calling `session/end` in `agentStop` conflates "agent paused" with "session ended" for Claude Code.

### Historical Context

The `/agentmemory/session/end` call was **intentionally added** in commit `355124141625ccc0d740ae08ddaaf77fe2c165ae` (#579, fixes #493) as a workaround for **Codex**, which does **not** fire a separate `SessionEnd` event. The commit message explicitly states:

> "Codex does not fire a separate SessionEnd event, so its Stop hook is the only signal we get when a Codex session terminates."
>
> "For Claude Code this is a harmless idempotent second call (session-end.mjs runs on the dedicated SessionEnd hook and sets the same endedAt + status fields). For Codex it's the only path that closes the lifecycle."

However, the assumption that this is "harmless idempotent" for Claude Code is **incorrect** in practice. Claude Code CLI fires `agentStop` after **every** assistant turn (not just at session exit), so the session gets repeatedly marked as `completed` mid-conversation.

## Reproduction Steps

1. Start Claude Code CLI with the agentmemory plugin enabled.
2. Send any message to the agent and wait for it to complete its response.
3. **Do not exit Claude Code** — just let the agent finish and return to the prompt.
4. Observe the agentmemory service logs:
   ```
   [agentmemory] warn event::session::stopped trigger failed
   ```
   (or pre-#731: `triggerVoid failed`)
5. Check the session status in agentmemory state — it will show `status: "completed"` and `endedAt` set, even though the CLI is still open.

## Expected Behavior

- **Claude Code**: `stop.mjs` (`agentStop`) should **only** call `/agentmemory/summarize` (to capture the turn's context) and **must not** call `/agentmemory/session/end`. `session-end.mjs` (`sessionEnd`) is the sole caller.
- **Codex**: Still needs a path to close the session lifecycle, since Codex lacks `SessionEnd`.

## Suggested Fix

The fix needs to **preserve Codex support** while fixing Claude Code.

### Option A: Agent-type guard (recommended)

Only call `session/end` in `stop.mjs` when running under Codex (or more generally, when no `SessionEnd` hook is available). The hook payload may contain an `entrypoint` or agent-type field that can be used to distinguish Codex from Claude Code.

```typescript
// In stop.ts — only end session for agents that lack a dedicated SessionEnd hook
const isCodex = data.entrypoint === "codex" || /* other Codex identifiers */;
if (isCodex) {
  fetch(`${REST_URL}/agentmemory/session/end`, {
    method: "POST",
    headers: authHeaders(),
    body: JSON.stringify({ sessionId }),
    signal: AbortSignal.timeout(5000),
  }).catch(() => {});
}
```

### Option B: Move Codex session-end to a Codex-specific hook/config

If the plugin system supports per-agent hook mappings, map Codex's `agentStop` to a separate script that includes `session/end`, while Claude Code continues to use the current `stop.mjs` without it.

### Option C: Remove `session/end` from `stop.mjs` and accept Codex regression

Remove the call entirely. This fixes Claude Code but regresses Codex session lifecycle (sessions will show as `active` in the viewer until Codex adds a `SessionEnd` event).

## Affected Files

- `src/hooks/stop.ts` (source)
- `plugin/scripts/stop.mjs` (build artifact, auto-generated)

## Related

- Commit `355124141625ccc0d740ae08ddaaf77fe2c165ae` — intentionally added `session/end` to `stop.mjs` as a Codex workaround (#579, fixes #493).
- #726 / #731 — the `triggerVoid` migration surfaced this because `stop.mjs` repeatedly triggered `session/end`, amplifying the `triggerVoid is not a function` noise.

DanielCarmingham:
## Problem

Claude Code fires `Stop` at the end of **every** assistant turn, not only at genuine session end, and the Stop hook posts the same `{sessionId}` payload to the same `POST /agentmemory/session/end` endpoint as the real `SessionEnd` hook. `api::session::end` writes `endedAt` + `status:"completed"` on every one of those posts, so every live session looks terminated after its first turn — the source of the phantom "abandoned session" diagnostics in #745.

Two things worth stating explicitly, because the obvious fix is wrong:

- `event::session::ended` has **no publisher anywhere in `src/`** — it's a dead subscriber, so "the terminal write lives elsewhere" is false.
- The per-turn and genuine-end payloads were byte-identical, so the server could not tell the two callers apart. Deleting the terminal write outright would mean *nothing* ever marks a session completed, which makes the "active over 24h" diagnostic worse, not better.

## Fix

Add an optional `final?: boolean` to the `session/end` request body. The `kv.update(endedAt, status:"completed")` write runs only when `body.final === true` (strict equality, so non-boolean values can't coerce into a terminal write). The `event::session::stopped` fan-out stays unconditional, so summarize / graph-extraction / consolidation keep running every turn exactly as before.

Callers updated to send `final: true` at genuine end only: the `SessionEnd` hook (`session-end.ts`, with `plugin/scripts/session-end.mjs` rebuilt via `npx tsdown` since it's a compiled artifact), the CLI, the viewer, the OpenCode capture plugin, the Pi integration, and the Hermes integration. An older plugin whose `SessionEnd` hook predates the flag simply never marks the session completed — strictly better than marking it completed every turn, and it self-heals on plugin update.

## Tests

Trigger-level tests that a `final`-less post leaves the session `active` with no `endedAt` while still firing the stop fan-out, and that `final: true` writes the terminal state; a hermes-plugin test pins `final: true` on its genuine end path.

Full suite: 1719 passed / 1 skipped. `tsc --noEmit` unchanged at the 30 pre-existing errors (none in touched files).

Closes #745.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session-ending actions now correctly mark sessions as completed.
  * Per-turn stop events no longer accidentally close active sessions.
  * Only an explicit `final: true` request completes a session.

* **Documentation**
  * Clarified when the session-ending API permanently terminates a session.

* **Tests**
  * Added coverage for final and non-final session-ending requests and lifecycle events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1618,7 +1618,7 @@ Create `~/.agentmemory/.env`:
 |--------|------|-------------|
 | `GET` | `/agentmemory/health` | Health check (always public) |
 | `POST` | `/agentmemory/session/start` | Start session + get context |
-| `POST` | `/agentmemory/session/end` | End session |
+| `POST` | `/agentmemory/session/end` | End session (pass `final: true` for genuine termination — a per-turn call without it leaves the session marked active) |
 | `POST` | `/agentmemory/observe` | Capture observation |
 | `POST` | `/agentmemory/smart-search` | Hybrid search |
 | `POST` | `/agentmemory/context` | Generate context |
--- a/integrations/hermes/__init__.py
+++ b/integrations/hermes/__init__.py
@@ -383,8 +383,14 @@ class AgentMemoryProvider(MemoryProvider):
         })
 
     def on_session_end(self, messages: list, **kwargs: Any) -> None:
+        # #745: this fires at genuine session end (not per-turn), so set
+        # final=True or the session would sit "active" forever and could
+        # trip the stale-session diagnostic - same fix already applied to
+        # every other first-party integration (src/hooks/session-end.ts,
+        # plugin/opencode/agentmemory-capture.ts, integrations/pi).
         _api(self._base, "session/end", {
             "sessionId": kwargs.get("session_id", self._session_id),
+            "final": True,
         })
 
     def on_pre_compress(self, messages: list, **kwargs: Any) -> None:
--- a/integrations/pi/index.ts
+++ b/integrations/pi/index.ts
@@ -394,8 +394,11 @@ export default function agentmemoryExtension(pi: ExtensionAPI) {
     if (event.reason !== "quit") return;
     if (!lastHealthOk || !sessionId) return;
     // session/end already fans out the summary server-side (#1203).
+    // #745: this fires only on a genuine quit (guarded above), not per-turn,
+    // so set final:true or the session would sit "active" forever and could
+    // trip the stale-session diagnostic.
     await callAgentMemory("session/end", {
-      body: { sessionId },
+      body: { sessionId, final: true },
       timeoutMs: 5_000,
     });
     void callAgentMemory("consolidate", { body: {} });
--- a/plugin/opencode/agentmemory-capture.ts
+++ b/plugin/opencode/agentmemory-capture.ts
@@ -330,7 +330,10 @@ export const AgentmemoryCapturePlugin: Plugin = async (ctx) => {
           if (DEBUG) console.error("[agentmemory] session.deleted with no session ID");
           return;
         }
-        await post("/session/end", { sessionId: sid });
+        // #745: session.deleted is a genuine one-shot session end (not a
+        // per-turn call), so set final:true or the session would sit
+        // "active" forever and could trip the stale-session diagnostic.
+        await post("/session/end", { sessionId: sid, final: true });
         post("/crystals/auto", { olderThanDays: 7 }, 30000);
         post("/consolidate-pipeline", { tier: "all", force: true }, 30000);
         if (sid === activeSessionId) activeSessionId = null;
--- a/plugin/scripts/session-end.mjs
+++ b/plugin/scripts/session-end.mjs
@@ -108,7 +108,10 @@ async function main() {
 	fetch(`${REST_URL}/agentmemory/session/end`, {
 		method: "POST",
 		headers: authHeaders(),
-		body: JSON.stringify({ sessionId }),
+		body: JSON.stringify({
+			sessionId,
+			final: true
+		}),
 		signal: AbortSignal.timeout(3e4)
 	}).catch(() => {});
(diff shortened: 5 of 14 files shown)


DanielCarmingham:
Comments trimmed in `a89cb78` — kept why the check is a strict `=== true` and why a missing flag is the safe backward-compatible path, dropped the narration.

On the Vitest mock: declining, with the same reasoning as #1285. The local `mockSdk`/`mockKV` helpers are this repository's dominant test pattern (53 files define them; 2 files mock `iii-sdk`, both via `importOriginal` to keep the real `TriggerAction`).

More to the point, the concern does not apply here: `test/session-end-final-flag.test.ts` registers `event::session::stopped` and asserts the fan-out fires, so it executes `src/triggers/api.ts:698` and the **real** `TriggerAction.Void()` on every run. All 7 cases pass. Mocking the module would replace a real call that currently works with a stub — weaker coverage, not stronger.

PR #1303, opened 2026-08-31
fix(replay): extract lessons as whole sentences, not clauses after a trigger

### LINKED ISSUE #1292: deriveCrystalAndLessons slices lessons mid-sentence: 775/981 (79%) unusable, and they cannot decay out
## Summary

`deriveCrystalAndLessons` slices lessons out of the **middle** of sentences, discarding the subject. On my store that produced **775 of 981 auto-import lessons (79%) as unusable clauses**, and several of them **invert the meaning of the source sentence**. Because lessons are surfaced by `mem::lesson-recall` and injected into agent context, these are not cosmetic — they are acted on.

Version: `@agentmemory/agentmemory@0.9.29`, engine `iii` 0.22.1, macOS 15 / node 22.

## Root cause

`dist/index.mjs:15296`:

```js
const LESSON_PATTERNS = [
  /\b(always|never|don'?t|do not|make sure|remember to|note:|caveat:|warning:)\b[^.\n]{10,200}[.!\n]/gi,
  /\b(prefer|avoid)\s[^.\n]{10,200}[.!\n]/gi,
];
```

`\b` matches the trigger token **anywhere** in a sentence, and the match *begins* at that token. Everything to its left — the subject, the actor, the condition — is thrown away, and the match runs to the next `.` or newline.

So a source sentence like

> The watchdog must never restart on `ROUTE_MISSING_404`.

is stored as the lesson

> `never restart on ROUTE_MISSING_404.`

which reads as a blanket prohibition on all restarts, including the documented recovery procedure. That specific lesson was recalled during an outage on my machine and had to be deleted by hand.

Two consequences fall straight out of the regex:

1. **100%** of my 981 auto-import lessons begin with a trigger token — that is the definition of the match, not a coincidence.
2. `\b` also fires inside identifiers, so slugs get minted as lessons. Real examples from my store:
   - `dont-repeat-yourself]]/some-page`
   - `never]]`
   - `dont-repeat-yourself|don't`

`deriveCrystalAndLessons` applies no validity filter beyond `20 <= length <= 220`, so anything the regex emits is written.

## Measurements

983 lessons in the store (981 auto-import via `source: "consolidation"`, 2 manual). Classifying the 981 by whether the trigger token was *sentence-initial* (whole sentence captured) or *mid-sentence* (subject discarded):

| bucket | count | share |
|---|---:|---:|
| mid-sentence fragment | 775 | 79.0% |
| structurally broken (unbalanced `**`/backtick, truncated at the cap, dangling connector) | 53 | 5.4% |
| usable as written | 153 | 15.6% |

Median lesson length is 55 characters. Verbatim examples of what gets stored:

```
"don't resurrect its framing."
"never in production or on every launch."
"never rounded, never paraphrased into vagueness, never"
"don't change, so NOT on the cadence."
"never pushed**); fair-use"
"never weave into the committed graph, never quote sensitive values in summaries —"
```

I reimplemented the extractor exactly (same two patterns, same `slice(0,200)` / `size < 40` / `slice(0,20)` / length filters) and ran it over transcripts that had actually been imported: **10 of 10 extracted candidates were present verbatim in the live store**, confirming the reproduction is faithful. Over 50 fresh transcripts it mints 365 candidate lessons, 241 of them (66%) mid-sentence fragments.

## The fragments are self-perpetuating — they cannot decay out, and they rank *higher*

This is the part that turns a quality problem into an operational one, and I think it is a second defect.

When `deriveCrystalAndLessons` re-derives a lesson that already exists, it takes the merge branch:

```js
const merged = {
  ...existing,
  sourceIds: mergedSources,
  tags: mergedTags,
  reinforcements: (existing.reinforcements || 0) + 1,
  updatedAt: createdAt,
  lastReinforcedAt: createdAt,      // <-- decay baseline reset
};
```

`mem::lesson-decay-sweep` computes its baseline as `lastDecayedAt || lastReinforcedAt || createdAt`. So **every re-derivation pushes the decay clock forward**. A fragment that keeps being re-extracted — and they do, because the same clause recurs across sessions — never reaches the `confidence <= 0.1` threshold that would auto-soft-delete it.

Measured on my store: **702 of the 775 fragments have `reinforcements > 0`**, one of them 225 times. 639 of those are still sitting at the initial `confidence: 0.4`, untouched by weeks of decay sweeps.

It gets worse at recall time. `mem::lesson-recall` scores as:

```js
score = l.confidence * relevance * recencyBoost
// recencyBoost = 1 / (1 + daysSince(lastReinforcedAt ?? createdAt) * 0.01)
```

`recencyBoost` reads the same field. So the *more often a fragment is re-extracted, the higher it ranks* — the junk actively outcompetes real lessons. My top-ranked fragments by reinforcement count:

```
x225  "never rounded, never paraphrased into vagueness, never"
x174  "don't link out to Spotify, and don't host music either."
x170  "never delegate critical thinking to"
x165  "never mix** (the statistical constraint, from"
```

Note also that this merge branch bumps `reinforcements` **without** calling `reinforceLesson()`, so confidence never rises — but `lastReinforcedAt` moves anyway. The record looks reinforced to the decay sweep and to the ranker while never actually gaining confidence. Whatever the intent, "this clause matched a regex again" is not evidence that a lesson is good, and it should probably not reset the decay baseline at all.

One upside worth stating, since it makes cleanup viable: the merge branch spreads `{...existing}` without clearing `deleted`, and `ensureLessonIndex` skips `l.deleted`. So a soft-deleted lesson **stays** deleted even if its source transcript is re-imported. (`mem::lesson-save` behaves differently — it checks `existing && !existing.deleted`.)

## Two more things that make this harder to live with

**1. `context` is not context.** Every lesson gets `context: firstPrompt || project` — the whole originating agent prompt, capped at 200 chars. In my store **775 lessons share just 58 distinct `context` values**, because every lesson from one session gets that session's first prompt. It cannot disambiguate a fragment; it just makes the record bigger.

**2. There is no repair path.** The lesson API is create (`mem::lesson-save`), strengthen (raises confidence only) and soft-delete. There is no update. A bad lesson cannot be re-tagged, demoted, or corrected in place — the only lever is deletion. And `mem::lesson-save` fingerprints as `lsn_…` while `deriveCrystalAndLessons` fingerprints as `lesson_…`, so re-saving identical content creates a second record rather than overwriting the first.

## Suggested fixes, cheapest first

1. **Anchor the match to a sentence start.** Split into sentences first and only accept a trigger token that is sentence-initial, or extend the match leftward to the preceding boundary so the subject survives. This alone would have removed 79% of the junk in my store.
2. **Add a validity gate before the write** — reject a candidate that starts lowercase, has unbalanced `**`/backticks/`[[ ]]`, ends on a dangling connector, or sits at the 220-char cap.
3. **Require a word boundary that is also not inside an identifier** so `dont-repeat-yourself]]` stops matching.
4. **Gate the whole derivation behind a flag.** `deriveCrystalAndLessons` is currently called only from `mem::replay::import-jsonl` and is controlled by no flag at all — notably *not* by `CONSOLIDATION_ENABLED`, which gates `mem::consolidate-pipeline` instead. Operators who want the LLM semantic layer but not the regex lessons have no way to express that today.
5. **Store the surrounding sentence** as `context` instead of the session's first prompt, so a human reviewing a lesson can see what it was cut from.

Happy to send a PR for (1)+(2) if that shape is agreeable.

ericwalisko:
Fixes #1292.

## The bug

`LESSON_PATTERNS` is applied with `pat.exec(text)` and the **match** is stored as the lesson. `\b` matches a trigger token anywhere in a sentence and the match *begins* there, so everything to its left — the subject, the actor, the condition — is discarded:

```
"The watchdog must never restart on ROUTE_MISSING_404."
  -> stored as "never restart on ROUTE_MISSING_404."
```

The stored lesson reads as a blanket prohibition rather than a rule about one component. Lessons are surfaced by `mem::lesson-recall` and injected into agent context, so these are acted on, not merely displayed — the example above was recalled during an outage and contradicted the documented recovery.

Measured on one store of 983 lessons (981 written by this path): **775 (79%) were clauses cut mid-sentence**, and 100% began with a trigger token — that is the definition of the match, not a coincidence. `\b` also fires inside identifiers, so wiki slugs and filenames containing `dont`/`never` were minted as lessons.

## The fix

Two small changes in `src/functions/replay.ts`:

1. **The trigger now only selects a sentence; the whole sentence is kept.** Text is split on sentence boundaries and a sentence containing a trigger is taken intact, so the subject survives. This is what removes the 79%.
2. **A validity gate, `isUsableLesson`,** rejects the shapes that indicate a cut: a lowercase opening, no terminal punctuation, unbalanced `**` / backtick / `[[ ]]`, a trigger that is part of an identifier, plus the existing length bounds.

Both are exported so they can be tested directly. The per-session caps (40 matches, 20 lessons) and the content-addressed `fingerprintId` are unchanged, so existing lesson ids for correctly-captured sentences stay stable.

## Tests

19 new tests in `test/replay-lesson-extraction.test.ts` covering:

- the reported failure — the subject survives, and no clause starting at the trigger is emitted
- each fragment shape observed in a real store (lowercase opening, truncated tail, unbalanced markup, identifier match)
- the sentences that were **already** being captured correctly, which must keep working
- one test that reproduces the old pattern inline, so the defect stays visible in the suite rather than only in this description

Full suite: **1730 passed, 1 skipped** — no other test changed behaviour.

## Note on scope

This does not repair lessons already in a store; there is no update path on the lesson API (create / strengthen / soft-delete only), so existing fragments can only be deleted. I raised two adjacent findings in #1292 that are **not** addressed here and may deserve their own issues:

- `api::lesson-save` hardcodes `source: "manual"` and drops `sourceIds`, though `mem::lesson-save` accepts both.
- Nothing that *uses* a lesson reinforces it — `mem::lesson-recall` never calls `reinforceLesson()` — so every lesson decays to `deleted` on a fixed schedule while a re-matched fragment has its decay baseline reset on every import. That inverts which lessons survive.

Happy to adjust the gate's strictness or split it into a separate PR if you'd prefer the sentence-anchoring change on its own.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved lesson extraction to return complete, readable sentences instead of partial fragments.
  - Preserves abbreviations and initialisms such as “U.S.”, “e.g.”, and “Dr.” when splitting sentences.
  - Filters out invalid lessons, including incomplete sentences, malformed markup, unsuitable lengths, and missing trigger words.
  - Correctly handles spacing and code identifiers when identifying lessons.

- **Tests**
  - Added coverage for sentence splitting and valid, invalid, and previously problematic lesson formats.

### DIFF
--- a/src/functions/replay.ts
+++ b/src/functions/replay.ts
@@ -63,18 +63,11 @@ function rawFromCompressed(obs: CompressedObservation): RawObservation {
   };
 }
 
-// A lesson is a whole sentence, not a clause starting at a trigger word.
-//
-// These patterns used to be applied with `pat.exec(text)` and the MATCH stored
-// as the lesson. Because `\b` matches a trigger token anywhere in a sentence
-// and the match begins there, everything to the left -- the subject, the actor,
-// the condition -- was discarded:
-//
-//   "The watchdog must never restart on ROUTE_MISSING_404."
-//     -> stored as "never restart on ROUTE_MISSING_404."
-//
-// which reads as a blanket prohibition rather than a rule about the watchdog.
-// The trigger is now only used to SELECT a sentence; the whole sentence is kept.
+// A lesson is a whole sentence, not the clause after a trigger word.
+// These patterns were previously applied with `pat.exec(text)` and the MATCH
+// stored, so "The watchdog must never restart on 404." became "never restart
+// on 404." -- a blanket prohibition instead of a rule about the watchdog. The
+// trigger now only SELECTS a sentence; the sentence is kept whole.
 const LESSON_TRIGGER =
   /\b(always|never|don'?t|do not|make sure|remember to|note:|caveat:|warning:|prefer|avoid)\b/i;
 
@@ -83,7 +76,35 @@ const LESSON_TRIGGER =
 const IDENTIFIER_TRIGGER =
   /(?:^|[\w])(?:always|never|don'?t|do not|prefer|avoid)(?=[-_|\]/])/i;
 
-const SENTENCE_SPLIT = /(?<=[.!?])\s+/;
+// Splitting naively on /(?<=[.!?])\s+/ breaks inside abbreviations, which
+// truncates a rule at exactly the point that changes its meaning:
+//   "Never ship to the U.S. without a compliance review."
+//     -> "Never ship to the U.S."   (the condition is gone)
+const ABBREVIATION =
+  /(?:^|[\s(["'])(?:e\.g|i\.e|etc|vs|cf|al|approx|incl|excl|resp|no|fig|ca|mr|mrs|ms|dr|prof|sr|jr|st)\.$/i;
+// An initialism: "U.S.", "I.B.M.", "J. R." -- a trailing single capital + dot.
+const INITIALISM = /(?:^|[\s(["'])(?:[A-Za-z]\.)+$/;
+
+/** Split text into sentences without breaking inside abbreviations. */
+export function splitSentences(text: string): string[] {
+  const parts: string[] = [];
+  const boundary = /[.!?]\s+/g;
+  let start = 0;
+  let m: RegExpExecArray | null;
+  while ((m = boundary.exec(text)) !== null) {
+    const head = text.slice(start, m.index + 1);
+    const rest = text.slice(m.index + m[0].length);
+    // Not a real boundary if this "end" is an abbreviation or an initialism,
+    // or if what follows does not begin a new sentence.
+    if (ABBREVIATION.test(head) || INITIALISM.test(head)) continue;
+    if (!/^[A-Z`"'([\d]/.test(rest)) continue;
+    parts.push(head.trim());
+    start = m.index + m[0].length;
+  }
+  const tail = text.slice(start).trim();
+  if (tail) parts.push(tail);
+  return parts.filter(Boolean);
+}
 
 export const MIN_LESSON_LENGTH = 20;
 export const MAX_LESSON_LENGTH = 220;
@@ -123,7 +144,7 @@ export function isUsableLesson(candidate: string): boolean {
  */
 export function extractLessonCandidates(text: string): string[] {
   const out: string[] = [];
-  for (const raw of text.split(SENTENCE_SPLIT)) {
+  for (const raw of splitSentences(text)) {
     const sentence = raw.replace(/\s+/g, " ").trim();
     if (!LESSON_TRIGGER.test(sentence)) continue;
     if (isUsableLesson(sentence)) out.push(sentence);
--- a/src/functions/replay.ts
+++ b/src/functions/replay.ts
@@ -63,10 +63,73 @@ function rawFromCompressed(obs: CompressedObservation): RawObservation {
   };
 }
 
-const LESSON_PATTERNS: RegExp[] = [
-  /\b(always|never|don'?t|do not|make sure|remember to|note:|caveat:|warning:)\b[^.\n]{10,200}[.!\n]/gi,
-  /\b(prefer|avoid)\s[^.\n]{10,200}[.!\n]/gi,
-];
+// A lesson is a whole sentence, not a clause starting at a trigger word.
+//
+// These patterns used to be applied with `pat.exec(text)` and the MATCH stored
+// as the lesson. Because `\b` matches a trigger token anywhere in a sentence
+// and the match begins there, everything to the left -- the subject, the actor,
+// the condition -- was discarded:
+//
+//   "The watchdog must never restart on ROUTE_MISSING_404."
+//     -> stored as "never restart on ROUTE_MISSING_404."
+//
+// which reads as a blanket prohibition rather than a rule about the watchdog.
+// The trigger is now only used to SELECT a sentence; the whole sentence is kept.
+const LESSON_TRIGGER =
+  /\b(always|never|don'?t|do not|make sure|remember to|note:|caveat:|warning:|prefer|avoid)\b/i;
+
+// A trigger that is part of an identifier is not prose -- it is a filename, a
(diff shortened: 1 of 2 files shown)


ericwalisko:
Thanks — the abbreviation one was a real catch and I've fixed it. Taking the three findings in turn:

**1. Abbreviation boundaries — fixed in 782c2c4.** This was a genuine bug and worse than it first looks, because it reintroduces the exact failure this PR exists to fix, one layer down:

```
"Never ship to the U.S. without a compliance review."
  -> "Never ship to the U.S."
```

The condition is gone and the rule inverts. It also passed the validity gate — capital opening, trigger present, terminal punctuation — so nothing downstream would have caught it.

`splitSentences()` now declines a boundary when the text before it ends in a known abbreviation (`e.g.`, `Dr.`, `etc.`) or an initialism (`U.S.`, `I.B.M.`), or when what follows doesn't begin a new sentence. Merging two sentences is the safe direction: the gate still applies and no condition is lost.

6 tests added, including the `U.S.` case. Reverting just the splitter to the naive version fails exactly those 3 abbreviation tests and no others.

**2. Rationale comments — condensed.** Trimmed from 12 lines to 5, keeping the one worked example. I'd rather not drop it entirely: the failure is subtle enough that "why isn't this just `exec`?" is a fair question for the next reader, and the repo already carries this kind of note (e.g. the `fingerprintId` comment just below it). Happy to cut it further if you'd prefer.

**3. `vi.mock("iii-sdk")` — respectfully skipping.** `replay.ts` imports `iii-sdk` as `import type { ISdk }` only, which is erased at compile time, so there's no runtime dependency to mock. The functions under test are pure and touch no KV. `test/replay-sensitive.test.ts` imports the same module and mocks nothing. Both the file and the full suite pass without it — adding the mock would be inert.

Full suite after all three: **1736 passed, 1 skipped.**

PR #1304, opened 2026-08-31
fix(state): reclaim vector generations when no vector index is loaded

inix-x:
> **Stacked on #1256.** That PR adds the generation ledger this builds on and is not merged
> yet, so this diff contains its three commits as well. The change proposed here is the
> final commit, .
> Review #1256 first; merging it will shrink this diff to that one commit.

`runSave` gates the vector save on `this.vector`, and `src/index.ts` builds a vector index only when an embedding provider exists:

```ts
const vectorIndex = embeddingProvider ? new VectorIndex() : null;
```

So a boot without a provider never calls `saveVectorIndex`. Reclaim lives inside that save, so the vector ledger is never read and its superseded generations are never collected. The BM25 save is unconditional and keeps cycling beside it, which is why the leak is one-sided.

Those shards are then unreachable. `StateKV.list` takes an exact scope rather than a prefix, so a generation that sits in no ledger entry cannot be enumerated back.

## What this looks like in a real deployment

A self-hosted store held four index generations. Two were live, two were not:

```
idx_mtgnx9tj  455 MB / 217 files  bm25     manifest refs 436   <- live
idx_mtgnxfkn    1 MB /   1 file   vectors  manifest refs   4   <- live
idx_mtb0by32   83 MB /  40 files  bm25     manifest refs   0
idx_mtfi222t   17 MB /   9 files  vectors  manifest refs   0
```

`idx_mtfi222t` is the one this PR addresses. All 9 files are vector shards, written minutes after a redeploy that came up without a usable embedding provider. The vectors ledger listed only the live generation:

```
data:manifest:gc      -> idx_mtgnx9tj, idx_mtgnxfkn
vectors:manifest:gc   -> idx_mtgnxfkn
```

The store had grown to 1,324 MB. Shards are loaded eagerly at boot, so the disk cost is also a resident cost.

## The change

Reclaim runs on its own when there is no vector index to save. It reads the vector manifest only to learn which generation is live, then hands off to the existing `reclaimGenerations`. A manifest that is missing, unreadable, or nameless reclaims nothing.

No new deletion logic. The classification of what is superseded is unchanged.

## Tests

One case: publish a vector generation, publish over it with deletes failing so the old one survives in the ledger as superseded, then save again with a null vector index.

It fails against the unmodified file. Returning early from the new path fails it again.

## What this does not do

- **It does not recover generations already stranded.** They are in no ledger and no manifest names them, so nothing in-process can find them. `idx_mtb0by32` in the store above is a separate, older case and stays where it is.
- **It does not change what counts as superseded.** Only whether the existing pass gets a chance to run.
- **It adds a manifest read to saves that skip the vector index.** One `state::get` per debounce on a path that previously did no vector work.

## Verification

- `npm test`: 1,725 passed, 0 failed, 1 skipped.
- `npx tsc --noEmit`: 30 errors, unchanged from the base. The one in this file (`:85`) is pre-existing, confirmed by stashing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved index storage cleanup so abandoned search-index generations are reclaimed after failed saves, interrupted operations, or startup recovery.
  * Prevented concurrent saves from overwriting the currently published index generation.
  * Improved resilience when cleanup records are missing, malformed, unreadable, or from a newer version.
  * Added throttled failure handling to reduce repeated persistence errors.

* **Tests**
  * Added coverage for garbage collection, rollback failures, interrupted saves, missing indexes, and coalesced concurrent saves.

### DIFF
--- a/src/state/index-persistence.ts
+++ b/src/state/index-persistence.ts
@@ -168,6 +168,18 @@ export class IndexPersistence {
       } catch (err) {
         this.logFailure("vector", err);
       }
+    } else {
+      // No embedding provider this boot, so there is no vector index to save
+      // (src/index.ts builds one only when a provider exists). The vector
+      // manifest and its ledger still sit on disk from a boot that had one,
+      // and nothing else will ever revisit them: reclaim runs inside the save
+      // this branch skips, and StateKV cannot enumerate scopes to find the
+      // shards later. Reclaim on its own so that generation is not stranded.
+      try {
+        await this.reclaimVectorGenerations();
+      } catch (err) {
+        this.logFailure("vector", err);
+      }
     }
   }
 
@@ -227,6 +239,21 @@ export class IndexPersistence {
       BM25_KEY,
       BM25_SHARD_SCOPE_PREFIX,
     );
+  }
+
+  /**
+   * Reclaim superseded vector generations without saving a vector index.
+   *
+   * Reads the manifest only to learn which generation is live; the ledger
+   * holds everything else. A manifest that is missing or nameless reclaims
+   * nothing, which is the fail-safe direction.
+   */
+  private async reclaimVectorGenerations(): Promise<void> {
+    const manifest = await this.kv
+      .get<IndexShardManifest>(KV.bm25Index, VECTOR_MANIFEST_KEY)
+      .catch(() => null);
+    if (manifest?.v !== 1 || !manifest.generation) return;
+    await this.reclaimGenerations(VECTOR_MANIFEST_KEY, manifest.generation);
   }
 
   private async saveVectorIndex(serialized: string): Promise<void> {
--- a/src/state/index-persistence.ts
+++ b/src/state/index-persistence.ts
@@ -90,15 +90,7 @@ export class IndexPersistence {
   private timer: ReturnType<typeof setTimeout> | null = null;
   private lastFailureLogAt = new Map<string, number>();
   private queue: Promise<void> = Promise.resolve();
-  /**
-   * A save that is queued but has not started running yet.
-   *
-   * It will serialise the index as it stands at the moment it runs, so it
-   * already covers every caller that arrives before then. Handing them this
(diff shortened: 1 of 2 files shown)


inix-x:
Closing unmerged. The change is redundant: `load()` already does this.

`load()` calls `loadVectorData()` unconditionally, without reading `this.vector`, and `loadShardedData` ends in `enqueue(() => this.reclaimGenerations(manifestKey, live))`. `src/index.ts:403` runs `load()` on every boot regardless of provider. So a provider-less boot already reclaims superseded vector generations.

Probed directly on the committed source, with the same setup this PR's test uses and calling only `load()`:

```
after gen_mid brownout save, gen_old shard on disk: true
after load() only (vector=null), gen_old shard on disk: false
```

The commit's own comment says "nothing else will ever revisit them". The same file contradicts it. That claim was wrong when I wrote it.

I could not build a test that isolates the change. Four attempts: asserting on a still-live generation, re-adding a shard the ledger had already dropped, a failed-delete brownout, and a boot-time manifest read failure. Each passed against the unmodified file once the real boot sequence was included.

The residual gap is real but narrow. Load-path reclaim runs once, and only when the manifest read succeeds and the shards load. A brownout wide enough to defeat it usually defeats the retry too, since both read the same manifest through the same connection. And the generation stays in the ledger either way, so the next boot collects it. That is a deferral, not a loss, which this file already treats as acceptable elsewhere.

Cost against that: two extra `state::get` per debounce for the life of every BM25-only process, on a path `flushIndexSave` awaits during deletes.

The production orphan that started this is still unexplained. 9 vector shards, 17 MB, survived at least one restart. If load-path reclaim works, a boot should have collected it. Next step is reading the live ledger to see whether it holds any entry before the live one — if it holds only the live entry, `reclaimGenerations` returns without deleting and no fix on this path can ever collect those shards.

Branch kept at `fix/reclaim-vector-gen-without-provider`.

PR #1307, opened 2026-08-31
fix(sessions): close idle sessions no client ever ended

inix-x:
## The bug

A session only reaches `completed` when a client POSTs `/agentmemory/session/end`
(`api::session::end` in `src/triggers/api.ts` is the sole writer of that status
on the live path). Several clients never send it:

- **Claude Agent SDK child sessions.** `src/hooks/stop.ts:35-39` returns when
  `isSdkChildContext(data)` is true, and that `return` sits *before* the
  `/session/end` POST at `:44`. `src/hooks/subagent-stop.ts:45` posts only to
  `/observe`. So no hook ever ends these.
- **opencode.** The vendored plugin posts `/session/end` only on
  `session.deleted` (`plugin/opencode/agentmemory-capture.ts:327`), an explicit
  delete. Normal termination fires `session.status` → idle, which posts
  `/summarize` (`:279`).

Those sessions stay `active` forever.

## What this does NOT touch

**A normal Claude Code session is never swept.** Its Stop hook posts
`/session/end` on *every turn* (`src/hooks/stop.ts:44`, and the comment at
`src/triggers/events.ts` says so explicitly), so it is `completed` from turn one
and the status guard skips it. The swept population is only: Agent SDK children,
subagents, opencode, and any `/session/start` client that never ends.

## Measured, not inferred

Production carried **733 sessions stuck `active`** against 3,509 completed.

A live census via `memory_sessions` (20-row sample, the tool returns one page):

- every `active` row lacks `endedAt` and carries a `firstPrompt` ending in
  `(@general subagent)` or `(@explore subagent)` — Agent SDK child naming,
  matching the guard above;
- the Claude Code session in the same census is `completed` with `endedAt` set.

1500 lines of production logs show **zero `mem::evict` executions**, confirming
it has never run.

## Why not `mem::evict`

`mem::evict` already has stale-session recovery, and scheduling it was the
obvious candidate. It does not work:

1. **It never yields `completed`.** `recoverStaleSession` fires
   `event::session::stopped`, which summarizes and extracts graph but never
   touches `status`. The caller then deletes the row. Net effect is a deleted
   row, not a completed session.
2. **30-day latency**, and it skips any session that already has a summary.
3. **Four unrelated deletion phases** ride along: low-importance observations,
   per-project cap, expired memories, non-latest memories, plus image refcount
   decrements.

## The change

`mem::session-sweep` ages sessions on `updatedAt`, which `observe.ts:250` stamps
on every observation. Keying on inactivity rather than on a termination event is
what makes it harness-agnostic: it does not care which signal a given client
failed to send, so it covers Claude Code, opencode, codex, and anything future.

Reuse over new code:

- Terminal-state writing is delegated to `event::session::ended`, which already
  existed and set `endedAt` + `status` but **had no publisher anywhere in the
  repo** — its subscriber registration was the only reference to its topic.
- The timer follows the existing `AUTO_FORGET_ENABLED` pattern (env-gated
  `setInterval` + `.unref()`), and interval parsing follows
  `getConsolidationCooldownMs` in `config.ts`.
- `api::session-sweep` mirrors `api::evict`.
- The summary fan-out matches `/session/end` except for `skipConsolidation`,
  which that path does not pass because it ends one session at a time.

Safety properties, each added because a review round asked for it:

- **Inputs are validated at both boundaries.** `idleMinutes` reuses api.ts's
  existing `parseOptionalPositiveInt`, which returns 400 on a boolean, array, or
  object rather than reaching `Number()` (`Number(true)` is `1`, which would
  have meant a one-minute threshold and swept nearly every active session).
  `event::session::ended` validates `endedAt`, since it is also a durable
  subscriber and the viewer derives duration from that field.
- **The interval is clamped at both ends.** `setInterval` collapses NaN, 0, and
  anything above 2^31-1 to ~1ms alike, so a garbage *or* oversized env var would
  have run the sweep in a hot loop.
- **Overlapping runs are prevented** by an in-flight guard, since `setInterval`
  does not await its callback and two runs would each pay for a summarize.
- **Each session is re-read immediately before writing**, narrowing the
  stale-snapshot window from the whole list scan to a single `get`.
- **The state change is audited** under a new `session_sweep` operation, using
  `safeAudit` because production logs show audit writes timing out under load.

`endedAt` records the session's last activity, not the moment the sweep noticed
it.

Defaults: **one day** idle, swept every 15 minutes, capped at 25 per run,
disable with `SESSION_SWEEP_ENABLED=false`. The threshold is a day rather than an
hour because the target sessions run and finish, so a tight window buys nothing
there while a client that merely idles overnight would be ended mid-use.

## Limitations, stated up front

- **The re-read narrows the duplicate-work window, it does not close it.**
  `StateKV` has no compare-and-set, so an atomic active-to-completed claim is not
  available. The in-flight guard is per-process, so a multi-instance deployment
  would need a real lease.
- **The existing backlog drains at 25 per run.** At four runs an hour that is
  roughly 7.5 hours for 733. `POST /agentmemory/session-sweep` runs it on demand.
- **It does not fix the two client bugs**, it compensates for them. The
  `stop.ts` SDK-child guard and the opencode idle branch are both still wrong and
  want separate PRs. The guard exists to prevent agent-sdk provider re-entrancy,
  so narrowing it needs care, which is why this PR does not touch it.
- **The census is a 20-row sample, not all 733.** The population count is a prior
  measurement, not re-measured here.
- The harness attribution is inferred from prompt naming plus the source guard;
  the exact client that POSTs `/session/start` with those `ses_` ids was not
  traced.
- A session whose timestamps are both unparseable is skipped rather than swept,
  deliberately — treating an unreadable stamp as infinitely idle would end live
  sessions.

## Tests

**39 tests** across `test/session-sweep.test.ts`,
`test/session-sweep-interval.test.ts`, and `test/events-session-ended.test.ts`.

**Mutation-checked, and mutation testing found what review did not.** All **17
mutations** across the three source files kill at least one test. Three gaps it
caught that no reviewer did:

- Removing the in-flight guard's release killed nothing, which would have turned
  the sweep into a one-shot that runs at boot and never again.
- Flipping the idle comparison from `<=` to `<` killed nothing, because no
  fixture sat on the boundary.
- Dropping the re-read guard killed nothing, because the KV mock's `list` handed
  back live object references while the real `StateKV.list` is an SDK round-trip
  returning fresh objects. The test was passing for the wrong reason. The mock
  now clones.

Two tests were removed as unsound rather than repaired: one asserted a dry run
proceeds alongside a real sweep but registered a fresh function first, so nothing
was in flight; the other claimed to cover a throwing run, but `sweep()` has no
reachable throw.

## Gates

- `npm test`: 162 files passed, 1 skipped; **1750 passed**, 1 skipped.
- `npm run build`: succeeds.
- `npx tsc --noEmit`: **30 → 29 errors, zero new**. Baseline verified by
  typechecking a throwaway worktree at the base commit rather than assuming it,
  and compared line-number-insensitively so shifts from added code do not read as
  regressions. The error that disappears is
  `src/triggers/events.ts: Property 'updatedAt' does not exist on type 'Session'`
  — `observe.ts` has always written that field and the type never declared it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -117,7 +117,7 @@ Hook scripts in `src/hooks/` are standalone Node.js scripts (no iii-sdk import).
 ## Current Stats (v0.9.29)
 
 - 54 MCP tools (8 visible by default, `AGENTMEMORY_TOOLS=all` for all)
-- 130 REST endpoints
+- 131 REST endpoints
 - 6 MCP resources, 3 MCP prompts
 - 12 hooks, 17 skills
 - 260+ iii functions
--- a/README.md
+++ b/README.md
@@ -1609,7 +1609,7 @@ Create `~/.agentmemory/.env`:
 
 <h2 id="api"><picture><source media="(prefers-color-scheme: dark)" srcset="assets/tags/light/section-api.svg"><img src="assets/tags/section-api.svg" alt="API" height="32" /></picture></h2>
 
-130 endpoints on port `3111`. The REST API binds to `127.0.0.1` by default. Protected endpoints require `Authorization: Bearer <secret>` when `AGENTMEMORY_SECRET` is set, and mesh sync endpoints require `AGENTMEMORY_SECRET` on both peers.
+131 endpoints on port `3111`. The REST API binds to `127.0.0.1` by default. Protected endpoints require `Authorization: Bearer <secret>` when `AGENTMEMORY_SECRET` is set, and mesh sync endpoints require `AGENTMEMORY_SECRET` on both peers.
 
 <details>
 <summary>Key endpoints</summary>
--- a/src/config.ts
+++ b/src/config.ts
@@ -467,6 +467,22 @@ export function getConsolidationCooldownMs(): number {
   return raw >= 0 ? raw : CONSOLIDATION_COOLDOWN_DEFAULT_MS;
 }
 
+const SESSION_SWEEP_INTERVAL_DEFAULT_MS = 900_000;
+const SESSION_SWEEP_INTERVAL_MIN_MS = 60_000;
+// setInterval coerces a delay above 2^31-1 to 1ms, exactly as it does NaN or 0,
+// so an interval that is merely too large fails the same way garbage does.
+const TIMER_MAX_MS = 2_147_483_647;
+
+export function getSessionSweepIntervalMs(): number {
+  const raw = safeParseInt(
+    getMergedEnv()["SESSION_SWEEP_INTERVAL_MS"],
+    SESSION_SWEEP_INTERVAL_DEFAULT_MS,
+  );
+  // safeParseInt already falls back on NaN, and parseInt never yields Infinity,
+  // so raw is finite here and only needs bounding.
+  return Math.min(Math.max(raw, SESSION_SWEEP_INTERVAL_MIN_MS), TIMER_MAX_MS);
+}
+
 export function isStandaloneMcp(): boolean {
   return getMergedEnv()["STANDALONE_MCP"] === "true";
 }
--- a/src/config.ts
+++ b/src/config.ts
@@ -478,8 +478,6 @@ export function getSessionSweepIntervalMs(): number {
     getMergedEnv()["SESSION_SWEEP_INTERVAL_MS"],
     SESSION_SWEEP_INTERVAL_DEFAULT_MS,
   );
-  // safeParseInt already falls back on NaN, and parseInt never yields Infinity,
-  // so raw is finite here and only needs bounding.
   return Math.min(Math.max(raw, SESSION_SWEEP_INTERVAL_MIN_MS), TIMER_MAX_MS);
 }
 
--- a/src/functions/session-sweep.ts
+++ b/src/functions/session-sweep.ts
@@ -11,7 +11,11 @@ import { logger } from "../logger.js";
 // dropped it. Separate from mem::evict on purpose: that pass deletes rows, which
 // is a different decision from marking a session finished.
 const MS_PER_MINUTE = 60 * 1000;
-const DEFAULT_IDLE_MINUTES = 60;
+// A day, not an hour. The sessions this exists for (SDK children, subagents)
+// run and finish, so a tight threshold buys nothing there, while a client that
+// merely idles overnight would be ended mid-use. mem::evict's comparable
+// "this is dead" call is 30 days.
+const DEFAULT_IDLE_MINUTES = 1440;
 // ponytail: fixed cap per run, not a work queue. Each swept session fans out a
 // summarize, so an unbounded first pass over a backlog is an LLM storm.
 const MAX_PER_RUN = 25;
@@ -45,6 +49,7 @@ export function registerSessionSweepFunction(sdk: ISdk, kv: StateKV): void {
       const now = Date.now();
       const sessions = await kv.list<Session>(KV.sessions).catch(() => []);
       let candidates = 0;
+      let attempted = 0;
       let swept = 0;
 
       for (const session of sessions) {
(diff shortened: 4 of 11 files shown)


inix-x:
Thanks. Two of three are fixed in 34e6617, one I'm pushing back on.

**Validate sweep settings (Major) — fixed.** Both reachable paths were real. A negative or zero `idleMinutes` made every active session a candidate, and `setInterval` treats a NaN `SESSION_SWEEP_INTERVAL_MS` as roughly 1ms, so a bad env var ran the sweep in a hot loop. `idleMinutes` now falls back to the default unless it is a finite positive number, and the interval is clamped to a 60s floor.

The third path you grouped in, the `KV.config "sessionSweep"` override, is deleted rather than validated. Nothing in the repo writes that key, or the `"eviction"` key it was copied from at `evict.ts:115`. The only `KV.config` writer is the consolidation marker at `events.ts:26`. Removing the dead path removes the thing needing validation.

**Audit coverage (Minor) — fixed.** Added a `session_sweep` operation to the `AuditEntry` union following the checklist at AGENTS.md:42-43, plus an audit write per swept session. It uses `safeAudit` rather than `recordAudit` because production logs on this deployment show audit writes timing out under load, and one failed audit write must not abort the sweep for every session queued behind it. Agreed that the client endpoint at `api.ts:668-671` is also unaudited, but that is pre-existing and outside this PR.

**`vi.mock("iii-sdk")` (Major) — not taking this one.** AGENTS.md:113 does say that. But AGENTS.md:115, two lines later, names `test/crystallize.test.ts` as the pattern to follow for function tests, and that file uses hand-written `mockKV()` and `mockSdk()` helpers at lines 10 and 31. So does the nearest sibling, `test/evict.test.ts`. No test file in the repo currently uses `vi.mock("iii-sdk")`.

The two guideline lines conflict, so I followed the one backed by a named working example. Happy to switch if the maintainer considers :113 canonical, but this would be the first test in the repo to do it, and matching the sibling test seemed more useful than matching the doc line.

inix-x:
Round 3 findings, all addressed in 190be96. Two taken as written, one declined with a substitute.

**`Number(true)` on idleMinutes (Minor) — fixed.** Correct and the worst of the three in practice: `{"idleMinutes": true}` would have requested a one-minute threshold and completed nearly every active session. `Number(["5"])` is `5`, so arrays coerced too. The endpoint now accepts a number from the body or a numeric string from the query, and rejects booleans, arrays, objects, and blank strings, falling back to the sweep's own default.

**Validate `endedAt` before persistence (Minor) — fixed.** Agreed on the reasoning, and the durable-subscriber point is the part that matters: the payload is untrusted, and the viewer derives session duration from that field. It now requires a string that parses to a finite timestamp and falls back to now otherwise. Covered by a test over garbage, empty string, number, null, and object.

**Distributed lease for an atomic active-to-completed transition (Major) — declined, with a narrower fix.** `StateKV` exposes `get`, `set`, `update`, `delete`, and `list` (`src/state/kv.ts`). There is no compare-and-set, so a lease built on it could not actually make the transition atomic. It would add coordination machinery without delivering the guarantee, and the failure mode would be worse than the one it replaces, because it would read as atomic without being so.

What is in instead:

1. An in-flight guard on the registered function, which covers both entry points (the timer and the endpoint) in the process. This is the realistic overlap source, since `setInterval` does not await its callback.
2. A re-read of each session immediately before writing, which narrows the window from the whole list scan to a single `get`.

The code comment states plainly that this narrows the window rather than closing it, so the residual is not hidden. Genuinely closing it needs a CAS primitive on `StateKV`, which is a larger change than this PR and is worth its own discussion if the maintainer wants it.

Worth noting that verifying the re-read was harder than writing it. The test I first added passed for the wrong reason, because the KV mock's `list` handed back live object references while the real `StateKV.list` is an SDK round-trip returning fresh objects. Mutation testing caught it. The mock now clones, and all 17 mutations across the three files kill at least one test.

Issue #158, opened 2026-04-17
Health may report critical from heap ratio alone when RSS is still low


mklusz:
The current health logic is too aggressive on memory severity.

`agentmemory` can report `degraded` / `critical` from `heapUsed / heapTotal` alone, even when there is no meaningful memory pressure.

Example from a live deployment:

- `heapUsed`: ~45 MB
- `heapTotal`: ~46 MB
- `rss`: ~120 MB

This produced `memory_critical_97%`, while the service remained responsive and the host had plenty of available RAM.
For a small steady-state Node process, high heap occupancy by itself is not a reliable signal of operational unhealthiness.
Overall health should only degrade when heap pressure is accompanied by a stronger signal, such as high RSS or another explicit memory-pressure heuristic.

If useful, I can open a PR.

PR #1317, opened 2026-09-01
fix(health): measure heap saturation against the heap ceiling, not heapTotal

kpadilha:
### 🚀 Summary

- **Fix:** `evaluateHealth` measured memory saturation as `heapUsed / heapTotal`. `heapTotal` is the heap V8 has committed so far, not a ceiling, and V8 keeps it close to the live set — so the ratio reads 80-95% in any healthy process under load. Now measured against `v8.getHeapStatistics().heap_size_limit`.
- **Impact:** On a `--max-old-space-size=8192` server during bulk import: **92% reported** (`memory_critical`) vs **3.5% actual**, with RSS 690 MB on a 125 GB host, event-loop lag 0.05 ms, CPU 0.6%. The `memoryRssFloorBytes` gate from #158 masks this only while RSS stays under the floor; the denominator was still wrong.
- **Compat:** `HealthSnapshot.memory.heapSizeLimit` is optional and falls back to `heapTotal`, so snapshots persisted by older versions still evaluate and the 5 existing threshold tests pass unchanged.
- **Test:** 3 cases added — distant ceiling stays healthy, exhausted ceiling still goes critical, absent field falls back. Suite: 1714 passed, 1 skipped. `tsc --noEmit`: same 30 pre-existing errors before and after, none in the touched files.

### DIFF
--- a/src/health/monitor.ts
+++ b/src/health/monitor.ts
@@ -1,3 +1,4 @@
+import { getHeapStatistics } from "node:v8";
 import type { ISdk } from "iii-sdk";
 import type { HealthSnapshot } from "../types.js";
 import type { StateKV } from "../state/kv.js";
@@ -71,6 +72,7 @@ export function registerHealthMonitor(
         heapTotal: mem.heapTotal,
         rss: mem.rss,
         external: mem.external,
+        heapSizeLimit: getHeapStatistics().heap_size_limit,
       },
       cpu: {
         userMicros: currentCpu.user,
--- a/src/health/thresholds.ts
+++ b/src/health/thresholds.ts
@@ -59,10 +59,17 @@ export function evaluateHealth(
     degraded = true;
   }
 
+  // Saturation must be measured against the heap ceiling, not against
+  // heapTotal. heapTotal is only the heap V8 has committed so far, and V8
+  // deliberately keeps it close to the live set, so heapUsed/heapTotal sits at
+  // 80-95% in any healthy process under load. Fall back to heapTotal for
+  // snapshots written before heapSizeLimit existed.
+  const memCeiling =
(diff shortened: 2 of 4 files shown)


kpadilha:
Thanks — addressed in c913dfb, though partially rather than by removal, and I want to be explicit about why so you can overrule me.

The repo carries two versions of this rule:

- `AGENTS.md:102` — "No code comments explaining WHAT — use clear naming instead"
- `CONTRIBUTING.md:59` — "No code comments that restate what the code does. Only write a comment when the *why* is non-obvious — a hidden constraint, an invariant, a workaround for a specific bug."

The review quoted the first. Measuring both comments against the second:

**`src/types.ts`** — you were right. The first sentence restated what `heapSizeLimit` already says in its name. Trimmed to the part that is a constraint:

```ts
/** Optional: snapshots persisted before this field existed must still deserialise. */
```

**`src/health/thresholds.ts`** — I kept a shortened version. The V8 behaviour it records is exactly the invariant `CONTRIBUTING.md` describes: `heapTotal` is the committed heap, not a ceiling, and V8 keeps it close to the live set by design. That is the entire reason this bug existed, it is not visible from the surrounding code, and the naming cannot carry it. Dropped the sentence that narrated the fallback, since that one did restate the code:

```ts
// heapTotal is the heap V8 has committed so far, not a ceiling: V8 keeps it
// close to the live set, so heapUsed/heapTotal reads 80-95% in any healthy
// process under load. Measuring saturation against it is what produced #158.
```

Five lines to three, two to one. Happy to remove the remaining block entirely if you read the guideline more strictly than I have — it is your codebase and your call. Suite still green: 1714 passed, 1 skipped.

PR #1350, opened 2026-09-07
fix(providers): clear the circuit breaker failure count on success

redeye1011:
## Problem

`recordSuccess()` only resets the failure counter when leaving `half-open`. In the `closed` state a success leaves it untouched, and `recordFailure()` clears it only when two failures are more than `failureWindowMs` apart:

```ts
recordSuccess(): void {
  if (this.state === "half-open") {
    this.state = "closed";
    this.failures = 0;
    ...
```

Under steady load that means the counter is not "failures in the last minute", it is "failures since the last minute-long gap in failures". A provider succeeding ~95% of the time still trips, because each scattered failure lands inside the window opened by the previous one, and the successes in between never clear it.

Observed against a hosted proxy that returns the occasional 502: a handful of real upstream errors produced a stretch of **1,119 consecutive `circuit_breaker_open` fast-fails**, and `mem::compress` reached 22,400 calls against 15,086 failures. Almost none of those failures were the provider being unavailable — the breaker was the thing failing them.

The blast radius is wide because `ResilientProvider` wraps compress and summarize for every caller, so one flaky upstream silently disables compression, summarization and graph extraction together.

## What this changes

- `recordSuccess()` clears `failures` / `lastFailureAt` in the closed state too, so the window means "since the last success".
- Thresholds move into config (`AGENTMEMORY_CIRCUIT_FAILURE_THRESHOLD`, `_FAILURE_WINDOW_MS`, `_RECOVERY_TIMEOUT_MS`) so a deployment behind a flaky upstream can widen them without a rebuild.
- Defaults go from 3 failures / 30s recovery to 10 / 15s. Three scattered failures is not evidence that a hosted LLM proxy is down, and a shorter recovery probe gets the provider back sooner when it was never down.

## Relationship to the PRs already open

Both #1259 and #1277 change **which errors count** as failures, in `src/providers/resilient.ts` — 429s and content-filter rejections respectively. This changes **the counter never being reset by success**, in `src/providers/circuit-breaker.ts`. Different files, different mechanism; all three compose, and none of them alone fixes the others.

## Result

Same install, after the change: 8 genuine upstream timeouts, **0 fast-fails**. Breaker state stayed `closed` with `failures: 0` throughout.

## Verification

```
npm run build     # clean
npm test          # 1,715 passed, 1 skipped
```

Added to `test/circuit-breaker.test.ts`:

- 100 calls at a 5% failure rate keep the breaker closed (this is the regression — it fails on `main`)
- a success in the closed state clears an accumulated count
- a genuinely failing provider still opens the breaker
- configured thresholds and recovery timeout are honoured

The existing cases are unchanged and still pass, including the default-threshold ones.

## Summary by CodeRabbit

- **New Features**
  - Added configurable provider circuit-breaker settings for failure thresholds, monitoring windows, and recovery timeouts.
  - Settings can be customized through environment variables, with safe defaults used when values are missing or invalid.

- **Bug Fixes**
  - Successful provider requests now consistently reset circuit-breaker failure tracking.
  - Improved circuit-breaker behavior prevents intermittent failures from unnecessarily opening the circuit and preserves configured recovery behavior.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -395,6 +395,14 @@ const CIRCUIT_BREAKER_FAILURE_THRESHOLD = 10;
 const CIRCUIT_BREAKER_FAILURE_WINDOW_MS = 60_000;
 const CIRCUIT_BREAKER_RECOVERY_TIMEOUT_MS = 15_000;
 
+function positiveEnvInt(value: string | undefined, fallback: number): number {
+  const parsed = safeParseInt(value, fallback);
+  // safeParseInt returns 0 or a negative number verbatim; CircuitBreaker
+  // rejects those and substitutes its own legacy fallbacks, so an override of
+  // "0" would silently select 3 / 60s / 30s rather than the values below.
+  return parsed > 0 ? parsed : fallback;
+}
+
 export function getCircuitBreakerOptions(): {
   failureThreshold: number;
   failureWindowMs: number;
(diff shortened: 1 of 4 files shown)


redeye1011:
Thanks — pushed fixes for both correctness findings.

**Fixed**

- **`openedAt` cleared while open.** This was a real bug I introduced, and the sharpest one here: a request already in flight when another request opened the breaker still reaches `recordSuccess`, and clearing `openedAt` there stripped the recovery deadline, so `isAllowed` could never promote to half-open and the provider stayed down permanently. Successes in the `open` state are now ignored. Regression test added for the interleaving.
- **Non-positive env overrides.** `safeParseInt` returns `0` and negatives verbatim, which `CircuitBreaker` then rejects in favour of its own legacy `3` / `60s` / `30s` fallbacks rather than the defaults configured here. Now clamped so an invalid override keeps the configured value.

**Not changed**

- **Removing the explanatory comments.** `CONTRIBUTING.md` says: *"No code comments that restate what the code does. Only write a comment when the *why* is non-obvious — a hidden constraint, an invariant, a workaround for a specific bug."* These are why-comments of exactly that kind — the reason a success clears the counter is a specific production failure mode that the code alone doesn't convey, and it is what would stop someone re-introducing the bug. Happy to trim them if a maintainer reads the guideline more strictly than I do.

Issue #1300, opened 2026-08-31
[search] hybrid graph stream enumerates the whole graph per query — O(corpus) latency, event-loop stalls, no disable short-circuit


vanja-emichi:
## Summary

The hybrid search graph stream (`GraphRetrieval.searchByEntities` / `expandFromChunks` inside `HybridSearch.tripleStreamSearch`) enumerates the ENTIRE graph — `kv.list(KV.graphNodes)` + `kv.list(KV.graphEdges)` — on every query. With entity extraction making the stream effectively always-on, every `/search` and `/smart-search` pays O(corpus) state transport plus synchronous JSON.parse of both scopes, stalling the event loop and dominating search latency as the graph grows.

## Environment

- agentmemory v0.9.29, `GRAPH_EXTRACTION_ENABLED=true`, graph: 12,310 nodes (97 MB nodes scope, 68 MB edges scope)
- Same box also serves observe/compress traffic

## Evidence

Measured on an otherwise idle server (post-restart, health p50 ~0.01s):

| path | with graph stream | with stream gated off |
|---|---|---|
| `/agentmemory/smart-search` | 12.0–19.9s (and 30–57s under concurrent load) | 0.18–0.63s |
| `/agentmemory/search` | ~16s | 0.19s |
| `/agentmemory/livez` | episodic 11.7s stalls | clean |

Result-quality contribution measured on 3 queries (top-5, stream on vs off): 2 of 3 queries returned an IDENTICAL top-5; the third differed in 2 of 5 positions. The graph stream cost ~12–20s per query to (mostly) re-rank what BM25+vector already returned.

## Mechanism (verbatim from v0.9.29)

```ts
async searchByEntities(entityNames, maxDepth = 2, maxResults = 20) {
  const allNodes = (await this.kv.list<GraphNode>(KV.graphNodes)).filter((n) => !n.stale);
  const allEdges = (await this.kv.list<GraphEdge>(KV.graphEdges)).filter((e) => !e.stale);
  ...
```

`expandFromChunks` does the same double `kv.list`. Both run whenever `extractEntitiesFromQuery` yields ≥1 entity, which for natural-language queries is essentially always.

Additional findings:

[...]

PR #1351, opened 2026-09-07
perf(search): reuse one graph view per store and bound traversal

redeye1011:
Refs #1300
Refs #309

## Why this is open again

#1300 is closed as completed, but `main` still carries the reported code. On `e04ba88`:

```
src/functions/graph-retrieval.ts:49   kv.list(KV.graphNodes)   # searchByEntities
src/functions/graph-retrieval.ts:50   kv.list(KV.graphEdges)
src/functions/graph-retrieval.ts:122  kv.list(KV.graphNodes)   # expandFromChunks
src/functions/graph-retrieval.ts:123  kv.list(KV.graphEdges)
src/functions/graph-retrieval.ts:166  kv.list(KV.graphNodes)   # temporalQuery
src/functions/graph-retrieval.ts:167  kv.list(KV.graphEdges)
```

and `hybrid-search.ts` still uses `graphWeight` only in the RRF fusion, with no short-circuit. So the behaviour described there is still present.

## What the measurement adds

I hit this on a 71.5K-node / 191K-edge corpus, where search took 48.2s and rose to 63.7s over an hour as the graph grew — past the 60s MCP tool timeout, so it presented as a broken client rather than a slow one.

The enumeration is not the dominant cost, which is worth correcting because it changes the fix:

| query | graph streams run | time |
|---|---|---|
| no entity match (expand only) | 1 | **24.5s** |
| ordinary query | 2 | **48.6s** |
| both full `kv.list` calls, measured alone | — | **3.0s** |

An exact 2:1 on stream count, with enumeration accounting for only 3s of it. The rest is `dijkstraTraversal` rebuilding `nodeIndex` and `adjacency` from the full arrays **on every call** — that is once per start node, not once per query — and `searchByEntities` accepting every bidirectional substring match as a start node with no cap. #1300's second "additional finding" names the hub problem; this is the same thing dominating the profile.

## What this changes

- **A per-`StateKV` graph view** holding nodes, edges, adjacency and an observation index. Built once and patched write-through from `StateKV.set/update/delete`, so every writer — graph-extract, mesh sync, import, cascade, snapshot restore — keeps it warm without having to know it exists. Keyed per store rather than globally so a process holding more than one store cannot cross-answer. Released after an idle period so a search-quiet daemon does not hold the graph resident.
- **Traversal reads that view** instead of rebuilding per start node.
- **Bounded expansion**: start nodes capped and ranked exact-match first then most specific (`AGENTMEMORY_GRAPH_MAX_START_NODES`, default 25), and visited nodes capped per traversal (`AGENTMEMORY_GRAPH_MAX_VISITED`, default 2000), so a hub node cannot walk the corpus.
- **`AGENTMEMORY_GRAPH_WEIGHT=0` now skips the stream** instead of only zeroing its score contribution. That is the sanctioned latency escape hatch #1300 asked for; today setting it to 0 leaves both traversals running at full cost, which means an operator who set it believes the subsystem is off while it is still on the hot path.

## Result

| | before | after |
|---|---|---|
| `/agentmemory/search`, warm | 48.2s → 63.7s and climbing | **0.89s** |
| graph view build | per start node, per query | **2.4s once**, then reused |
| daemon RSS | 1636 MB | **922 MB** |

The RSS drop is the same provenance the traversal used to re-materialise on every call no longer being re-read.

## Verification

```
npm run build     # clean
npm test          # 1,716 passed, 1 skipped
```

New `test/graph-view-reuse.test.ts` asserts the enumeration count directly rather than timing anything: four retrieval calls produce exactly one nodes list and one edges list. It also covers a node written after the view was built being visible without re-enumerating, a node marked stale dropping out, and explicit invalidation forcing a rebuild.

The existing 14 `graph-retrieval` cases pass unchanged, which is the useful signal that the traversal refactor preserved behaviour.

## Summary by CodeRabbit

* **Performance**
  * Improved graph-powered search responsiveness by reusing indexed graph data across related searches.
  * Reduced repeated graph loading during entity, chunk, and time-based queries.

* **Search Improvements**
  * Entity searches now normalize queries, prioritize exact matches, and apply configurable traversal limits.
  * Graph retrieval can now be fully disabled when its weighting is set to zero.

* **Reliability**
  * Search results stay current as graph data is added, updated, deleted, or marked stale.

### DIFF
--- a/src/functions/graph-retrieval.ts
+++ b/src/functions/graph-retrieval.ts
@@ -2,8 +2,23 @@ import type {
   GraphNode,
   GraphEdge,
 } from "../types.js";
-import { KV } from "../state/schema.js";
 import type { StateKV } from "../state/kv.js";
+import { getGraphView, type GraphView } from "../state/graph-cache.js";
+import { getEnvVar } from "../config.js";
+
+// Traversal bounds. Both are here because this corpus grew to 69K
+// nodes / 185K edges: an unbounded start-node set means one Dijkstra
+// per matching node, and an unbounded expansion lets a single hub node
+// walk most of the graph on maxDepth=2.
+const DEFAULT_MAX_START_NODES = 25;
+const DEFAULT_MAX_VISITED = 2000;
+
+function envInt(key: string, fallback: number): number {
+  const raw = getEnvVar(key);
+  if (!raw) return fallback;
+  const parsed = parseInt(raw, 10);
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;
+}
 
 export interface GraphRetrievalResult {
   obsId: string;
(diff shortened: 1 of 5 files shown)


redeye1011:
Thanks — pushed fixes for both correctness findings.

**Fixed**

- **Discovery vs expansion bound.** Correct, and the cap was weaker than the PR claimed: `maxVisited` only bounded nodes popped off the heap, so one hub node added every neighbour to `pathTo` in a single iteration and all of them were returned and scored. The limit now applies when a node is first discovered. Added a star-graph test that asserts the bound on returned nodes.
- **Stale publish from an in-flight build.** Also correct. A write landing while `buildView` awaited its two `kv.list` calls had no view to patch and was silently dropped, and the finished build then published a snapshot predating it for the full TTL. Mutations arriving mid-build are queued and replayed onto the finished view; an invalidation arriving mid-build discards the result rather than caching it. Added a gated-list test for the race.

**Not changed**

- **"Route graph retrieval through iii-engine primitives."** I read that guideline as prohibiting alternative *storage backends* — the sentence it comes from names standalone SQLite — rather than in-process indexes. This module holds no data of its own: KV remains the only source of truth, every write still goes through `sdk.trigger`, and the view is a derived cache rebuilt from KV. The codebase already keeps exactly this kind of in-process derived structure in `SearchIndex` and `VectorIndex`, which are likewise built from KV and consulted on the hot path.

  Removing it would mean re-reading 71K nodes and 191K edges per query, which is the bug. Happy to move it behind a registered function if a maintainer wants it structured that way, but I'd rather confirm the intent before restructuring.

PR #1149, opened 2026-08-03
Fix slot MCP errors and Windows test portability

arkhipowiktor-dev:
## Summary

- return a structured MCP response when memory-slot tools are disabled instead of triggering an unregistered function and producing HTTP 500
- treat sessionless memories as intentionally global while retaining migration warnings for session-linked unscoped memories
- make Windows tests use platform-aware executable names and paths
- synchronize the documented REST endpoint count with the current 130 registered paths

## Root cause

The MCP tool surface was visible independently of the `AGENTMEMORY_SLOTS` registration gate. Diagnostics also assumed every unscoped memory could be backfilled from sessions, although sessionless memories have no project evidence. Several tests encoded Unix-only path and binary-name assumptions. After rebasing, the current upstream endpoint counter was also one behind the registered API paths.

## Changed files

- runtime: `src/mcp/server.ts`, `src/functions/diagnostics.ts`
- regression coverage: `test/mcp-slots-disabled.test.ts`, `test/diagnostics.test.ts`
- Windows portability: `test/cli-remove.test.ts`, `test/compress-file.test.ts`, `test/copilot-plugin.test.ts`, `test/obsidian-export.test.ts`, `test/slots-flag-gate.test.ts`
- synchronized counts: `README.md`, `AGENTS.md`, `src/index.ts`

## Validation

- `npm test` — passed on the rebased head
- `npx tsdown` — passed on the rebased head
- `git diff --check` — passed
- live `/agentmemory/health` — healthy
- live diagnostics — 3 pass, 0 warn, 0 fail
- live `memory_slot_list` — actionable disabled-feature response, no HTTP 500

## Risks and limitations

- Memory slots remain opt-in; this change does not enable them automatically.
- Sessionless memories remain global and no stored memory data is rewritten.
- Vercel deployment authorization is controlled by the upstream owner and is not exercised by this runtime/test-only change.

## Scope confirmation

- No environment files, credentials, stored memory data, deployment configuration, or unrelated product areas were changed.
- Generated hook bundles were excluded from the commit.

## Exact head

- Commit: 31776a78f871fe4215fee7648c7dd535ccd4122c
- CI: pending after rebased force-with-lease update

Closes #1148

Issue #1148, opened 2026-08-03
Fix disabled slot MCP errors and Windows test portability


arkhipowiktor-dev:
## Problem

AgentMemory exposed MCP slot tools even when `AGENTMEMORY_SLOTS` was disabled. Calling `memory_slot_list` then triggered an unregistered function and returned HTTP 500.

Diagnostics also classified sessionless global memories as missing project scope and suggested a migration that could not infer a project (`ambiguous` result).

The full unit test suite additionally had Windows-only failures caused by Unix path literals, `HOME`-only test sandboxes, and the Windows `iii.exe` suffix.

## Expected behavior

- Disabled slot tools return an actionable MCP response instead of HTTP 500.
- Sessionless memories are treated as intentionally global.
- Session-linked unscoped memories still produce the migration warning.
- The unit test suite passes on Windows and Unix-compatible path behavior remains covered.

## Validation

- `npm test`
- `npx tsdown`
- `git diff --check`
- Live `/agentmemory/health`, diagnostics, and `memory_slot_list`

PR #1352, opened 2026-09-07
fix(mcp): bound tool responses so one call cannot exhaust an agent context

redeye1011:
## Problem

Two MCP tools returned their whole backing collection. For an MCP consumer that is not merely slow — the response consumes the one budget the caller cannot expand.

**`memory_sessions`** was `kv.list(KV.sessions)` verbatim, pretty-printed, and declared no arguments at all:

```ts
inputSchema: { type: "object", properties: {} }
```

despite the description promising *recent* sessions. On a 3,646-session store that is **7.7 MB** in a single tool response, with no way for the caller to ask for less. The `summary` field alone was 76% of those bytes, and it is exactly the field a list view does not need.

**`memory_graph_query`** inherited the REST default of 500 nodes, which suits the viewer and is far too large for an agent: **14.8 MB** on the same install.

## What this changes

- `memory_sessions` takes `limit` (default 20, max 200), `project` and `status`, and returns a projected summary row — id, project, status, timestamps, observation count, title — newest first, with `total` / `returned` / `truncated` alongside. The full row is one `memory_recall` away.
- `memory_graph_query` defaults to 25 nodes and leaves provenance projected out unless `includeSources` is passed.
- One response ceiling applied to every tool on the way out (`AGENTMEMORY_MCP_MAX_RESPONSE_BYTES`, default 256 KB), so a tool added later cannot reintroduce the same failure. When it truncates it says so, and says which argument to narrow — a cut JSON payload is otherwise silently unparseable.
- Pretty-printing dropped from 57 responses, where it was roughly 13% pure whitespace.

## Result

Measured through the MCP shim, not the REST endpoint:

| | before | after |
|---|---|---|
| `memory_sessions` | 8,062,197 chars | **5,715** |
| `memory_graph_query` | 14,792,717 chars | **30,872** |

## Scope

Deliberately excludes two adjacent fixes already open, so this should not conflict with either:

- the `/sessions` REST `limit` param is **#1034** (issue #1022)
- the slot-tool guard and MCP error detail are **#894** and **#1149** (issue #1148)

This is only the MCP payload shaping.

## Verification

```
npm run build     # clean
npm test          # 1,711 passed, 1 skipped
```

Exercised against a live server by driving the shim over stdio: `initialize`, `tools/list`, then every read-only tool with schema-valid arguments.

## Summary by CodeRabbit

* **New Features**
  * Added configurable response-size limits for MCP tool and prompt results, with a notice when content is truncated.
  * Added filtering, projection, and result limits for session listings.
  * Added result limits and optional source details for memory graph queries.

* **Improvements**
  * Reduced response formatting overhead with compact JSON output.
  * Session results are now presented newest-first.
  * Added validation for session and graph query parameters to provide clearer errors.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -55,16 +55,43 @@ function mcpMaxResponseBytes(): number {
     : MCP_MAX_RESPONSE_BYTES_DEFAULT;
 }
 
-function capMcpResponse(res: McpResponse, toolName: string): McpResponse {
-  const content = (res?.body as { content?: unknown })?.content;
-  if (!Array.isArray(content)) return res;
+function truncationNotice(label: string, max: number): string {
+  return (
+    `\n\n[agentmemory] ${label} produced more than ${max} bytes and was ` +
+    `truncated, so the payload above is cut mid-value and will not parse ` +
+    `as JSON. Re-run with a narrower request (lower limit, add a filter, ` +
+    `or ask for fewer fields), or raise ` +
+    `AGENTMEMORY_MCP_MAX_RESPONSE_BYTES.`
+  );
+}
+
+function capMcpResponse(res: McpResponse, label: string): McpResponse {
+  const body = res?.body as
+    | { content?: unknown; messages?: unknown; error?: unknown }
+    | undefined;
+  if (!body) return res;
   const max = mcpMaxResponseBytes();
+
+  // Error bodies are not content arrays, and some of them echo a
+  // caller-supplied name, so they need the ceiling too.
+  if (typeof body.error === "string" && body.error.length > max) {
+    return { ...res, body: { ...body, error: body.error.slice(0, max) } };
+  }
+
+  const content = body.content;
+  if (!Array.isArray(content)) return res;
+
+  // Reserve the notice up front, so a response that overshoots by one
+  // character does not come back larger than the advertised ceiling.
+  const notice = truncationNotice(label, max);
+  const budget = Math.max(0, max - notice.length);
+
   let used = 0;
   let truncated = false;
   const next = content.map((part) => {
     const c = part as { type?: string; text?: string };
     if (c?.type !== "text" || typeof c.text !== "string") return part;
-    const room = max - used;
+    const room = budget - used;
     if (room <= 0) {
       truncated = true;
       return { ...c, text: "" };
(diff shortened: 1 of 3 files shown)


redeye1011:
Thanks — all four fixed and pushed.

- **Uncapped non-content bodies.** Correct; the `default:` branch echoed a caller-supplied tool name straight into `body.error`, past the ceiling. Error bodies now go through the cap, and the reflected name is bounded at the source rather than only downstream.
- **Notice pushed past the budget.** Correct — the notice was appended after the budget was already spent, so a response one character over came back larger than the ceiling it advertises. It's reserved up front now.
- **Coerced arguments.** Correct, and the `project` / `status` case was the one worth catching: a malformed filter was silently dropped, which widens the query to every project rather than failing. Both tools now return 400 for explicit-but-invalid `limit`, `project`, `status` and `includeSources`.
- **`mcp::prompts::get` bypassing the cap.** Correct; it can echo a caller-supplied `task_description`. Its dispatch now routes through the same cap.

Added `test/mcp-response-bounds.test.ts` covering projection, filtering, argument rejection, the ceiling holding on a truncated response, and the bounded unknown-tool error.

redeye1011:
Second round — one real bug in my own fix, now pushed.

**Fixed: the ceiling was measured in the wrong unit.** `AGENTMEMORY_MCP_MAX_RESPONSE_BYTES` is named in bytes, but the cap counted `String.length`, i.e. UTF-16 code units. CJK or emoji content passed the check at roughly **three times** the advertised limit, and slicing on the same metric could cut a multibyte sequence in half. Now measured with `Buffer.byteLength` and cut on a code-point boundary, dropping a trailing partial sequence rather than emitting a `U+FFFD` the caller never sent. Applies to content parts, the reserved notice, and error bodies.

Added a multibyte fixture that fails on the previous metric, and the test now measures bytes rather than characters.

**Fixed: the ordering assertion.** Right — checking that `startedAt` merely exists would also pass on insertion order. It now asserts the sequence is descending and pins the expected first row.

**Not changed: `vi.mock("iii-sdk")`.** I checked before declining: **0** test files in this repo mock `iii-sdk`, and **55** define a local `mockSdk`, including the MCP tests this one sits beside (`mcp-prompts.test.ts`, `mcp-resources.test.ts`). I followed the established house pattern deliberately so the file reads like its neighbours. Switching just this one file would make it the odd one out; if the convention is meant to change, that's a repo-wide change rather than something to land inside this PR.

Issue #1171, opened 2026-08-10
graph nodes: sourceObservationIds capped at creation but unbounded on merge


msbharathurs:
Node creation caps provenance at ten:

```js
sourceObservationIds: observations.slice(0, 10).map((o) => o.id)
```

but `mergeNode()` re-unions with no cap:

```js
function mergeNode(existing, incoming, obsIds, capturedAt) {
  return {
    ...existing,
    sourceObservationIds: [...new Set([
      ...existing.sourceObservationIds,
      ...incoming.sourceObservationIds,
      ...obsIds
    ])],
    ...
```

Because extraction re-observes the same entities continuously, the array grows monotonically for
the lifetime of the graph. There is no `slice`, length check, or config knob on that path (I
grepped `slice|MAX|cap|limit` around the field in `dist/index.mjs` for 0.9.28 — the create-time
slice is the only one).

**Impact.** `graph-query` returns node objects verbatim and accepts no field projection
(`limit`, `maxDepth`, `nodeType`, `offset`, `query`, `startNodeId`), so every consumer pays for
the accumulated provenance on every call. For an LLM agent that cost is context window.

Measured on a ~3-day-old graph (1,870 nodes / 2,974 edges):

| | |
|---|---|
| `graph/query "circuit"` → 3 nodes | **22,514 chars** |
| worst node (`asset_capital.py`) | 500 ids, 15,159 chars, of which **633 chars is signal** |
| waste on that node | **~95%** |

500 is not a ceiling, just that day's value. The cost is worst precisely where the graph is most
useful, since the most heavily-referenced entities accumulate the most provenance.

There is also no per-node update endpoint (`extract` / `build` / `reset` /
`snapshot-rebuild` only), so operators cannot compact existing nodes in place; `reset` discards
the whole graph.

**Suggested fixes (either would resolve it):**

1. Bound the merge, keeping the N most recent ids — mirrors the create-time cap and makes the two
   paths consistent:
   ```js
   const MAX = getGraphMaxSourceIds();   // e.g. GRAPH_MAX_SOURCE_IDS, default 10
   sourceObservationIds: [...new Set([...])].slice(-MAX)
   ```
2. Add a projection/verbosity parameter to `graph-query` (e.g. `includeSources: false`, or
   returning `sourceObservationCount` instead of the full array) so consumers can opt out.

Ideally both: (1) bounds storage growth, (2) lets callers control payload independently.

[...]

Issue #1168, opened 2026-08-08
Unbounded sourceObservationIds / sourceMemoryIds are what push collections past the worker limit in #1142 / #1124


krickwix:
## Summary

`sourceObservationIds` on graph nodes and `sourceMemoryIds` on insights grow without any cap. In both collections the provenance array ends up as **~97% of all bytes**, which is what pushes a collection past the worker limit described in #1142 and triggers the blast radius described in #1124.

Those two issues cover the *symptom* well. This one is about *why* collections get big enough to hit it in the first place — the size is not a function of how much you actually stored.

I hit this on two separate collections on one install and reset both, so there are before/after numbers below. The reflect measurement is probably the most useful part for #655.

## The two fields

| Collection | Field | Records | Field share of bytes | Worst single record |
|---|---|---|---|---|
| `mem:insights` | `sourceMemoryIds` | 8,969 | 97.0% | 1,545 ids |
| `mem:graph:nodes` | `sourceObservationIds` | 8,894 | 97.3% | **8,750 ids — 256 KB in one node** |

4,562,366 memory ids and 2,623,559 observation ids respectively. Strip the provenance and the same data is a few MB.

The worst node is a single `file` node for a `MEMORY.md` path. Nothing caps how many times a hot path can be re-observed.

## Two different mechanisms

**1. `mergeNode` / `mergeEdge` — monotonic union, never pruned:**

```js
sourceObservationIds: [...new Set([
  ...existing.sourceObservationIds,
  ...incoming.sourceObservationIds,
  ...obsIds
])],
```

Re-observing a node only ever adds. A frequently-touched file node grows unboundedly.

**2. `mem::reflect` — cluster-proportional at creation:**

```js
sourceMemoryIds: cluster.factIds,
```

Not accumulation (`reinforcements` was 0 on my largest record) but it scales with corpus size. At 25,408 semantic memories a single cluster contributed 1,545 ids.

## Measured impact, and a data point for #655

#655 attributes `memory_reflect` timeouts to sequential N+1 KV operations. That's plausible, but on my install **collection size dominated**. `mem::reflect` enumerates six collections (`graphNodes`, `graphEdges`, `semantic`, `lessons`, `crystals`, `insights`).

I reset the two bloated collections without changing anything else — no code change, no config change, same cluster count:

| | before | after |
|---|---|---|
| `mem::reflect` | `Invocation timeout after 180000ms`, failing 5 of 6 runs | **success in 15.2 s** |
[...]

Issue #1221, opened 2026-08-17
Graph node identity is the absolute path, so git worktrees and agent scratch dirs create duplicate nodes for the same file


reymondzzzz:
## Summary

`mem::graph-extract` uses the absolute filesystem path as the identity of a `file` node. On any workflow that materialises the same repository at more than one path — git worktrees, agent job scratch directories, temporary review checkouts — the same logical file becomes several unrelated nodes.

Two consequences:

1. **Knowledge is fragmented.** Facts learned about `backend/api/v1_orders_public.py` attach to whichever path the agent happened to be working in. A graph traversal from one of them never reaches the others, so the graph knows five partial things about one file instead of one complete thing.
2. **Nodes outlive their paths.** Worktrees get deleted and job directories get cleaned up, but their nodes stay in the graph forever. They are unreachable by any future session and cannot be distinguished from live nodes.

This is a different mechanism from #1168 / #1171 (unbounded `sourceObservationIds`). Those make each node bigger; this creates nodes that should never have existed separately. They compound: every duplicate node carries its own provenance array.

## Environment

| | |
|---|---|
| `@agentmemory/agentmemory` | 0.9.29 |
| iii engine | 0.11.2 (pinned) |
| OS | macOS 26 arm64 |
| LLM / embeddings | OpenRouter `deepseek/deepseek-v4-pro` + `openai/text-embedding-3-small` |
| Flags | `GRAPH_EXTRACTION_ENABLED=true`, `AGENTMEMORY_AUTO_COMPRESS=true`, `AGENTMEMORY_INJECT_CONTEXT=true` |
| Scale at measurement | 84 sessions, 11,040 observations, 8,092 graph nodes, 12,793 edges |
| Workflow | Claude Code with the `superpowers` plugin (`using-git-worktrees`, `dispatching-parallel-agents`) |

Paths below are anonymised; structure and counts are unchanged.

## Measurement

```bash
curl -s -X POST localhost:3111/agentmemory/graph/query \
  -H 'Content-Type: application/json' -d '{"limit":100000}' -o graph_dump.json
```

[...]

PR #1353, opened 2026-09-07
feat(graph): add mem::graph-prune with in-place provenance compaction

redeye1011:
Refs #1171
Refs #1168

> **Stacked on #1349.** This branch contains that commit as its base, so the diff shows both. Review the second commit, or merge #1349 first and this rebases to a single commit. The compaction pass reuses `GRAPH_MAX_SOURCE_IDS` from #1349 on purpose, so the backfill and the write path cannot disagree about the cap.

## Problem

The knowledge graph has no collector. Every extraction appends nodes and edges and nothing removes them, so a corpus eventually passes the 25K-node ceiling that `mem::graph-snapshot-rebuild` refuses to run above, and there is no path back short of `graph-reset`, which discards the whole graph.

#1171 names the half that bounding does not solve:

> There is also no per-node update endpoint (`extract` / `build` / `reset` / `snapshot-rebuild` only), so operators cannot compact existing nodes in place; `reset` discards the whole graph.

Capping the write path only helps rows written afterwards. An install that already accumulated millions of provenance ids keeps paying for them on every read, forever, unless it throws the graph away.

## What this adds

`mem::graph-prune`, exposed at `POST /agentmemory/graph/prune`.

**Collection** — three classes that are provably dead:

- rows already tombstoned as `stale`
- edges whose endpoint node no longer exists, so they can never be traversed
- temporal edges marked `isLatest: false` past a retention cutoff (default 90 days), i.e. history already replaced by a newer revision

It also clears the matching `graphNameIndex`, `graphEdgeKey` and `graphNodeDegree` entries, so `graph-extract` dedup cannot later resolve a name into a deleted node — index entries are only removed when they still point at the row being deleted.

**Compaction** — with `compactSourceIds`, trims oversized provenance arrays to the newest ids using the same cap the write path applies. This is the backfill the issue asks for.

**Safety** — defaults to a dry run, which reports what each class would collect without writing.

Duplicate-name merging was in the first revision and has been removed after review: it rewrites live rows rather than removing dead ones, and the review surfaced four separate correctness problems in it (uncapped keeper union, losers marked for deletion before the keeper write succeeded, edge-key collisions when two duplicates share a same-type edge to the same node, degree reconciliation). It was opt-in and never exercised on a real corpus. The survey still reports the duplicate count, which is the useful half; merging deserves its own PR with its own design.

## Result

On a 71.5K-node / 191K-edge corpus:

```json
{ "oversizedNodes": 26686, "oversizedEdges": 28884,
  "droppableSourceIds": 2628533, "staleNodes": 0,
  "danglingEdges": 0, "supersededEdges": 0, "duplicateNodes": 885 }
```

Compaction run: 26,686 nodes and 28,884 edges in **22.4s**, 0 errors. Daemon RSS **1636 MB → 922 MB**. A second dry run reports zero of everything, so it is idempotent.

Worth noting what the survey found: on this corpus there was no stale, dangling or superseded garbage at all. The entire recoverable size was provenance. That is a useful argument for the cap in #1349 — collection alone would have recovered nothing here.

## Verification

```
npm run build     # clean
npm test          # 1,720 passed, 1 skipped
```

New `test/graph-prune.test.ts`: dry run reports without writing, stale and dangling rows are removed, provenance is left alone unless compaction is asked for, compaction keeps the newest ids, and a second pass finds nothing left.

Live:

```bash
curl -s -X POST localhost:3111/agentmemory/graph/prune \
  -H 'content-type: application/json' -d '{"dryRun":true}'
# then {"dryRun":false,"compactSourceIds":true}
```

## Open questions

- Retention default for superseded edges is 90 days; happy to change it.
- The endpoint is deliberately manual. It could be driven from a timer, but deleting graph rows on a schedule felt like something an operator should opt into rather than inherit.

## Summary by CodeRabbit

* **New Features**
  * Added a graph maintenance endpoint with safe dry-run defaults for identifying stale, dangling, duplicate, and oversized graph data.
  * Added optional cleanup for stale graph data and provenance compaction, with configurable provenance limits.
  * Graph queries now return provenance counts and a sample by default, with an option to include full details.
* **Documentation**
  * Updated documented REST API endpoint totals from 130 to 131.
* **Tests**
  * Added coverage for provenance limits, graph maintenance, compaction, deletion, and repeatable cleanup.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -117,7 +117,7 @@ Hook scripts in `src/hooks/` are standalone Node.js scripts (no iii-sdk import).
 ## Current Stats (v0.9.29)
 
 - 54 MCP tools (8 visible by default, `AGENTMEMORY_TOOLS=all` for all)
-- 130 REST endpoints
+- 131 REST endpoints
 - 6 MCP resources, 3 MCP prompts
 - 12 hooks, 17 skills
 - 260+ iii functions
--- a/README.md
+++ b/README.md
@@ -1609,7 +1609,7 @@ Create `~/.agentmemory/.env`:
 
 <h2 id="api"><picture><source media="(prefers-color-scheme: dark)" srcset="assets/tags/light/section-api.svg"><img src="assets/tags/section-api.svg" alt="API" height="32" /></picture></h2>
 
-130 endpoints on port `3111`. The REST API binds to `127.0.0.1` by default. Protected endpoints require `Authorization: Bearer <secret>` when `AGENTMEMORY_SECRET` is set, and mesh sync endpoints require `AGENTMEMORY_SECRET` on both peers.
+131 endpoints on port `3111`. The REST API binds to `127.0.0.1` by default. Protected endpoints require `Authorization: Bearer <secret>` when `AGENTMEMORY_SECRET` is set, and mesh sync endpoints require `AGENTMEMORY_SECRET` on both peers.
 
 <details>
 <summary>Key endpoints</summary>
--- a/src/config.ts
+++ b/src/config.ts
@@ -385,6 +385,26 @@ export function getGraphBatchSize(): number {
   return safeParseInt(getMergedEnv()["GRAPH_EXTRACTION_BATCH_SIZE"], 10);
 }
 
+// Upstream #1168 / #1171: `sourceObservationIds` on graph nodes (and the
+// same field on edges) was capped at creation but re-unioned without a
+// cap on every merge. Because extraction re-observes the same entities
+// continuously, the array grew monotonically for the life of the graph
+// and ended up as ~97-99% of all bytes in the collection — measured here
+// at 9.55 MB of 9.4 MB across 500 nodes, with one `package.json` node
+// holding 4,707 ids in 133 KB. Provenance past the most recent handful
+// carries almost no signal, so bound it and keep the newest.
+const GRAPH_MAX_SOURCE_IDS_DEFAULT = 10;
+
+export function getGraphMaxSourceIds(): number {
+  return Math.max(
+    1,
+    safeParseInt(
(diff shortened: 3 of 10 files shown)


redeye1011:
Thanks — this was useful, and it changed the shape of the PR.

Four of the five findings were on the duplicate-merge path: the uncapped keeper union, losers marked for deletion before the keeper write succeeded, edge-key collisions when two duplicates share a same-type edge to the same node, and degree reconciliation. They're all correct, and together they say the merge path wasn't designed carefully enough. It was opt-in and I never ran it against a real corpus.

**I've removed duplicate merging from this PR** rather than patch four holes in a feature that isn't the subject of the change. The survey still reports how many duplicate names exist, which is the useful half; merging deserves its own PR with its own design — including the collision resolution you describe. Worth noting the duplicates on my corpus mostly trace to #1221 (absolute path as node identity), so merging may be the wrong layer to fix them at anyway.

The two findings that apply to what this PR does ship are fixed:

- **Snapshot not reflecting compaction.** Correct and important — the snapshot serves `/graph/query` on the no-argument path, so a compaction run reported success while queries kept returning the provenance it had just trimmed. `topNodes` / `topEdges` are now patched with the compacted arrays.
- **Snapshot reconciled against candidates rather than successes.** Correct; one failed delete in a run could drop a still-persisted row out of snapshot-backed queries. Successful deletions are tracked separately and drive the filter and the totals.

redeye1011:
Heads-up: the five comments dated 19:01 were posted against the previous revision — they reference `mergeDuplicateNames`, `idRemap` and the repoint block at lines 250–286, all of which are gone as of `a4e367c`. `grep -c mergeDuplicateNames src/functions/graph-prune.ts` is now 0.

The two findings that survive the removal are fixed, as described in my previous comment. I've also dropped a stale line in the file header that still described merging as an opt-in flag.

PR #1366, opened 2026-09-12
fix(hooks): back-fill prompts from Claude Code transcripts too

### LINKED ISSUE #1365: session-end transcript back-fill extracts no prompts from a Claude Code transcript
`extractTranscriptPrompts` in `src/hooks/session-end.ts` returns an empty array for every
Claude Code transcript, so the prompt back-fill added for the SessionEnd hook silently posts
nothing. Sessions then reach the daemon carrying only `{endedAt, status}`, and `mem::summarize`
logs `No observations to summarize` for them indefinitely.

## Cause

The extractor matches one transcript dialect:

```ts
if (msg.role !== "user") continue;
for (const block of msg.message?.content ?? []) {
  if (block.type !== "text" || typeof block.text !== "string") continue;
```

That is the Cursor shape — `role` at the top level, `content` always an array of typed blocks,
which is what `test/session-end-transcript.test.ts` builds. A Claude Code transcript line is
shaped differently:

```json
{"type":"user","message":{"role":"user","content":"the prompt text"},"sessionId":"...","cwd":"..."}
```

The role is nested under `message` and the top-level field is `type`, so `msg.role` is
`undefined` and every line is skipped. Past that first guard there is a second mismatch: a plain
user turn carries `content` as a **bare string**, so `for (const block of content)` would
iterate the string's characters and `block.type` would be `undefined` on each.

## Reproduction

```js
// transcript.jsonl
{"type":"user","message":{"role":"user","content":"first prompt"}}
{"type":"user","message":{"role":"user","content":[{"type":"text","text":"second prompt"}]}}
```

```
echo '{"session_id":"s1","hook_event_name":"sessionEnd","reason":"completed","transcript_path":"transcript.jsonl"}' \
  | node plugin/scripts/session-end.mjs
```

Expected two POSTs to `/agentmemory/observe`; observed none, followed by the usual
`/agentmemory/session/end`.

Measured on a real 1187-line Claude Code transcript holding 109 user records: **0 prompts
extracted**. Five of those records are genuine user turns; the other 107 are `tool_result`
blocks that should stay excluded.

## Effect

On a store where this is the only capture path, nothing has been summarized since the last
manual import: 9201 session records, of which only 574 carry `startedAt` / `observationCount` /
`summary`, and those 574 all came from `mem::import`. `Consolidation pipeline complete` keeps
reporting success against that frozen corpus (`totalSummaries` constant), so nothing in the logs
indicates capture has stopped.

## Fix

Accept `type: "user"` and `message.role === "user"` alongside the existing top-level `role`, and
read a string `content` as a single text. Skipping `isSidechain` turns keeps subagent traffic out.
PR follows.

dmazhukov:
Makes the SessionEnd prompt back-fill recognise Claude Code transcripts. Fixes #1365.

`extractTranscriptPrompts` matched only the Cursor dialect — `role` at the top level, `content`
always an array of typed blocks — so on a Claude Code transcript, where the role is nested under
`message` and a plain user turn carries `content` as a bare string, every line was skipped and
the back-fill posted nothing. Sessions then arrive as bare `{endedAt, status}` records and
`mem::summarize` reports `No observations to summarize` for them forever, while
`Consolidation pipeline complete` keeps reporting success against the frozen corpus.

The extractor now accepts `type: "user"` and `message.role === "user"` alongside the original
top-level `role`, and reads a string `content` as a single text. Only text blocks are collected,
so the `tool_result` records that make up most Claude Code user lines stay out, and `isSidechain`
turns are skipped because a subagent is not the user speaking. The Cursor path, the
`<user_query>` unwrap and the 50-prompt / 8000-character caps are untouched.

Measured on a real 1187-line Claude Code transcript with 109 user records: 0 prompts extracted
before, 5 after — the session's actual user turns, with the remaining 107 `tool_result` records
correctly excluded. Verified end-to-end against a live daemon: five observations captured, the
session record gained `observationCount` and `project`, and `Session summarized` fired for a
session that had produced nothing before.

The new case in `test/session-end-transcript.test.ts` covers the bare-string turn, `tool_result`
exclusion and the sidechain skip.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved session transcript processing across supported transcript formats, including Claude Code sessions.
  - User prompts are now detected more reliably from nested role metadata and alternate content formats.
  - Text prompts are extracted correctly when tool results or other non-text blocks are present.
  - Sidechain and subagent turns are excluded from prompt submissions.

- **Tests**
  - Added coverage for Claude Code transcripts, tool-result blocks, and sidechain turns.

### DIFF
--- a/plugin/scripts/session-end.mjs
+++ b/plugin/scripts/session-end.mjs
@@ -45,6 +45,17 @@ function authHeaders() {
 	if (SECRET) h["Authorization"] = `Bearer ${SECRET}`;
 	return h;
 }
+function isUserTurn(msg) {
+	if (msg.isSidechain) return false;
+	return msg.role === "user" || msg.type === "user" || msg.message?.role === "user";
+}
+function turnTexts(content) {
+	if (typeof content === "string") return [content];
+	if (!Array.isArray(content)) return [];
+	const texts = [];
+	for (const block of content) if (block?.type === "text" && typeof block.text === "string") texts.push(block.text);
+	return texts;
+}
 function extractTranscriptPrompts(data) {
 	const path = data.transcript_path;
 	if (typeof path !== "string" || !path.endsWith(".jsonl")) return [];
@@ -63,12 +74,11 @@ function extractTranscriptPrompts(data) {
 		} catch {
 			continue;
 		}
-		if (msg.role !== "user") continue;
-		for (const block of msg.message?.content ?? []) {
+		if (!isUserTurn(msg)) continue;
+		for (const raw of turnTexts(msg.message?.content)) {
 			if (prompts.length >= 50) return prompts;
-			if (block.type !== "text" || typeof block.text !== "string") continue;
-			const m = block.text.match(/<user_query>\n?([\s\S]*?)\n?<\/user_query>/);
-			const text = (m ? m[1] : block.text).trim();
+			const m = raw.match(/<user_query>\n?([\s\S]*?)\n?<\/user_query>/);
+			const text = (m ? m[1] : raw).trim();
 			if (text) prompts.push(text.slice(0, 8e3));
 		}
 	}
--- a/src/hooks/session-end.ts
+++ b/src/hooks/session-end.ts
@@ -25,11 +25,6 @@ type TranscriptLine = {
   message?: { role?: string; content?: string | TranscriptBlock[] };
 };
 
-// Two transcript dialects reach this hook. Cursor writes the role at the top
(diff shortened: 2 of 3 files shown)


dmazhukov:
Taken in f05f5b6. Both comments are gone: `AGENTS.md` asks for no comments explaining what the code does, and the type already carries what the prose described, since an optional top-level `role` sits beside `type`, `message.role` is nested, and `content` is typed `string | TranscriptBlock[]`. The remaining sentence was the history of the fix, which reads better in the PR description than in the source.

